PR #179, opened 2014-09-29
Added support for feature parsing from the genbank file. 

paolopavan:
It supports also qualifiers and complex nested locations

PR #224, opened 2015-01-06
Added support for feature parsing from the genbank file (rebased)

sbliven:
Incorporates work by @paolopavan and @jgrzebyta on a new and improved genbank parser. It supports feature parsing, qualifiers and complex nested locations.

This fork rebases the original development in order to avoid a large merge commit such as in pull #179, but is otherwise equivalent.


sbliven:
Note that intermediate commits don't compile here. That's not ideal. I personally prefer the merged #179 version, which better reflects the actual dev history and has better intermediate steps.

andreasprlic:
Thanks Spencer. In this case let's stick to the original #179 pull request.

Issue #225, opened 2015-01-07
Drop support for non-split PDB directories


sbliven:
Following #181 we use a heirarchical directory structure for storing PDB files, e.g. `PDB_DIR/data/structures/divided/mmCIF/xx/1xx0.cif.gz`. But we still have an isSplit flag, which controls whether it is written to the `divided/mmCIF/xx/` or the `all/mmCIF/` directory.

I don't see any use cases where you would prefer the `data/.../all/...` directory. I suggest we remove the isSplit option from StructureIOFile implementations and AtomCache.

sbliven:
One usecase I *can* see is that for storing all files directly in `PDB_DIR`. For instance, PyMOL downloads files to a flat directory (which I make the same as PDB_DIR), so that would reduce redundant downloads between the two programs.

A more complete (and complex) solution would be to replace the isSplit boolean with an enum of various preferred directory layouts:

```java
public enum PDBDirectoryLayout {
  HEIRARCHICAL_SPLIT, // use PDB_DIR/data/.../divided/.../xx
  HEIRARCHICAL_ALL, // use PDB_DIR/data/.../all/...
  FLAT_SPLIT, // use PDB_DIR/xx
  FLAT_ALL; // use PDB_DIR/ for everything
}
```

This would be pretty straightforward to implement if anyone thinks it would be useful. Note that the FLAT_ALL option would also put obsolete files, biological assemblies, chemical components, etc. directly in PDB_DIR.

pwrose:
+1 from me.

Issue #173, opened 2014-09-19
Default to mmCIF 


sbliven:
With the [wwPDB's announcement](http://www.wwpdb.org/news/news_2014.html#18-September-2014) that starting Dec 10 large structures (#156) will not be available in PDB format, perhaps now is a good time to switch the default format to mmCif.

Simply changing the default setting of `AtomCache.useMmCif` should be sufficient.

sbliven:
#228 adds obsolete file support, so I think this issue is done.

Issue #169, opened 2014-09-17
Why different default ChemCompProviders for different parsers?


josemduarte:
It seems that the `ChemCompProvider` defaults are:
- for PDB parser it is `ReducedChemCompProvider` 
- for mmCIF parser it is `DownloadChemCompProvider`

is there a reason for that?

See also issue #145

PR #228, opened 2015-01-12
Refactor PDB/MMCIFFileReader

### LINKED ISSUE #226: Drop support for File-based Biological assemblies
Is there any reason to continue using the PDB-provided biological assembly files? It makes more sense to me to generate them directly from the files.

sbliven:
Extracts common code from PDBFileReader and MMCIFFileReader into the common LocalPDBDirectory. Most of this functionality dealt with managing the local cache of structure files from the PDB ftp server. Thus, both mmcif and pdb file formats now have a consistent interface (relating to #173, but not solving #169) and equivalent capabilities with respect to file fetching and obsolete file handling.

Two major behaviors are specified with new enums:
- **`FetchBehavior`** controls when files are touched from the FTP server, and replaces the old autoFetch parameter. Old methods using are deprecated and should be replaced with the `get/setFetchBehavior` methods in `AtomCache` and `StructureIOFile` implementations. Also, the default behavior for `LocalPDBDirectory` was changed to fetch files, to match the previous AtomCache default.
- **`ObsoleteBehavior`** controls whether to update obsolete structures to the current version. This was previously available for PDB files only.

Support for the file-based biological assemblies has been dropped, as BioJava can load these from the header of both PDB and MMCIF files (fix #226).

Removes support for non-split local directory layouts, since nobody seems to be using this feature (#225).

References to newly deprecated references were removed, as well as a number of unrelated warnings.


andreasprlic:
nice one!

PR #229, opened 2015-01-13
Bio-assembly handling improvements

josemduarte:
A few commits related to bioassembly handling:
- some bug fixes
- better docs
- BiologicalAssemblyTransformation now using vecmath's Matrix4d. In this way it is more inter-operable with code in the `org.biojava.bio.structure.xtal` package
- introduced new class `BioAssemblyInfo` to encapsulate the info needed to reconstruct the bio assembly. It includes a new feature: macromolecular size of the assembly which is now parsed from the PDB/mmCIF files


sbliven:
Thanks for fixing my bugs from #228

PR #247, opened 2015-01-28
Fix #244: AtomPositionMap bug fixes

### LINKED ISSUE #244: Incorrect length calculation in AtomPositionMap
The calcLength() methods of AtomPositionMap don't include the end residue, lead to off-by-one errors. This is particularly apparent when parsing ResidueRangeByLength objects, where "A:1-2" has a length of 1.

sbliven:
Provides many changes to `AtomPositionMap` and `ResidueRangeAndLength`.

* Replaces calcLength methods with getLength methods, which return the full length rather than length-1.
* Better inference for full-chain ranges
* Limited support for scop-style "_" wildcard chain names
* Better testing

### DIFF
--- a/biojava3-structure/src/main/java/org/biojava/bio/structure/AtomPositionMap.java
+++ b/biojava3-structure/src/main/java/org/biojava/bio/structure/AtomPositionMap.java
@@ -23,18 +23,16 @@
 
 package org.biojava.bio.structure;
 
-import org.biojava.bio.structure.io.mmcif.chem.ResidueType;
-
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.NavigableMap;
-import java.util.Set;
 import java.util.TreeMap;
-import java.util.TreeSet;
+
+import org.biojava.bio.structure.io.mmcif.chem.PolymerType;
+import org.biojava.bio.structure.io.mmcif.chem.ResidueType;
 
 /**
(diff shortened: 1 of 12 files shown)


sbliven:
@dmyersturnbull  I didn't look at the equals method. If I broke the contract we should fix it. However, for me that's the logical behavior. A ResidueRange should never equal a ResidueRangeAndLength (or maybe it should have a separate method for this). And if you have the map anyways, why not fill in those null values?

However, if you feel strongly about this it would be easy to change back without breaking much else. I don't rely on lengths or non-nulls anywhere important.

dmyersturnbull:
Sorry, I mistakenly wrote ResidueRange rather than ResidueRangeAndLength. There are two points.

The first issue is that: ResidueRangeAndLength.parse("B:", map).equals(new ResidueRange**AndLength**("B", null, null, length)) is false.

The second is that ResidueRangeAndLength.parse("B:").getStart() returns ResidueNumber("1") while ResidueRange.parse("B:").getStart() returns null, which doesn't break any contract but I think is very confusing. It also means that ResidueRangeAndLength.parse("B:", map).toString() returns "B:1-30", while ResidueRange.parse("B:").toString() returns "B:".

sbliven:
I'm going to merge this because of the upcoming code freeze, but we can continue this discussion and add a quick patch if needed.

I don't find it confusing at all that objects constructed with different constructors would be unequal. If we switched things to store nulls, then we would have ResidueRangeAndLength.parse("B:1-30") to be different from ResidueRangeAndLength.parse("B:"), even though they give equivalent substructures. In all cases there are multiple unequal ResidueRanges which will give the same result when reducing a structure; this is an unavoidable consequence of having a flexible format. To me it seems preferable to use all information available at the time of construction.

Issue #81, opened 2013-11-30
Make loading of structures more consistent


dmyersturnbull:
Right now, loading of data from SCOP, CATH, and AtomCaches is handled very differently. It would be nice to have 1) an interface that is more consistent between SCOP and CATH, and 2) a more robust way to load arbitrary "structures" given unique identifiers.

sbliven:
I'm adding this to the 4.0.0 milestone, since it will add a lot of methods and is closely related to #237.  Let's git 'er done!

sbliven:
I feel like what I really want here is a full-blown selection logic. `StructureIdentifier` implementations should be a set of methods which map from a String (either a short unique ID, as in domains) to some sort of unambiguous, memory-compact structure query.

Currently we support queries which contain a unique PDB ID and one or more sequential sequences of residues. Is there a way we could generalize this to extract any arbitrary subset of a Structure? `StructureIdentifier` operators could be applied on any level of the `Structure` hierarchy (with more levels potentially being added by #220):

1. PDB ID
2. Model
3. Chain
4. Residue
5. Atom

The problem is that the representation needs to be compact, unambiguous, and independent of the Structure it applies to.

pwrose:
How about:

Without model specification:
<pdbid>.<chainid>.<residueid>.<atomname>.<alternative location id>

With model specification (consider pdbid as an array of models),
<pdbid>[modelnumber].<chainid>.<residueid>.<atomname>.<alternative location
id>

residue id would need optionally include a residue number and insertion
code:
ASP - any ASP
ASP25 -  with residue number
ASP25C - with residue number and insertion code

atom name:
CA - C alpha
CA.A - C alpha alternative location A

-Peter

dmyersturnbull:
I think the selection logic needs to be non-hierarchical. Otherwise, (e.g.) to specify the backbone of 1HIV, we would need "1HIV.A.1.CA, 1HIV.A.2.CA, ..., 1HIV.A.n.CA, 1HIV.B.1.CA, 1HIV.B.2.CA, ..."

Perhaps Identifier can hold a PDB id and an order-independent set of Selection objects, where each Selection has a unique string representation. For example, an Identifier for the above case could store pdbId = "1HIV" and the Selection object "#CA", where "#" selects atoms.

For more complex cases, such as selecting just CA atoms of chain A but all atoms of chain B, we could permit nesting of Selection objects or an aggregate Selection type.

We could have a Selection interface with a single apply(Structure) method.
[...]

PR #248, opened 2015-01-29
ResidueRange bug fixes

dmyersturnbull:
a7bd1c4 is an important fix that would be nice to have in a Biojava 4 hotfix after the release.

### DIFF
--- a/biojava3-structure/src/main/java/org/biojava/bio/structure/AtomPositionMap.java
+++ b/biojava3-structure/src/main/java/org/biojava/bio/structure/AtomPositionMap.java
@@ -23,6 +23,9 @@
 
 package org.biojava.bio.structure;
 
+import org.biojava.bio.structure.io.mmcif.chem.PolymerType;
+import org.biojava.bio.structure.io.mmcif.chem.ResidueType;
+
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashMap;
@@ -30,9 +33,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.NavigableMap;
 import java.util.TreeMap;
-
-import org.biojava.bio.structure.io.mmcif.chem.PolymerType;
-import org.biojava.bio.structure.io.mmcif.chem.ResidueType;
 
 /**
  * A map from {@link ResidueNumber ResidueNumbers} to ATOM record positions in a PDB file.
@@ -49,7 +49,8 @@ import org.biojava.bio.structure.io.mmcif.chem.ResidueType;
  * </pre>
  * 
  * <p>Note: The getLength() methods were introduced in BioJava 4.0.0 to replace
- * the calcLength methods, which returned the length-1
+ * the calcLength methods. The new method returns the number of residues between
+ * two residues, inclusive, whereas the previous method returned 1 less than that.
  * @author dmyerstu
  */
 public class AtomPositionMap {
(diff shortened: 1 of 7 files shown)


dmyersturnbull:
@sbliven The ResidueRange regex doesn't accept "1a4w.I", etc. Should we update it to allow this case? And what about "1a4w.I_" and "1a4w.I_1-"?

dmyersturnbull:
Also, StructureName.parseRanges() is broken because it passes the PDB Id to ResidueRange.parse().

andreasprlic:
do you want me to still pull it into 4.0

andreasprlic:
currently fails with 

java.lang.IllegalArgumentException: Could not understand ResidueRange string 4hhb:1
	at org.biojava.bio.structure.ResidueRange.parse(ResidueRange.java:88)
	at org.biojava.bio.structure.ResidueRange.parseMultiple(ResidueRange.java:101)
	at org.biojava.bio.structure.align.client.StructureName.parseRanges(StructureName.java:54)
	at org.biojava.bio.structure.align.client.StructureName.<init>(StructureName.java:50)
	at org.biojava.bio.structure.align.util.AtomCache.getStructure(AtomCache.java:365)
	at org.biojava.bio.structure.TestAtomCache.testAtomCacheNameParsing(TestAtomCache.java:93)

dmyersturnbull:
That test failed because it was using the syntax "4hhb:1" and "4hhb(A:10-20,A:30-40)" (with parenthesis). I believe we no longer support these, so I disabled the test.

I think we should merge this if @sbliven agrees. He authored these changes for #81 and #248 .

andreasprlic:
java.lang.IllegalArgumentException: Could not understand ResidueRange string ent.gz?args=test
	at org.biojava.bio.structure.ResidueRange.parse(ResidueRange.java:88)
	at org.biojava.bio.structure.ResidueRange.parseMultiple(ResidueRange.java:101)
	at org.biojava.bio.structure.align.client.StructureName.parseRanges(StructureName.java:54)
	at org.biojava.bio.structure.align.client.StructureName.<init>(StructureName.java:50)
	at org.biojava.bio.structure.align.util.AtomCache.getStructure(AtomCache.java:365)
	at org

sbliven:
In principle I'm ok with removing the chains via numbers syntax. It can be re-added in 4.1.0 after my #81 changes get merged, which redo this test case properly.

However, the tests were passing before your changes. Why do you want to add partial ranges? They weren't supported before.

dmyersturnbull:
@sbliven  I didn't add support for partial ranges. The issue was that ResidueRangeAndLength.parse("C_1023-", map) for 2eke returned a range for chain A rather than the expected length for chain C. Also, ResidueRange.parse("%#*&%!@") would return a ResidueRange. To fix this, I did a7bd1c4, which causes an IllegalArgumentException to be thrown when the range regex doesn't match. Unfortunately, this change broke a number of tests, which needed to be fixed.

The related issue was that StructureName.parseRanges() was wrong because it would pass "1hiv.A_1-5" to ResidueRange.parse, which doesn't know what to do with a PDB Id. The tests were passing only because ResidueRange.parse was just returning ResidueRange(null, null, null) rather than throwing an IllegalArgumentException.

andreasprlic:
is it easier to merge this at the same time as #81 for release 4.1 ?

sbliven:
So all these issues are fixed in #81, but I wanted to discuss & test those changes more. Should we revert the 2aaed4769e56 merge too?

dmyersturnbull:
@sbliven For Biojava 4, can we just make StructureName not implement StructureIdentifier and remove its getRanges()? I think that would solve these problems.

sbliven:
I actually spent quite a bit of time converting StructureName to be a catch-all StructureIdentifier. I think it works ok. After the release maybe we can skype re #81. I fixed a lot of these problems here, but it still makes me nervous how fragile string parsing is.

sbliven:
If tests pass I'd say keeping StructureName independent for now is a good solution.

dmyersturnbull:
Looks good to merge now to me.

PR #256, opened 2015-02-24
Try to fix TimeoutException

jgrzebyta:
I try to fix TimeoutException in the unit tests. I identified place where the exception occures and added solution (customised timeout to 20 sec).


andreasprlic:
I'd be surprised if adding 5 more seconds fixes the timeout issues, but let's give it a try...

jgrzebyta:
Hi,

I would be surprised either. :-) In total it's 60sec waiting + in case of TomoutException the whole block is repeated (up to 10 times) and that's the proper solution imho. :-)

PR #257, opened 2015-02-28
Cleanup of unnecessary casts

pibizza:
I have tried to remove lots of unnecessary casts from the code.


pwrose:
Hi Paolo,

At first sight it might seem like a good idea to change for example:
public Object clone() to public Atom clone() to avoid the casts,

however, there is a catch!

A class that implements the Cloneable interface must to overwrite public Object clone();

-Peter

pibizza:
Oh, I see. My mistake. You never stop learning something new. Thanks.

phidias51:
The other problem is you won't know you've broken anything until you try to
execute the statement.

Mark

PR #258, opened 2015-03-11
Evolutionary Classification of Protein Domains (ECOD) support

sbliven:
Adds support for the ECOD classification, similar to the existing SCOP and CATH classifications.


sbliven:
The ECOD tests require downloading two 70MB files from the server. This could lead to problems with Travis, and is rather slow.

Perhaps we should exclude EcodInstallationTest from maven, so that it only runs when specifically requested? If so, here's the patch: 273dc2d832

sbliven:
Tests pass locally.

andreasprlic:
Thanks, this is great! Also good not to include the 70 MB download in the tests...

PR #260, opened 2015-03-16
ECOD updates

sbliven:
Integrates recent improvements by the ECOD team to the file format.


sbliven:
Also limits the number of warning lines, which should fix the travis problems

sbliven:
Testing over all versions leads to memory problems in maven/travis.

PR #261, opened 2015-03-17
ECOD updates + merge patch

sbliven:
This adds improments to the ecod package for file format and memory usage.

It also merges the patch branch, which includes some genbank improvements.


sbliven:
Also disables ECOD from unit tests unless specified with -Dtest=EcodInstallationTest

PR #265, opened 2015-03-26
Patch

lafita:
Summary of the changes:

AlignmentTools: new method to calculate block gaps, extended updateSuperposition to calculate the score and RMSD of each block, method to replace an optimal alignment in an AFPChain.

DisplayAFP: in the alignment panel non FatCat alignments only considered a gap between two residues if a dash ('-') was present in any of the two sequences, but in the CESymm case, a gap can be also described by a space in the symb variable of the AFPChain (without a dash in any of the two sequences).

CECalculator: matrix listener used to maintain the black out regions of the matrix mat during the optimization step of the alignment (only applied to the CeSymm class, the matrix listener was already present in the CECalculator).

AFPChain: added a new member variable to set the colors of the jmol display, named blockColors. Code in the AligPanel has been extended to color the subunits (blocks) accordingly if the variable is not null.

### DIFF
--- a/biojava-structure-gui/src/main/java/org/biojava/nbio/structure/align/gui/DisplayAFP.java
+++ b/biojava-structure-gui/src/main/java/org/biojava/nbio/structure/align/gui/DisplayAFP.java
@@ -75,16 +75,16 @@ public class DisplayAFP
 
 	private static boolean isAlignedPosition(int i, char c1, char c2, boolean isFatCat,char[]symb)
 	{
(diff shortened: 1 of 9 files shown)


sbliven:
@andreasprlic Any idea why in DisplayAFP#isAlignedPosition, space characters were counted as gaps in fatcat but not in CE? Treating them as gaps globally seems to work fine, but I was wondering if there was some special edge case that necessitated that test in the first place.

Issue #126, opened 2014-05-12
Improve structural alignment datastructures


sbliven:
BioJava contains a number of algorithms for aligning protein structures. In the most general case, an alignment consists of a mapping between residues of two (or more) proteins. However, for historical and performance reasons alignments are stored as linear, sorted arrays. This makes it difficult to express cases where the order of aligned residues differs between the two proteins. For instance, storing the following alignment requires some creative work-arounds:

     123456
     456123

Additionally, the class to store structural alignments (AFPChain) contains a number of unneccessary, poorly documented, or algorithm-specific parameters which should be removed or refactored.

AFPChain should be refactored into a new data structure that
*    Is flexible enough to store topology-independent alignments
*    Efficiently utilizes memory
*    Has good performance for common tasks 
* Works for rigid body, flexible, and local structural alignments
* Maintains consistency when modified

_Suggested for [GSoC 2013](http://biojava.org/wiki/Google_Summer_of_Code_2013)_

lafita:
This should have been closed when the MultipleAlignment Data Structure (to store multiple structural alignments) was merged #278.

PR #277, opened 2015-06-05
Multiple Structure Alignment Datastructures

sbliven:
Introduces new data structures for structure alignments, created along with @lafita. The data structure can represent standard pairwise alignments, but also multiple alignments, flexible alignments, and non-topological alignments (#126).

The structure consists of a hierarchy of objects:

1. **MultipleAlignmentEnsemble** A collection of alignments over a set of structures
2. **MultipleAlignment** A single alignment
3. **BlockSet** A portion of the alignment with a single rigid superposition
4. **Block** A portion of the alignment with preserved sequence order. Stores the actual aligned residues for each position (which may be gaps).

Some documentation still needs to be written and will be added to the cookbook.

A few other design decisions bear mention:

* The ensemble stores references to the aligned atom arrays.
* In the normal case, atom arrays can be regenerated from the structure names list, which is currently of type `String` but will change to `StructureIdentifier` following the completion of #81.
* All levels of the hierarchy can serve as a cache for various scores (e.g. RMSD, TM-Score, etc), but such scores are not standardized and should be recalculated when needed by client code
* The superposition matrices are now stored using vecmath `Matrix4d` objects. To support flexible alignments, the definitive matrices are stored in each `BlockSet`. However, a default matrix can be stored in `MultipleAlignment` to save memory for rigid alignments.
* `AFPChain` can be converted directly to `MultipleAlignmentEnsemble`

This pull request also bundles concurrent development of:
* GUI improvements
* Ghe creation of a monte-carlo based optimization strategy for refining structural refinements
* A new `AtomCache.getRepresentativeAtoms()` method (that should replace `getAtoms()` everywhere)

etc. etc. !

This is a fairly major feature addition, so I'll leave this request open for a few days to allow comments.


sbliven:
Oops, this can go in 4.1

sbliven:
Replaced by #278

Issue #234, opened 2015-01-14
Parsing mmCIF files produced by phenix


josemduarte:
Currently the mmCIF parser can't parse phenix-produced files. The test added in commit 9cae80fcfaa8174caaadea1620882dbfff3ed4e2 demonstrates the issue (The Test tag is commented out for now in order for the build to work. It should be uncommented once this is fixed)

Looking at it quickly I think that the problem comes from the separation between loop fields. In PDB-deposited mmCIF files a line starting with a "#" is used, in the phenix-produced files a blank line is used instead. In fact the mmCIF specification doesn't even say that the "#" are needed. Lines starting with hashes are simply comment lines.

See http://mmcif.wwpdb.org/docs/tutorials/mechanics/pdbx-mmcif-syntax.html

Issue #188, opened 2014-10-06
Implement mmCIF file writing


josemduarte:
I couldn't find a way to write structural info in mmCIF format at the moment (please correct me if I'm wrong!). It would be cool to have a mmCIF writer to complete the support for mmCIF format.

See also #173

PR #279, opened 2015-06-09
Improving mmCIF support

josemduarte:
This should solve issues #188 and #234 
- basic mmcif writing: coordinates and crystallographic info
- any data in one of the beans in package `org.biojava.nbio.structure.io.mmcif.model` can be written to mmcif format using `org.biojava.nbio.structure.io.MMCIFFileTools.toMMCIF()`, which uses java reflection
- many fixes to the parser so that it can parse a broader set of files (including phenix-produced ones and biojava-produced ones!)
- all biojava-structure tests pass. The long parsing test in biojava-integrationtest also passes


andreasprlic:
Great!! Did not know you were working on this!

josemduarte:
It just came up in the last days... out of a sudden need to represent bio assembly predictions with something better than multi-model pdb files ;)

PR #291, opened 2015-06-26
Combine master and minor branches

sbliven:
I propose we merge all changes from master into minor. Other than the version number commits, I believe all of the commits are suitable for the 4.1.x line, rather than 5.0.0. I think that any differences are simply from pushing to the wrong branch.


sbliven:
I confirmed that the merge only changes some GenBank stuff.

I've manually merged on my fork & am testing now.

sbliven:
Note that this was writen as a 'into minor from master' merge since minor is still the main development branch. However, the intent is that master will be fast forwarded immediately after and minor & patch will be deleted.

Issue #282, opened 2015-06-17
Endless loop when using FastaReader


andreasprlic:
Toorn reported on mailing list that he encountered an endless loop with one of his fasta files. I can reproduce this.

andreasprlic:
also added documentation for this to tutorial

https://github.com/biojava/biojava-tutorial/blob/master/core/readwrite.md

aalhossary:
Actually there is no bug there. The function documentation never mentioned that a null might be returned.
The proper way of using the function is 

            do {
                b = fastaReader.process(10);
                for (String seq : b.keySet()) {
                    System.out.println(seq);
                }
            } while (b.size() == 10);
It has been there for a while and others may have been using it

Shall we revert this commit?

andreasprlic:
Please try the test case, it did not work before the patch.

aalhossary:
I updated the function and the test case

 

There are two different issues here:

 

1)      The functions fails because it was not designed to never return null at all, that’s why falls into an infinite loop. Instead it returns a LinkedHashMap with size in the in the range [0, max], where size < max indicates last (probably incomplete) read batch of sequences.

 

It should not be called within a 

while ((ret = process(int)) != null){/*code*/}

loop

 

Instead it should be called within a 

do{ret = process(int); /*code*/} while ( ret.size() < max);

loop

 

2)      If there is a blank line after the last sequence of the last batch, it fails to interpret it as either a new sequence or a closed stream. I corrected this by passing any blanl lines, and adding an EOF flag.

 

P.S. I made a pull request but I am not sure it will catch the new release.

andreasprlic:
not sure why this one is still open, this one got fixed already for last release.

PR #292, opened 2015-06-27
reverting #282 to never return null, managed blank lines anywhere

aalhossary:
reverting #282 to never return null, managed presence of blank lines
anywhere in the file, and updated test cases.

Signed-off-by: aalhossary <amr_alhossary@hotmail.com>


andreasprlic:
not sure why you want to revert #282, there clearly was a bug there?

sbliven:
Can we close this? Both the size-check pattern and the null-check pattern work fine now. No need to revert.

PR #298, opened 2015-07-13
created tool for fetching a list of uniprotKb entries in multiple formats

parit:
Recently, there has been multiple changes in the programmatic access to uniprotkb website. 2 main among them:
1. Fetching a list of uniprotkb entries across multiple format
2. Mapping a list of external database ids to uniprotkb entries.
There is already a way to fetch a single uniprotkb entry sequence in org.biojava.nbio.core.sequence.loader.UniprotProxySequenceReader but no way to deal with a list of entries. This commit intends to deal with point 1. point 2 will be in an upcoming fix.

### DIFF
--- a/biojava-ws/pom.xml
+++ b/biojava-ws/pom.xml
@@ -16,6 +16,11 @@
 			<scope>compile</scope>
 		</dependency>
 		<dependency>
+		  <groupId>org.apache.httpcomponents</groupId>
+		  <artifactId>httpclient</artifactId>
+		  <version>4.5</version>
+		</dependency>		
+		<dependency>
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
 			<scope>test</scope>
--- a/biojava-ws/src/main/java/demo/UniprotKbBatchRetrievalDemo.java
+++ b/biojava-ws/src/main/java/demo/UniprotKbBatchRetrievalDemo.java
@@ -1,0 +1,99 @@
+package demo;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.Reader;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.Scanner;
+
+import org.biojava.nbio.ws.uniprot.BatchRetrieve;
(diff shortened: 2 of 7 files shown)


andreasprlic:
Hi Parit,

Thanks for your pull request! This is clean code to fetch multiple UniProt files from their website. However there is actually not much functionality contained here in terms of parsing data out of the files.  The code to download files is rather generic and could be easily made more general to download any sets of files from any location. As such I feel it might be better to keep this on the side of an application and stay out of BioJava. 

Taking this also in another direction, it would be nice to have some code that can maintain a (complete) local installation of UniProt (and/or Trembl).

josemduarte:
Regarding UniProt remote querying, there is a very good Java API to remotely access UniProt data: http://www.ebi.ac.uk/uniprot/remotingAPI . Actually they have just released a whole new version which is now in beta. In any case I think that should be the preferred way to access UniProt annotations remotely. 

I agree with @andreasprlic that it would be a good idea to have some kind of code that could install a local version of UniProt.

parit:
@josemduarte  @andreasprlic  thankx for the feedback. I definitely see your point. 
As per providing a local install of uniprot data is concerned I think there are far more 'non-technical' issues associated with that. Anyhow I shall have a discussion here about the same and lets see what comes out of it. 
Thankx for the feedback again!!

andreasprlic:
Just to bring this up in this context: I have been working on a piece of software that can load UniProt XML files locally in a database, which I am going to open-source soon.

sbliven:
Can we close this pending the improvements discussed?

phidias51:
FYI, I have some code in the biogroovy library that reads uniprotkb records
and returns lists of Protein objects that can be persisted into a database.
I currently use it to populate some databases in several Grails projects.

There's also support for a wide variety of other model objects and file
formats, including MyGene.info, EntrezGene, PubMed to name a few. You can
use biogroovy as a Java library, or use it as a scripting library.

Regards,

Mark Fortner

andreasprlic:
There has been no follow-up in half a year. I am closing this for now.

andreasprlic:
see also https://github.com/rcsb/uniprot-or-mapping

Issue #301, opened 2015-07-16
For a given Feature allow multiple Qualifier values per Qualifier key


zachcp:
Hello Team BioJava,

Thanks for the great library! 

I have recently started to use BioJava for parsing Genbank files and I noticed a difference between BioJava and BioPython in their behavior that leads me to propose that BioJava change the `Qualifiers` slot in `org.biojava.nbio.core.sequence.features.AbstractFeature`  from `HashMap<String,Qualifier>` to `HashMap<String, ArrayList<Qualifier>>`.  Here's why:

In Biopython a `Feature` can have multiple `Qualifier` values for the same key. This is great for Genbank parsing a file like the one below where Features have multiple entries for the same key. Here are two examples: 

```
#Example 1 taken from http://mibig.secondarymetabolites.org/
CDS             5287..14826
                     /product="SalB"
                     /codon_start=1
                     /transl_table=11
                     /note="salinomycin polyketide synthase type I"
                     /note="smCOG: SMCOG1022:Beta-ketoacyl_synthase (Score:
                     285.5; E-value: 1.2e-86);"
                     /db_xref="GI:108743979"
                     /sec_met="Type: t1pks"
                     /sec_met="Domains detected: PP-binding (E-value: 3e-13,
                     bitscore: 39.3, seeds: 164); PP-binding (E-value: 1.2e-14,
                     bitscore: 43.8, seeds: 164); PKS_AT (E-value: 8.7e-106,
                     bitscore: 342.7, seeds: 1682); PKS_AT (E-value: 4.4e-109,
                     bitscore: 353.5, seeds: 1682); mod_KS (E-value: 1.4e-216,
                     bitscore: 708.3, seeds: 217); mod_KS (E-value: 1.6e-219,
                     bitscore: 718.0, seeds: 217); adh_short (E-value: 5.3e-57,
                     bitscore: 181.7, seeds: 230); adh_short (E-value: 1.8e-54,
                     bitscore: 173.5, seeds: 230)"
                     /sec_met="Kind: biosynthetic"
                     /sec_met="NRPS/PKS subtype: Type I Modular PKS"
                     /sec_met="NRPS/PKS Domain: PKS_KS (36-450). E-value:
                     1.8e-175. Score: 575.2;"
                     /sec_met="NRPS/PKS Domain: PKS_AT (565-853). E-value:
                     4e-105. Score: 342.7;"
                     /sec_met="NRPS/PKS Domain: PKS_KR (1193-1367). E-value:
[...]

PR #302, opened 2015-07-17
Change AbstractFeature Qualifiers to a key:Array from key:value

zachcp:
This is a pull request for reference #301 

The basic idea is to allow multiple key/values when parsing genbank files. That is the core change. The rest of the changes are to update getting/setting the fields and to modify the various places this code is called to handle the new format. Also add a test. This passes `mvn test`
 on my machine. Happy to modify in response to feedback.

zach cp

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/sequence/features/AbstractFeature.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/sequence/features/AbstractFeature.java
@@ -47,7 +47,7 @@ public abstract class AbstractFeature<S extends AbstractSequence<C>, C extends C
     private String description = "";
     private String shortDescription = "";
     private Object userObject = null;
-	private HashMap<String,Qualifier> Qualifiers = new HashMap<String,Qualifier>();
+	private HashMap<String, ArrayList<Qualifier>> Qualifiers = new HashMap<String, ArrayList<Qualifier>>();
 
     /**
      * A feature has a type and a source
@@ -271,13 +271,13 @@ public abstract class AbstractFeature<S extends AbstractSequence<C>, C extends C
     }
     
 	@Override
-	public HashMap<String, Qualifier> getQualifiers() {
+	public HashMap<String, ArrayList<Qualifier>> getQualifiers() {
 		// TODO Auto-generated method stub
 		return Qualifiers;
 	}
 
 	@Override
-	public void setQualifiers(HashMap<String, Qualifier> qualifiers) {
+	public void setQualifiers(HashMap<String, ArrayList<Qualifier>> qualifiers) {
 		// TODO Auto-generated method stub
 		Qualifiers = qualifiers;
 		
@@ -285,8 +285,17 @@ public abstract class AbstractFeature<S extends AbstractSequence<C>, C extends C
 
 	@Override
 	public void addQualifier(String key, Qualifier qualifier) {
-		// TODO Auto-generated method stub
-		Qualifiers.put(key, qualifier);
+		// Check for key. Update list of values
+        if (Qualifiers.containsKey(key)){
+            ArrayList<Qualifier> vals = Qualifiers.get(key);
+            vals.add(qualifier);
+            Qualifiers.put(key, vals);
+        } else {
+            ArrayList<Qualifier> vals = new ArrayList<Qualifier>();
+            vals.add(qualifier);
+            Qualifiers.put(key, vals);
+        }
 		
 	}
+
 }
--- a/biojava-core/src/main/java/org/biojava/nbio/core/sequence/ProteinSequence.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/sequence/ProteinSequence.java
@@ -109,7 +109,7 @@ public class ProteinSequence extends AbstractSequence<AminoAcidCompound> {
 
         // cases if a protein has more than 1 parent are not supported yet
         if (CDSFeatures.size() == 1) {
-            Qualifier codedBy = CDSFeatures.get(0).getQualifiers().get("coded_by");
+            Qualifier codedBy = CDSFeatures.get(0).getQualifiers().get("coded_by").get(0);
 
             if (codedBy != null) {
                 String codedBySeq = codedBy.getValue();
--- a/biojava-core/src/main/java/org/biojava/nbio/core/sequence/features/FeatureDbReferenceInfo.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/sequence/features/FeatureDbReferenceInfo.java
@@ -49,7 +49,7 @@ public class FeatureDbReferenceInfo<S extends AbstractSequence<C>, C extends Com
     private String description = "";
     private String shortDescription = "";
     private Object userObject;
-    private HashMap<String,Qualifier> qualifiers = new HashMap<String,Qualifier>();
+    private HashMap<String,ArrayList<Qualifier>> qualifiers = new HashMap<String,ArrayList<Qualifier>>();
     
     
     public FeatureDbReferenceInfo(String database, String id) {
@@ -137,22 +137,30 @@ public class FeatureDbReferenceInfo<S extends AbstractSequence<C>, C extends Com
     }
 
     @Override
-    public HashMap<String, Qualifier> getQualifiers() {
+    public HashMap<String, ArrayList<Qualifier>> getQualifiers() {
         return qualifiers;
     }
 
     @Override
-    public void setQualifiers(HashMap<String, Qualifier> qualifiers) {
+    public void setQualifiers(HashMap<String, ArrayList<Qualifier>> qualifiers) {
         this.qualifiers = qualifiers;
     }
 
     @Override
     public void addQualifier(String key, Qualifier qualifier) {
         if (qualifiers == null) {
-            qualifiers = new HashMap<String, Qualifier>();
+            qualifiers = new HashMap<String, ArrayList<Qualifier>>();
         }
-        
-        qualifiers.put(key, qualifier);
+        // Check for key. Update list of values
+        if (qualifiers.containsKey(key)){
+            ArrayList<Qualifier> vals = qualifiers.get(key);
+            vals.add(qualifier);
+            qualifiers.put(key, vals);
+        } else {
+            ArrayList<Qualifier> vals = new ArrayList<Qualifier>();
+            vals.add(qualifier);
+            qualifiers.put(key, vals);
+        }
+
     }
-    
 }
--- a/biojava-core/src/main/java/org/biojava/nbio/core/sequence/features/FeatureInterface.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/sequence/features/FeatureInterface.java
@@ -29,6 +29,7 @@ import org.biojava.nbio.core.sequence.location.template.AbstractLocation;
 import org.biojava.nbio.core.sequence.template.AbstractSequence;
 import org.biojava.nbio.core.sequence.template.Compound;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 
@@ -175,14 +176,14 @@ public interface FeatureInterface<S extends AbstractSequence<C>, C extends Compo
      * @return
      */
 
-    public HashMap<String, Qualifier> getQualifiers();
+    public HashMap<String, ArrayList<Qualifier>> getQualifiers();
 
     /**
      * Set the qualifiers
      * @param qualifiers
      */
 
-    public void setQualifiers(HashMap<String, Qualifier> qualifiers);
+    public void setQualifiers(HashMap<String, ArrayList<Qualifier>> qualifiers);
     /**
      * Add a qualifier
      * @param qualifier
--- a/biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericInsdcHeaderFormat.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericInsdcHeaderFormat.java
@@ -121,8 +121,10 @@ public class GenericInsdcHeaderFormat<S extends AbstractSequence<C>, C extends C
 		formatter.close();
 		
 		//Now the qualifiers...
-		for(Qualifier  q : feature.getQualifiers().values()) {
-			line += _write_feature_qualifier(q.getName(), q.getValue(), q.needsQuotes());
+		for(ArrayList<Qualifier>  qualifiers : feature.getQualifiers().values()) {
+			for(Qualifier q : qualifiers){
+				line += _write_feature_qualifier(q.getName(), q.getValue(), q.needsQuotes());
+			}
 		}
 		return line;
 		/*
--- a/biojava-core/src/main/java/org/biojava/nbio/core/sequence/template/AbstractSequence.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/sequence/template/AbstractSequence.java
@@ -129,7 +129,10 @@ public abstract class AbstractSequence<C extends Compound> implements Sequence<C
                 }
             }
             // success of next statement guaranteed because source is a compulsory field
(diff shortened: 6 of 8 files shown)


biojava-core/src/main/java/org/biojava/nbio/core/sequence/features/AbstractFeature.java
```diff
@@ -271,13 +271,13 @@ public void setUserObject(Object userObject) {
     }
     
 	@Override
-	public HashMap<String, Qualifier> getQualifiers() {
+	public HashMap<String, ArrayList<Qualifier>> getQualifiers() {
```
heuermh: Since this is a breaking change, I would recommend changing the method signature to
```java
public ListMultimap<String, Qualifier> getQualifiers() { ... }
```
where [ListMultimap](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/ListMultimap.html) comes from the [Guava](https://code.google.com/p/guava-libraries/) library.

biojava-core/src/main/java/org/biojava/nbio/core/sequence/features/AbstractFeature.java
```diff
@@ -285,8 +285,17 @@ public void setQualifiers(HashMap<String, Qualifier> qualifiers) {
 
 	@Override
 	public void addQualifier(String key, Qualifier qualifier) {
-		// TODO Auto-generated method stub
-		Qualifiers.put(key, qualifier);
+		// Check for key. Update list of values
```
heuermh: e.g. this kind of stuff isn't necessary with implementations of ListMultimap, you just put the key, value and it uses a List for the value

biojava-core/src/main/java/org/biojava/nbio/core/sequence/features/AbstractFeature.java
andreasprlic: The core-module does currently not have any dependencies on external libraries and I'd prefer to keep it that way. Only the genomics module uses Guava. as such my vote is for 
```java
public Map<String, List<Qualifier>> getQualifiers() { ... }
```

(note: use of interfaces, not specific implementations)

heuermh: Fair enough.

The sequencing module depends on Guava too https://github.com/biojava/biojava/blob/master/biojava-sequencing/pom.xml#L15

biojava-core/src/main/java/org/biojava/nbio/core/sequence/features/AbstractFeature.java
zachcp: I am happy to go either way. However, when I try to use the generic `Map` and`List` I run into errors like the following: 

screen shot 2015-07-20 at 11 59 11 am
.

So I am not really sure how to implement the changes as you suggest.

andreasprlic:
That is coming from the FeatureInterface via inheritance.  Not sure why we don't use the interface name in that Interface, must be an oversight...

andreasprlic:
This has been dormant for half a year. I think it would be a great patch to have. However there is still a tiny modification necessary to get this to completion. Any takers to wrap this one up?

zachcp:
Hi @andreasprlic,

I tried to implement this along the lines you suggested but I kept getting tangled up by the inheritance problem. It would be nice if a more experienced Java programmer could put some eyeballs on this.

zach cp

zachcp:
thanks @emckee2006 , @josemduarte , @andreasprlic . I didn't see  PR #316 until just now. 

Looks good on a basic GBK file. I can try it on a few others over the next few days. Thanks.

stefanharjes:
This patch breaks a lot of my code and I have a problem understanding the underlying cause. 
In the summary you claim "The basic idea is to allow multiple key/values when parsing genbank files."

In the genbank format a feature like *Source* or *CDS* have lots of qualifiers such as  /organism="Saccharomyces cerevisiae"
                     /db_xref="taxon:4932"
                     /chromosome="IX"
                     /map="9"
     CDS             <1..206
                     /codon_start=3
                     /product="TCP1-beta"
                     /protein_id="AAA98665.1"
                     /db_xref="GI:1293614"
                     /translation

zachcp:
Stefan,

This is definitely a breaking change and it involves a tradeoff between flexibility and convenience, as you suggest. The road that the BioPython community has taken is to have the qualifiers field be a key/value store that can hold arbitrarily many items. While this was annoying at first, I came to realize the utility when dealing with files where protein-specific information  - dbrefs, extra attributes and so on - were stored in these fields. While this flexibility may not be necessary for most of the entries derived from  Genbank itself, I have parsed many GBK files from various editors that have stored feature-information in these fields - I have given one example in issue #301 or #302.  A lack of this feature effectively means loss of information upon parsing a file which is a showstopper for my use of BioJava.

I also think multiple values may be more common than you think. Below is an example from one of the BioPython core developers walking through a GenBank [parsing example](http://www2.warwick.ac.uk/fac/sci/moac/people/students/peter_cock/python/genbank/). Note that the db_xref dictionary has two values. Parsing this in BioJava would previously only return one of the values. 

```
>>> gb_feature = gb_record.features[26]
>>> print gb_feature
type: CDS
location: (7397..8423)
ref: None:None
strand: -1
qualifiers: 
	Key: codon_start, Value: ['1']
	Key: db_xref, Value: ['GI:41614806', 'GeneID:2654552']
	Key: locus_tag, Value: ['NEQ010']
	Key: product, Value: ['hypothetical protein']
	Key: protein_id, Value: ['NP_963304.1']
	Key: transl_table, Value: ['11']
	Key: translation, Value: ['MENREGIIEFVKEAYPITKGKYRIPLAVTEFFIEKEKLKEIEEK
LNAIFLRIKEIVFEFLDNSLKGYPIDEFRKNLYLKIYENYNYKFKGIIRFDMINFKLIEINADAVEGVIQLDITSKWFS
ELFDFVGEPRYNREPFLSLFDKPAIIVYPDNYLFVDEYYLESKLLKGPYIKESEFYENKKYRIYPIYRRAIDIDKVKKL
DYWLEKQNINDILFTIMGNKANLAFLDKEVDWALDIVAKTSFKPEFKKYVCKPYFGTHGEVYFEKRDNCVYQEFIEIPK
REISYLDIDDKIKKGIFYYDFNPYAFFDKKLRFGNILLRFSKDHILNVAKGGGVGFYTIV']
```
Certainly, I think the flexibility is worth the inconvenience but I see your point about breaking code. I will leave this decision to the core devs.

zach cp

stefanharjes:
Hi Zach,

As a use case you should consider loading a couple bacterial genomes each with several thousand qualifiers which are now not qualifiers but lists of qualifiers and see the drop in performance (which I expect, but have not measured). I am not so much worried about flexibility or convenience, but of performance which is pretty bad already (I have a database of about 50 (Haha) sequences and for some operations I can go to lunch five times a day).

As I see in the BioPython example you also do not want a List of Qualifiers, but a Qualifier with one key which stores an array as value. That is a totally different shoe and would also not be such a load, as Strings are arrays anyhow. This would also prevent the overhead to check that all qualifiers in  a particular list have the same keys!  Obviously the reader/writer will have to io the key several times, but simply using a list<Qualifers> is a pretty dirty hack. 

Regards
Stefan

p.s. Your sec_met example refers to a sequence feature in the CDS while also storing its own position in the main sequence. If you would define it as a feature in itself, you could enter the feature location as a simple location and then you could define as many of them as you want while having normal access to its location. which now you would have to pull out of a String.

zachcp:
Hi Stefan,

I believe there is a miscommunication. Qualifiers remains a map  but puts the values into a list (an ArrayList). This is the simplest way to allow for multiple values for the same key and it is the way that at least one other Bio- project deals with the problem. As you point out, it should not be such a performance load:

>As I see in the BioPython example you also do not want a List of Qualifiers, but a Qualifier with one       key which stores an array as value. That is a totally different shoe and would also not be such a load, as Strings are arrays anyhow. 

As for the particular example I have shown it is only one file and it was not generated by me. The basic issue is that the GenBank file spec allows some flexibility and unless all of the editors/parsers agree on where/how to store information there will be difference in the files you encounter. For annotations on a protein I commonly find that the annotations adorn the CDS feature and not the master sequence record. But if you come across one of these files and parse it, you shouldn't lose information.

zach cp

p.s. Out of curiosity, what are the sorts of operations you are performing on your bacterial genomes?

andreasprlic:
@stefanharjes Can you share more details into what you are doing, perhaps provide some code? Then we might be able to help with some comments on where the performance bottlenecks are.

stefanharjes:
Hi biojava,

I submitted a pull request for an improved qualifer. The commits changes a list of qualifiers to one qualifier which holds a string[] of values. Before in the list of qualifiers each qualifier had its own name, which was assumed to be the same, but never checked. I removed this pitfall and implemented each qualifier to be able to hold lots of values under one name. For this I implemented a new Map.
As this breaks the former interface I concentrated on keeping things as seemless as possible. Tell me if you find any bumps. 

Altogether I think that the code is now much shorter and readability, performance and storage should be a lot better.

@andreasprlic I think the main time consumer in my java servlet are the pairwise and multiple alignments.  I feel that the cause for this is the implementation of several physical constants. Correct me if I am wrong, but I think the nucleotides are string variables and the string comparison is noted in the biojava source as being horribly expensive. Also chemical elements are variable classes? Such constants should be implemented as java constants. Here I think primarily of enums. If I find a student or steal some time from my experimental work, maybe I have a look. Otherwise it would be great to put up a feature request to implement chemical constants as enums.

Regards

pwrose:
FYI, there is an Element enum here:
https://github.com/biojava/biojava/blob/master/biojava-structure/src/main/java/org/biojava/nbio/structure/Element.java

I'm not sure that is useful for this application and it would create a
dependency on the structure package.

-Peter

On Thu, Mar 3, 2016 at 12:44 AM, Stefan Harjes <notifications@github.com>
wrote:

> Hi biojava,
>
> I submitted a pull request for an improved qualifer. The commits changes a
> list of qualifiers to one qualifier which holds a string[] of values.
> Before in the list of qualifiers each qualifier had its own name, which was
> assumed to be the same, but never checked. I removed this pitfall and
> implemented each qualifier to be able to hold lots of values under one
> name. For this I implemented a new Map.
> As this breaks the former interface I concentrated on keeping things as
> seemless as possible. Tell me if you find any bumps.
>
> Altogether I think that the code is now much shorter and readability,
> performance and storage should be a lot better.
>
> @andreasprlic <https://github.com/andreasprlic> I think the main time
> consumer in my java servlet are the pairwise and multiple alignments. I
> feel that the cause for this is the implementation of several physical
> constants. Correct me if I am wrong, but I think the nucleotides are string
> variables and the string comparison is noted in the biojava source as being
> horribly expensive. Also chemical elements are variable classes? Such
> constants should be implemented as java constants. Here I think primarily
> of enums. If I find a student or steal some time from my experimental work,
> maybe I have a look. Otherwise it would be great to put up a feature
> request to implement chemical constants as enums.
>
> Regards
>
> —
> <https://github.com/biojava/biojava/pull/302#issuecomment-191655720>.
>



-- 
Peter Rose, Ph.D.
Site Head, RCSB Protein Data Bank West (http://www.rcsb.org)
San Diego Supercomputer Center (http://bioinformatics.sdsc.edu)
University of California, San Diego
+1-858-822-5497

stefanharjes:
Hi Peter,
thanks I only had a short glimpse at the structure module and will look at Element. 

Also I was not precise enough in my comment. I looked at the Compound.java type of classes (in core/sequence/compound). Most fields are indeed defined as final, and thus can not be changed. However, If I read correctly, the actual compounds are stored as strings. Also all comparisons are string.equals(string) comparisons which are quite expensive computationally?  

Now I observed: when I compare the computational time consumed by a biojava pairwise alignment to the c/c++ emboss needle for the same alignment I have a score of 18.1 sec versus 6.9 sec. This in itself is probably not so bad, as C is usually faster than Java. However I thought that the number of comparisons in a pairwise alignment is probably pretty large and thus the main cost might be the String comparison? 

Thus, if I store nucleotides/amino acids in a enum type, where I more or less store the constant as integer, later comparisons will be integer:integer comparisons, which should be much faster?
Regards 
Stefan




 

    Peter Rose <notifications@github.com> schrieb am 6:32 Freitag, 4.März 2016:
 

 FYI, there is an Element enum here:
https://github.com/biojava/biojava/blob/master/biojava-structure/src/main/java/org/biojava/nbio/structure/Element.java

I'm not sure that is useful for this application and it would create a
dependency on the structure package.

-Peter

On Thu, Mar 3, 2016 at 12:44 AM, Stefan Harjes <notifications@github.com>
wrote:

> Hi biojava,
>
> I submitted a pull request for an improved qualifer. The commits changes a
> list of qualifiers to one qualifier which holds a string[] of values.
> Before in the list of qualifiers each qualifier had its own name, which was
> assumed to be the same, but never checked. I removed this pitfall and
> implemented each qualifier to be able to hold lots of values under one
> name. For this I implemented a new Map.
> As this breaks the former interface I concentrated on keeping things as
> seemless as possible. Tell me if you find any bumps.
>
> Altogether I think that the code is now much shorter and readability,
> performance and storage should be a lot better.
>
> @andreasprlic <https://github.com/andreasprlic> I think the main time
> consumer in my java servlet are the pairwise and multiple alignments. I
> feel that the cause for this is the implementation of several physical
> constants. Correct me if I am wrong, but I think the nucleotides are string
> variables and the string comparison is noted in the biojava source as being
> horribly expensive. Also chemical elements are variable classes? Such
> constants should be implemented as java constants. Here I think primarily
> of enums. If I find a student or steal some time from my experimental work,
> maybe I have a look. Otherwise it would be great to put up a feature
> request to implement chemical constants as enums.
>
> Regards
>
> —
> <https://github.com/biojava/biojava/pull/302#issuecomment-191655720>.
>



-- 
Peter Rose, Ph.D.
Site Head, RCSB Protein Data Bank West (http://www.rcsb.org)
San Diego Supercomputer Center (http://bioinformatics.sdsc.edu)
University of California, San Diego
+1-858-822-5497

josemduarte:
I agree that Compound stored as a String is a waste of resources: not only compute-wise as @stefanharjes points out but also memory-wise. From your benchmark against needle, indeed it looks like it is not very efficient computationally. C should have a bit of an edge versus Java but only a marginal one. If the Java implementation was optimal, we wouldn't see that enormous difference in runtime.

I also agree with you that the best way to implement this is with an enum. By the way the `Element` enum in the structure module is about atoms not about aminoacid/nucleotides which is what we have in the Compound case here. 

In any case perhaps an easy optimisation for now would be to simply convert the field `base` in `AbstractCompound` from String into a char. That is if we are sure that biojava's implementation a Compound can't be more than 1 character long.

paolopavan:
Interesting point. In this short lasting tests, consider also that Java has
a considerable startup time to instantiate virtual machine that the native
executables have not. This usually isn't an issue in real application but
biases those kind of evaluations.

2016-03-08 7:22 GMT+01:00 Jose Manuel Duarte <notifications@github.com>:

heuermh:
> I agree that Compound stored as a String is a waste of resources: not only compute-wise as @stefanharjes points out but also memory-wise.

We could always go back to the flyweight Symbol pattern as in biojava-legacy (1.x) ```<ducks>```
:)

stefanharjes:
Hi Paolo,
regarding the startup time it is the other way around in fact. 
I measured the java time in the code by using:long startTime = System.currentTimeMillis();

while with emboss needle I used the bash command: "time needle ...." So the start of the java vm is not included, but the bash overhead is included (which I assumed to be negligible)...

Regards
Stefan

 

    Paolo Pavan <notifications@github.com> schrieb am 21:56 Dienstag, 8.März 2016:
 

 Interesting point. In this short lasting tests, consider also that Java has
a considerable startup time to instantiate virtual machine that the native
executables have not. This usually isn't an issue in real application but
biases those kind of evaluations.

2016-03-08 7:22 GMT+01:00 Jose Manuel Duarte <notifications@github.com>:

Issue #313, opened 2015-08-04
ChemCompGroupFactory cache not being reset properly


lafita:
This issue is related to the closed issue #145, about a problem with singletons in the structure module. The fix145 made sure that the cache is reset if the `ChemCompProvider` is set (changed).

However there is another problem: when the method `setLoadChemCompInfo()` of the `FileParsingParameters` is set, neither the cache is cleaned or the `ChemCompProvider` is changed in the `ChemCompGroupFactory`. That generated unexpected results if the `ChemCompProvider` was set previously and the parameters are changed afterwards.

The fix will need to reset the `ChemCompProvider` if the parameters are changed, i.e. every time `setLoadChemCompInfo()` is called.

lafita:
After revising the code with @josemduarte we realized that the real problem was a bug in  `PDBFileReader` that was setting the static variables of `ChemCompGroupFactory` inconsistently. This bug has been fixed, and the other parts of the code behave the same way.

This is closely related to issue #169, because the user has to specify the `ChemCompProvider` he/she needs, independently of the parser used. Because it is a static variable, the same default will apply to MMCiff and PDB parsers.

josemduarte:
Just a note on that: it seems that as @andreasprlic says in #169, the PDB parser sets the `ChemCompProvider` intentionally so that it is faster and more streamlined (versus the more correct but slower mmCIF parser). That in principle is fine, but before the patches above, the `ChemCompProvider` was setting the singleton and affecting the whole JVM globally. So that affected all parsers and would result in unexpected side effects. 

With these patches, in order to produce the "quick and dirty" PDB parsing behaviour that we had before, the user would need to explicitly set it like this:

```java
ChemCompGroupFactory.setChemCompProvider(new ReducedChemCompProvider());
```

So we should clarify that in documentation, but otherwise the functionality we had before would still be available.

lafita:
Regarding the Test created for https://redmine.open-bio.org/issues/3334, it does not apply anymore because the value of the variable *loadChemCompInfo* does not change the static *ChemCompProvider*, which is valid for the whole JVM. The user needs to change the Provider explicitly as @josemduarte explained.

andreasprlic:
Can we keep the ```ReducedChemCompProvider``` the default provider, unless the user specifies a different one?

Issue #309, opened 2015-08-03
Get representative Atoms from Structure different than from Chain


lafita:
Calling: <pre>StructureTools.getRepresentativeAtoms(Structure s)</pre> Gives a different result than calling:
<pre>StructureTools.getRepresentativeAtoms(Chain c)</pre>
The first call returns an array of size 100 and the second an array of size 98 for 1hiv.A. The first method returns two Atoms more, corresponding to a modified residue CSO, in the middle of the array, and the ligand, at the end of the array. The ligand is 1ZK and has an Atom called CA.

This difference between the methods has caused some IndexOutOfBoundsExceptions when manipulating Atoms of a Structure. A solution could be using the chain method from the structure call. A Test and some test cases will be needed to make sure they are consistent.

lafita:
Another issue popped up: the ChemComp information is not being loaded correctly for HETATMs that are modified aminoacids (loaded as nulls and unknowns), like CSO or LLP, so they are not returned as part of the representative Atom array. This breaks some of the existing tests.

andreasprlic:
What ChemCompProvider are you using?

lafita:
The problem was in internal dependencies of the test. The cache for the *ChemCompGroupFactory* was not reset properly when the *ChemCompProvider* was reset. I will create another issue regarding that and propose a fix. The differences between representative Atoms of chains and structures is fixed now.

PR #314, opened 2015-08-04
Fix313 - ChemCompProvider and ChemComp cache consistency

lafita:
See issue #313.

This pull request solves issue 313 by forcing the user to specify which *ChemCompProvider* is used when parsing a PDB or MMCiff file. The cache is kept and increased while the JVM is running, so *ChemComp* downloaded and cached by the *DownloadChemCompProvider* can also be user later by the *ReducedChemCompProvider*. The cache is never reset in the same JVM.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/io/FileParsingParameters.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/io/FileParsingParameters.java
@@ -183,9 +183,9 @@ public class FileParsingParameters implements Serializable
 		return loadChemCompInfo;
 	}
 
-	/** 
-	 * Sets if chemical component defintions should be loaded from the web.
-	 * Modifies also the static ChemCompProvider and resets its cache.
+	/** Sets if chemical component defintions should be loaded or not.
+	 * The decision from where the definitions are obtained is
+	 * in the static variable inside {@link ChemCompGroupFactory}. 
 	 * 
 	 * @param loadChemCompInfo flag
 	 */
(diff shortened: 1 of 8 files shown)


andreasprlic:
I'd prefer to use ```ReducedChemCompProvider``` as the default, unless the user specifies differently. We should keep the code required to get going with BioJava to a minimum.

josemduarte:
That's a good point, it sounds logical that the `ReducedChemCompProvider` is the default one since it will give some minimal functionality that will work in any machine and with/without internet connection (in terms of code or complexity it is really not adding anything because it is all transparent to the user).

At the same time the `DownloadChemCompProvider` does a much better job in getting the right chemical descriptions for molecules and thus things work much better with it, so I think that's also a reason to keep that one by default.

So it's difficult to say what's the better default. We are going to try and check how the tests do when changing the default. That can give us some more clues of what things can go wrong.

lafita:
I think that there was no problem before in using the `ReducedChemCompProvider`, because the method `getRepresentativeAtomArray()` was not correctly implemented (issue #309). After #312, where this is fixed, to correctly obtain the Atoms in the backbone of a structure we need the information of the modified aminoacids, which is only possible with the `DownloadChemCompProvider`.
If the `ReducedChemCompProvider` is set and the structure contains modified aminoacids, those won't be returned when calling `getRepresentativeAtomArray()`.

I have tried to set the `ReducedChemCompProvider` as the default and run the tests. The only failures are when checking for array sizes after calling `getRepresentativeAtomArray()` on structures that have modified aminoacids. We have two options, either:
1. We change the failing tests to use the `DownloadChemCompProvider`, or
2. We modify the expected array lengths in the tests.

In any case we need to document in the `getRepresentativeAtomArray()` that modified aminoacids won't be returned as part of the backbone if the `ReducedChemCompProvider` is set and they were not cached before. I will do it now.

PR #325, opened 2015-09-16
Adding some documentation to the symmetry.core package

sbliven:
Adds some documentation. I don't completely understand this package, so a double-check from @pwrose would be appreciated.


sbliven:
This has been a while so I'm merging it.

Issue #111, opened 2014-04-03
Improving exception handling


josemduarte:
Exception handling in a few places in biojava-structure could be improved. I think it is a good idea to apply the practice of "throw early, catch late" (see for instance http://programmers.stackexchange.com/questions/231057/exceptions-why-throw-early-why-catch-late).

So as a very general rule there should be less catching and more throwing forward so that Exceptions are packed together and can be handled by the end-user (i.e. the application that makes use of biojava).

This is also a good read: http://wikijava.org/wiki/10_best_practices_with_Exceptions

Some things that should be avoided in my opinion:
- `catch (Exception e)` especially if deep in the library, it can hide NullPointerExceptions (or similars) and make debugging very difficult
- in tests it's also good to throw exceptions instead of catching them and failing, so that they show as errors and not as failures (of course this depends on the case)
- printing stack traces and re-throwing: the printing of stack traces should also be a responsibility of the final user, so that he/she has full control of verbosity of their output

In any case exception handling is always very difficult, but I think we can try to improve it a bit. This will surely require some changes in interface in cases where more Exceptions need to be thrown, so I'd propose we could include this in next release if there is agreement that we move forward to version 4.

andreasprlic:
I think the work on this has concluded for now and I am closing this one. Please re-open if I missed anything.

dmyersturnbull:
I'm reviving for now so that I can reference it in commits. A major release (e.g. Biojava 4) is probably best for these changes since they're all somewhat backwards-incompatible.

dmyersturnbull:
Closing for now, since that was all I had time for. There's still a lot of code that needs improvement, perhaps for Biojava 5.

PR #329, opened 2015-09-18
SearchIO module

paolopavan:
Dear all,
I am submitting changes largely discussed on the mailing list.
Since those changes introduces several variation to the API in terms of alignment package data structures that have moved to core, my advice is to merge this with next major version to warn all developers.

Bye bye,
Paolo


paolopavan:
I'm sorry guys, I'm now realizing that the structure module that actually I didn't touch it is failing.
I need help by its maintainers: does this module refers some resource as substitution matrices or some test files that previously were in alignment package? The test message is not so clear and seems that an InputStreamReader is failing.

emckee2006:
You might check that your patch didn't accidentally delete those (and other
non intended files)...

paolopavan:
Ok, did it! As supposed, it was a hard coded reference to a substitution matrix in a test file. Not so painful.
Now it is patched!

biojava-core/src/main/java/org/biojava/nbio/core/search/io/Hsp.java
```diff
@@ -0,0 +1,229 @@
...
+        String allInOne = hspQseq+"\n"+hspIdentityString+"\n"+hspHseq;
+        return allInOne.hashCode();
+    }
+    
+    @Override
+    public boolean equals(Object o){
+        if (!(o instanceof Hsp)) return false;
+        
+        return o.hashCode() == this.hashCode();
+    }
```
josemduarte: Two comments here:

- equals should not be implemented by comparing hashCodes. Hash codes do not guarantee equality. Hash equality is a necessary but not sufficient condition for equality (2 different objects can have the same hash)

- the best way to implement a hashCode is to follow the indications in [this SO thread](http://stackoverflow.com/questions/113511/best-implementation-for-hashcode-method). Actually if you are using eclipse it can even implement such a thing for you automatically, just go to 'Source' -> 'Generate hashCode and equals', you then only need to select which fields of the class should be used (I imagine other IDEs will have similar functions).

biojava-core/src/main/java/org/biojava/nbio/core/search/io/Result.java
```diff
@@ -0,0 +1,149 @@
...
+        }
+        return (prefix+suffix).hashCode();
+    }
+    
+    @Override
+    public boolean equals(Object o){
+        if (!(o instanceof Result)) return false;
+        
+        return o.hashCode() == this.hashCode();
+    }
```
josemduarte: Same comments as above about hashCode and equals. Try using the eclipse built-in function to generate them.

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLQuery.java
```diff
@@ -0,0 +1,270 @@
...
+    }
+    /**
+     * 
+     * @param maxEScore
+     * @return
+     * @throws Exception 
+     */
+    @Deprecated
+    public LinkedHashMap<String, ArrayList<String>> getHitsQueryDef(double maxEScore) throws Exception {
+        LinkedHashMap<String, ArrayList<String>> hitsHashMap = new LinkedHashMap<String, ArrayList<String>>();
```
josemduarte: If I understand it correctly you have copied the `org.biojava.nbio.genome.query.BlastXMLQuery` into `org.biojava.nbio.core.search.io.blast.BlastXMLQuery` and then added all the new parsing stuff to it. That's fine, but as you kept the original one intact (which I think is a good idea), there's no reason to keep these deprecated methods around. You can simply remove them if they are not in use. We really don't want a new class to be born with deprecated methods because that doesn't make much sense. Deprecations are to support API transitions, not for new parts of the API.

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLQuery.java
```diff
@@ -0,0 +1,270 @@
...
+
+        return hitsHashMap;
+    }
+    
+    private void readFile(String blastFile) throws Exception{
+        log.info("Start reading " + blastFile);
+        blastDoc = XMLHelper.loadXML(blastFile);
+        log.info("Read finished");
+    }
+
```
josemduarte: Try to avoid throwing generic Exceptions and be as specific as possible. When you catch this one below, you are catching not only the ones thrown by loadXML, but also any other potential Exception like a null pointer. That's not so bad but it can make debugging quite complicated. We are not very consistent with this in biojava yet, but we are trying to improve the situation, see #111

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLQuery.java
```diff
@@ -0,0 +1,270 @@
...
+            String program = XMLHelper.selectSingleElement(blastDoc.getDocumentElement(),"BlastOutput_program").getTextContent();
+            String version = XMLHelper.selectSingleElement(blastDoc.getDocumentElement(),"BlastOutput_version").getTextContent();
+            String reference = XMLHelper.selectSingleElement(blastDoc.getDocumentElement(),"BlastOutput_reference").getTextContent();
+            String dbFile = XMLHelper.selectSingleElement(blastDoc.getDocumentElement(),"BlastOutput_db").getTextContent();
+
+            log.info("Query for hits");
+            ArrayList<Element> IterationsList = XMLHelper.selectElements(blastDoc.getDocumentElement(), "BlastOutput_iterations/Iteration[Iteration_hits]");
+            log.info(IterationsList.size() + " results");
+
+            resultsCollection = new ArrayList();
```
josemduarte: This should be `new ArrayList<Result>`. In general try to parameterize classes that require them and keep the warnings to a minimum. There are a few easily avoidable warnings in this class and others (mostly related to generics parameterization).

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLQuery.java
```diff
@@ -0,0 +1,270 @@
...
+                databaseReferencesMap.put(id, databaseReferences.get(counter));
+            }  
+        }
+    }
+
+    @Override
+    public void storeObjects(List<Result> results) throws Exception {
+        throw new UnsupportedOperationException("This parser does not support writing yet.");
+    }
+}
```
josemduarte: No need to declare `throws Exception` here, `UnsupportedOperationException` is a runtime exception.

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLQuery.java
andreasprlic: something else: it is better to use interfaces rather than implementations. As such, it would be good to write this as something like this  ```List<Result> resultsCollection  = new ArrayList<Result>()```

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLQuery.java
paolopavan: Hi Jose,
I share what you say and in fact this was an attempt to do what you assert
in your #111, when you said "throw early, catch late" since there are 3
underlying exceptions there: SAXException, IOException,
ParserConfigurationException and my will was to pass them to the top module
level: SearchIO class.
Anyway this brought me to think and refine this handling because
SAXException and ParserConfigurationException are related to the underlying
xml and can be conceptually grouped in a more generic ParseException to be
forwarded to the top level.
See modifications.

2015-09-22 20:49 GMT+02:00 Jose Manuel Duarte <notifications@github.com>:

biojava-core/src/main/java/org/biojava/nbio/core/search/io/Hsp.java
paolopavan: I have changed the full stuff. I have revisited the way equality is
evaluated and I have modified properties accordingly.

2015-09-22 20:23 GMT+02:00 Jose Manuel Duarte <notifications@github.com>:

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLQuery.java
paolopavan: Yes, but it is defined by its interface and actual parsers will use it.
Anyway I chaged it in IOException + ParseException

2015-09-22 21:00 GMT+02:00 Jose Manuel Duarte <notifications@github.com>:

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLQuery.java
paolopavan: This doesn't make difference since the signature declared in the interface
for createObjects() is List<Results>. ArrayList<Result> is only an internal
variable, not exposed.

2015-09-23 0:07 GMT+02:00 Andreas Prlic <notifications@github.com>:

paolopavan: It is not compulsory since I previously have declared ArrayList<Result> resultsCollection. 
Sentences like resultsCollection.add(new String()) will already fail at compilation time. I  usually prefer it, because is less verbose and in my opinion cleaner. But it should be the same. 


2015-09-22 20:57 GMT+02:00 Jose Manuel Duarte <notifications@github.com>:

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLQuery.java
paolopavan: This actually escaped to me. Yes, the original class was
org.biojava.nbio.genome.query.BlastXMLQuery.
Since to an automatic usage search it seems to be poorly used, my original
will was to refactor this class moving from it's original position to
SearchIO while mantaining old method and declaring them obsolete.
Then caution reasons that you point out brought me to duplicate the class
and ask for advise to the mailing list.

My impression is that it is only a stub, to be further developed. If it is
so, we could just do as in my original plan because I have fear of
confusion it can generate since we are going to have new API for the task.
Otherwise I can clean my class as you are proposing.
What do you think?

2015-09-22 20:44 GMT+02:00 Jose Manuel Duarte <notifications@github.com>:

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLQuery.java
josemduarte: Alright now I see what you wanted to do. In that case, unfortunately you really need to catch the exception explicitly:
```java
try {
} catch (SAXException e) {
// do something
} catch (ParserConfigurationException e) {
// do the same thing again
}
```
If we were in Java 7 (we really need to upgrade soon!) we could do that much more elegantly:
```java
try {
} catch (SAXException|ParserConfigurationException e) {
// do something
}
```
If you don't do this kind of explicit catching then you are catching any other exception that occurs within that try block (anywhere in the call hierarchy within the block), e.g. NullPointer, OutOfBounds or any other thing. In general it is a better idea to let those go through because here you are only concerned on what to do when this particular XML parsing fails.

paolopavan: Yes I agree, new should already accomplish this, isn't it?

2015-10-08 21:17 GMT+02:00 Jose Manuel Duarte <notifications@github.com>:

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLQuery.java
josemduarte: Alright in that case I think it is a good idea to move it to SearchIO (removing the old one `org.biojava.nbio.genome.query.BlastXMLQuery`) and deprecating the methods as you did. Since as you say it seems to be only a stub and it isn't used much. 

If you go that way, one thing is important: whenever we deprecate things we should not use the deprecated methods within Biojava anymore (their usages would show as warnings). Deprecations are there to let other users know that we are moving forward and that in next versions the methods will be gone.

Another possibility can be simply leave the `org.biojava.nbio.genome.query.BlastXMLQuery` class as it is and don't touch it. Then create your own XML parser class with a different name in `search.io` package. I think that might be cleaner after all.

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLQuery.java
paolopavan: * was: "new version should already accomplish this", sorry

2015-10-08 21:35 GMT+02:00 Paolo Pavan <paolo.pavan@gmail.com>:

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLQuery.java
paolopavan: You mean that new code must not use deprecated method any more? Otherwise
this would mean to delete the classOutputHitsGFF in the same package
org.biojava.nbio.genome.query. Actually it seems to be safe at a first
glance. Author is Scooter Willis.

2015-10-08 21:36 GMT+02:00 Jose Manuel Duarte <notifications@github.com>:

josemduarte: Yes it would require removing other classes... and that's not ideal. That's why I'm thinking now that the best way to go is the last option I gave above. Leave all that `org.biojava.nbio.genome.query` code alone and create your own XML parser class in `search.io` with a different name so that things are clear. Both can then coexist. Yours will be the most complete and the one we should document for people to know how to deal with blast searches in Biojava.

paolopavan: Ok, I have separate the two stuffs in commit f059f0e. 
Anyway for the records, after having one more look at it I am pretty sure that the whole package `org.biojava.nbio.genome.query` with its two files can be deleted. In fact it appears to be parser that transform hits in a blast report to gff format, but actually the output generation is inhibited by a comment so none can use as it is at this moment.

In my opinion, it could be a good occasion for some cleaning; since there are already API changes that compell to a new major version and backward compatibility will not be guaranteed.

josemduarte: Thanks @paolopavan , the commit looks great! Regarding `org.biojava.nbio.genome.query` I'd be happy removing it if it really looks like dead code. But I wouldn't want to do something wrong, perhaps @willishf can shed some light on this?

paolopavan: Thank you Jose. Apart for further comments by @willishf that anyway does not impact on what is done any more, I think that the pull request is ready to be merged. Don't you?
Greetings!
Paolo

biojava-core/src/main/java/org/biojava/nbio/core/search/io/Hsp.java
```diff
@@ -0,0 +1,251 @@
...
+        
+        try {
+            if (sequenceString.matches("^[ACTG]+$")) 
+                returnSeq = new DNASequence(sequenceString, DNACompoundSet.getDNACompoundSet());
+            else if (sequenceString.matches("^[ACUG]+$"))
+                returnSeq = new RNASequence(sequenceString, DNACompoundSet.getDNACompoundSet());
+            else
+                returnSeq = new ProteinSequence(sequenceString, AminoAcidCompoundSet.getAminoAcidCompoundSet());
+        } catch (CompoundNotFoundException ex) {
+            Logger.getLogger(Hsp.class.getName()).log(Level.SEVERE, null, ex);
```
josemduarte: Logging in Biojava (since Biojava 4) is happening through the slf4j interface. You should change this to use slf4j. You simply need to import on top:
```java
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
```
then declare a static variable for the logger:
```java
private static final Logger logger = LoggerFactory.getLogger(Hsp.class);
```
and then use it like:
```java
logger.error("Unexpected error, could not find compound blah blah ", ex);
```
See for instance `org.biojava.nbio.structure.StructureImpl` for an example of the above.

biojava-core/src/main/java/org/biojava/nbio/core/search/io/Hsp.java
```diff
@@ -0,0 +1,251 @@
...
+        }
+        return returnSeq;
+    }
+    
+    private List<Step> getAlignmentsSteps(String gappedSequenceString){
+        List<Step> returnList = new ArrayList();
```
josemduarte: You are missing the type in ArrayList

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLParser.java
```diff
@@ -0,0 +1,232 @@
...
+        } catch (XPathException e) {
+            throw new ParseException(e.getMessage(),0);
+        }
+        log.info("Parsing of "+targetFile+" finished.");
+        
+        return resultsCollection;
+    }
+    
+    public List<String> getFileExtensions(){
+        ArrayList<String> extensions = new ArrayList(1);
```
josemduarte: Missing type

biojava-core/src/main/java/org/biojava/nbio/core/search/io/ResultFactory.java
```diff
@@ -0,0 +1,55 @@
...
+     * @throws Exception 
+     */
+    void storeObjects(List<Result> results) throws IOException, ParseException;
+    
+    /**
+     * Specify the collection of sequences objects used as queries in the Search run. 
+     * They will be associated back to the query during the construction of the Result object.
+     * @param sequences 
+     */
+    void setQueryReferences(List sequences);
```
josemduarte: Can the List here be parameterized?

biojava-core/src/main/java/org/biojava/nbio/core/search/io/ResultFactory.java
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Specify the collection of sequences objects used as database in the Search run. 
+     * They will be associated back to the Hit during the construction of the Hit object.
+     * @param sequences 
+     */
+    void setDatabaseReferences(List sequences);
```
josemduarte: Same about this List

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLParser.java
```diff
@@ -0,0 +1,232 @@
...
+ * 
+ * You may want to find my contacts on Github and LinkedIn for code info 
+ * or discuss major changes.
+ * https://github.com/paolopavan
+ * 
+ *
+ * @author Paolo Pavan
+ */
+public class BlastXMLParser implements ResultFactory {
+    private static final Logger log = Logger.getLogger(BlastXMLParser.class.getName());
```
josemduarte: See my previous comment on using slf4j

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLParser.java
```diff
@@ -0,0 +1,232 @@
...
+            String program = XMLHelper.selectSingleElement(blastDoc.getDocumentElement(),"BlastOutput_program").getTextContent();
+            String version = XMLHelper.selectSingleElement(blastDoc.getDocumentElement(),"BlastOutput_version").getTextContent();
+            String reference = XMLHelper.selectSingleElement(blastDoc.getDocumentElement(),"BlastOutput_reference").getTextContent();
+            String dbFile = XMLHelper.selectSingleElement(blastDoc.getDocumentElement(),"BlastOutput_db").getTextContent();
+
+            log.info("Query for hits in "+ targetFile);
+            ArrayList<Element> IterationsList = XMLHelper.selectElements(blastDoc.getDocumentElement(), "BlastOutput_iterations/Iteration[Iteration_hits]");
+            log.info(IterationsList.size() + " results");
+
+            resultsCollection = new ArrayList();
```
josemduarte: Missing type. There's still a few missing types for generics here and elsewhere in the pull request. Just follow any warnings that eclipse gives and try to solve as many as you can. If you can try to tidy those up that would be great. After that I think it would be ready for merging, nice work!

paolopavan: Dear Jose,
As I previously mentioned, type safe if guaranteed here since I previously have declared ArrayList<Result> resultsCollection. 
Sentences like resultsCollection.add(new String()) for example will already fail at compilation time. I  usually prefer it, because is less verbose and in my opinion cleaner. But if you really think it is important I can add it, just I can't do it until next week.

biojava-core/src/main/java/org/biojava/nbio/core/search/io/ResultFactory.java
paolopavan: Here you are right.

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastXMLParser.java
josemduarte: I think you are totally right, in this instance it is harmless. In any case I am of the opinion that warnings are important and should be reduced to a minimum. So if you can avoid them easily (as you can here), let's try to avoid them. If the code is free of warnings, then they become a lot more useful when they do appear because in many occasions they can really point out actual issues or bugs.

paolopavan: Ok

Inviato da iPhone

paolopavan:
Added discussed final improvements.

josemduarte:
Awesome, thanks so much @paolopavan ! The merge needs some manual attention, I'll try to get it merged when I have some time later today.

josemduarte:
So I've been looking into the pull request to see if I could merge it and all looks good, except for an issue with the refactoring of some of the alignment classes. 

I absolutely agree that the data structures to hold alignments should be in the core module (i.e. SimpleProfile, SequencePair etc). But together with those, some substitution matrix-related classes have gone into the core module. The reason for that is [a single call in SimpleProfile](https://github.com/paolopavan/biojava/blob/SearchIO_branch/biojava-core/src/main/java/org/biojava/nbio/core/alignment/SimpleProfile.java#L561). Which in turn is just used for some ad-hoc text display of the alignment in `SimpleProfile.toString`. I think we really could do without that dependency and have just the alignment data structures in a single package in core (`org.biojava.nbio.core.alignment`) without any of the substitution matrices stuff (which should stay in the alignment module). 

Could you have a look at it @andreasprlic? I think it should be easy to remove the dependency in `SimpleProfile.toString` but I don't know where it has been used and what for.

andreasprlic:
Can take a look but I am out of town this week, so has to wait until next.

paolopavan:
Hi Jose,
You're welcome. I could add some more features that I have in mind later in
these days, when I have some spare time.
Also note in my repository Biojava-Run, a project coupled with this. It is
already working but I'm going to refine it a bit with more generalization
and unit tests.

About the dependency from substitution matrices, I have already tackled
this since I was not very happy with that moving but the situation seems to
be hard to solve. In fact the matrix property is used both by toString()
method that is not so important (even if having a representation of the
alignment in toString() is a nice facility) but also in *getNumSimilars()
in SimpleSequencePair* class, extended from *SimpleProfile*, that is
involved also in alignment process (*FractionalSimilarityScorer* class).
One option may be to implement just the Blosum65 matrix in core and leave
the rest in Alignment, but it would lead to a code duplication.

Paolo


2015-11-03 2:02 GMT+01:00 Andreas Prlic <notifications@github.com>:

heuermh:
This may have been merged too early -- the changes are not binary compatible and should justify a bump to the next major version number, i.e. 5.0.

josemduarte:
This has been discussed at length here and in the [mailing list](http://mailman.open-bio.org/pipermail/biojava-dev/2015-August/007979.html). The incompatible change is the move of alignment data structures and substitution matrices from **alignment** to **core**. That should help going forward in other issues (e.g. #346). 

I'd say next version could still be a minor (i.e. 4.2) but if that's against semantic versioning I don't see a problem in going for 5.0. Anyway since the release of 4.1 we have a few new big things around: better support for multiple structure alignments, full implementation of secondary structure assignment, search IO. If we target a release date by the end of the year, I'm sure there will be a few more things to include.

sbliven:
+1 to version 5.0.

@josemduarte For instance, this broke the EPPIC build where we use substitution matrices.

Issue #330, opened 2015-09-21
DBREF records in PDB files written from StructureImpl cause IndexOutOfRange exception when reading back in


sroughley:
DBREF records do not all contain 80 characters when written out using StructureImpl#toPdb(), and so re-reading a PDB file written out this way causes an exception to be thrown as the DBREF parser in PDBFileParser at the following line:

String dbinsEnd = line.substring(67,68);  (line 2198)

There are a number of other StructureImpl Member classes which also do not write full length lines (e.g. HEADER is truncated after ~67 chars).  In most cases, these do not affect the biojava parser but may effect other tools which expect the full 80 char lines.  However, e.g. TITLE wraps the title over multiple lines each of 67 chars instead of using the full width.

I suggest:
1) Add an option to FileParsingParameters for allowShortLines, which simply adds ' ' chars to each line shorter than 80 characters during the reading process when it is 'true'
2) (Optional) Ensure all parsing methods which parse lines with optional parameters at the end are "short line safe"
3) Fix the toPDB() methods of all classes to ensure they write full 80-char lines

If agreed, I can start to work through these (starting with the first option)

josemduarte:
I totally agree with your comments, the parser should be able to parse lines with or without padded spaces. From your suggestion I'd definitely support doing 2 and 3, but regarding 1 I would say it's not a good idea to introduce another parsing parameter if not strictly needed. The parser can be made robust enough to parse both style of files, I would not see a use-case for having it as a parameter.

sroughley:
OK - in that case, for most records, the simplest way to achieve 2 is probably to simply right-pad out to 80 chars.  I will take a look at this in the next day or two, and then will start checking through some of the other methods to get 3 implemented.  I guess the one possible advantage of 1) is that it enables some sort of 'strict' parsing mode - but as you say I'm not really sure why anyone would do that!

sroughley:
I have submitted a pull request to deal with the simple case of padding out the lines to 80 chars during parsing, and have started work on the individual #toPDB() methods.

I was also looking, while I was dealing with the DBRef class to add support for DBREF1/2.  That doesn't appear too difficult, but I wondered why StructureException is not part of the method signature for the PDBFileParser#parsePDBFile methods, as this is described as "An exception during the parsing of a PDB file."?  I was about to change the method signature and realised that a lot of classes and testcases break, so decided to take a step back...

josemduarte:
[...]

Issue #305, opened 2015-07-24
More robust PDB/mmCIF parsing for non-deposited files


josemduarte:
BioJava is very good at parsing deposited PDB/mmCIF files but some work is required to make it more robust to unconventional formatting when parsing non-deposited PDB/mmCIF files.

The aim should be that the parsers are robust enough to produce informative warnings for bad formatting, if possible without crashing altogether.

The 2 formats (especially PDB format) are abused a lot and many softwares depart from the standard, so  surely all cases are impossible to cover. 

A good idea would be to add a test and a set of files from the most popular softwares that produce coordinate files (refmac, phenix, cyana, rosetta, modeller... ). Note that there is already a test for some of these issues in `org.biojava.nbio.structure.io.TestNonDepositedFiles`.

andreasprlic:
While I agree that we should not be too strict about the file format definitions, I don't think we should try to  go out of our way too far, either. The PDB file format has been abused quite heavily in the past and the future is with the more extensible (and XML-like) mmCif.

andreasprlic:
What is the status of this one.   I see there were several commits. Can we close it?

josemduarte:
I think it needs some more testing with a wider set of files. We are going to try doing more testing on this in the next 2-3 weeks, I'll close it after.

larsonmattr:
Before this ticket is closed, I propose to add a commit for better supporting non-deposited files.  The pdb parser is already robust for handling chains containing only non-polymeric hetatm and water Groups.  In this case, these Groups will still be added to the produced Structure instance.  

The current mmCIF parser will discard these non-polymeric groups after parsing to prevent them from being included in the Structure.  

In this branch (https://github.com/edlunde-dnastar/biojava/tree/mmcif-nonpolymer-chains) an added FileParsingParameter sets these Groups to be included in the parsed Structure as they are for the PDB parser.  This would be beneficial to make the mmCIF parser behave more closely to the PDB parser and better handle non-deposited structures that often will contain newly added ligands on unique chains.  Discarding such ligands in this case would be presenting incorrect data.

andreasprlic:
@larsonmattr This is a rare thing in official PDB files. e.g. one such case is 3o6j, which has a single water molecule in chain Z. If you want to get such cases added, your branch makes sense. 

[...]

PR #332, opened 2015-09-22
Modified PDBParser to fix bug #330

sroughley:
The PDBFileParser has been modified to ensure that all lines are right-padded to 80 characters prior to parsing to prevent certain parsers failing if the line is shorter than the position of some of the later optional fields, e.g. in DBRef

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/DBRef.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/DBRef.java
@@ -131,13 +131,13 @@ public class DBRef implements PDBRecord, Serializable{
 //        DBREF  3EH2 A    2   767  UNP    P53992   SC24C_HUMAN    329   1094
 //        DBREF 3EH2 A    2   767     UNP   P53992  SC24C_HUMAN   329   1094
 //        DBREF  3ETA A  990  1295  UNP    P06213   INSR_HUMAN    1017   1322
-        formatter.format("DBREF  %4s %1s %4d%1s %4d%1s %-6s %-8s %-12s%6d%1c%6d%1c",
+        formatter.format("DBREF  %4s %1s %4d%1s %4d%1s %-6s %-8s %-12s%6d%1c%6d%1c            ",
                 idCode, chainId,seqbegin,insertBegin,seqEnd,insertEnd,
                 database,dbAccession,dbIdCode,
                 dbSeqBegin,idbnsBegin,dbSeqEnd,idbnsEnd
                 );
 
-        buf.append(formatter.toString().trim());
+        buf.append(formatter.toString());
         formatter.close();
 
     }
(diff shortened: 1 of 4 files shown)


josemduarte:
This looks great! My only comment is that with the new padding there is a lot of redundant checks for length of line in the rest of the parser. It should be possible to make it slightly more efficient by removing those extra checks. Anyway that would be a marginal gain.

sroughley:
OK, I will try to track as many of those down as possible and remove them

sbliven:
Wouldn't it make more sense to add additional checks for the length, rather than padding with spaces and then trimming each field down to nothing?

andreasprlic:
The discussion here has not been active in some time. As such I am giving this a ping. I am actually with @sbliven . Rather than inflating the length of the Strings, I'd prefer more checks for the correct lengths. where needed.

sroughley:
I think there are potentially two issues here, which are closely related.  The PDB format requires 80 character lines, even if in many cases not all 80 characters are used.  In the case of the DBRef#toPDB() call, the 'writer' had a #trim() call at the end of the method which removed trailing spaces, with the effect that a broken DBREF line was written when some of the last fields in the line were blank.  Other methods (I think TITLE was one) also write shorter lines than they should.

It was for these reason that I chose to right-pad to the "correct" length as a pragmatic solution.  Obviously, we could assume that either a short line is incorrect and throw a parsing exception, and check and fix all #toPDB() methods through Biojava - probably the most "correct" method, or for each parser method handle short lines accordingly (e.g. either decide that a field which is beyond the end of the supplied line is actually populated by blank spaces, and treat it accordingly, or decide that the line is non-parsable, and again throw a parsing exception).

Any thoughts?

josemduarte:
Whatever the solution to this, I don't think we should be very strict with padding spaces at parsing time. I.e. if a line is correct (it contains all necessary information) but is not padded with spaces until column 80 we should still parse it. 

Many files produced by other softwares (other than official PDB files) don't have all the padding spaces and we really want to be able to parse them as long as everything else is correct (see issue #305 in relation with parsing non-deposited PDB files).

sroughley:
How about having an option on the parser "Allow short lines", which defaults to 'true'.

In this case (i.e. the option is 'true'), wherever possible, short lines are parsed.  I think the simplest way of achieving this is to right pad with spaces.  Some lines will then fail to parse if there are fields with spaces which do not translate to some sort of default or optional value.

If the user decides to set the flag to 'false', i.e. this requires a 'strict' PDB format of 80 characters, then any line which isnt long enough (do we check for length < 80 chars, or just that all fields are within the length?) then a parsing exception is thrown

Finally, I think we should then make sure that all toPDB() methods do return 80 chars

josemduarte:
Rather than introducing another parsing mode (which adds maintenance overhead), I'd rather do something like produce a warning when lines are not padded correctly. In my opinion the padding issue is really not important enough to deserve an exception, especially noting that the PDB file format should be considered already a legacy format and thus it's only of secondary importance.

+1 to all `toPDB()` methods to return 80 chars

sbliven:
So to close this issue, would it be ok if I merge 07de4bb (the DBRef stuff) while omitting the changes to PDBFileParser? If there are still cases where the parser fails then we can fix them individually with bounds-checks.

andreasprlic:
That would be a good way to take this forward.

sroughley:
OK - let's go with that and see where and if problems remain

PR #334, opened 2015-09-23
adding Fastq.convertTo(FastqVariant) and supporting methods

heuermh:
Adds support for explicit conversions, e.g.

```java
// illumina --> sanger variant conversion
FastqReader reader = new IlluminaFastqReader();
FastqWriter writer = new SangerFastqWriter();
FileWriter fileWriter = new FileWriter(new File("sanger.fastq"));

for (Fastq fastq : reader.read(new File("illumina.fastq")))) {
  writer.append(fileWriter, fastq.convertTo(FastqVariant.FASTQ_SANGER));
}
```

I'm not sure if the next release version will be 4.1.1 or 4.2 or 4.2.0, the ```@since``` tags will need to be updated accordingly.

### DIFF
--- a/biojava-sequencing/src/main/java/org/biojava/nbio/sequencing/io/fastq/AbstractFastqWriter.java
+++ b/biojava-sequencing/src/main/java/org/biojava/nbio/sequencing/io/fastq/AbstractFastqWriter.java
@@ -20,7 +20,14 @@
  */
 package org.biojava.nbio.sequencing.io.fastq;
 
-import java.io.*;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+
 import java.util.Arrays;
 
 /**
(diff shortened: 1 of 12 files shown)


dkatzel-home:
Thanks,
  This would work but it forces the user to remember to convert it themselves.

The writer instance should know what format it's in.  A better API would be to implicitly do the conversion for you.  So using your example:

    for (Fastq fastq : reader.read(new File("illumina.fastq")))) {
      writer.append(fileWriter, fastq);
    }

And then inside the `AbstractFastqWriter`
 
     @Override
    public final <T extends Appendable> T append(final T appendable, final Iterable<Fastq> fastq) throws IOException
    {
       //...null checks
        for (Fastq f : fastq)
        {
            Fastq converted = f.convertTo(getVariant());
            if (f != null)
            {
                appendable.append("@");
                appendable.append(converted.getDescription());
                appendable.append("\n");
                appendable.append(converted.getSequence());
                appendable.append("\n");
                appendable.append("+\n");
                appendable.append(converted.getQuality());
                appendable.append("\n");
            }
        }
        return appendable;


where `getVariant()` would be a new abstract method that returns a `FastqVariant` that the concrete classes would implement to return which ever variant they are.


Just my 2 cents.

heuermh:
I agree, an implicit conversion was how I thought this API worked in the first place.

Note this pull request makes four binary incompatible changes, as reported by the clirr maven plugin
```bash
$ mvn clirr:check
...
[ERROR] 7013: org.biojava.nbio.sequencing.io.fastq.FastqVariant:
  Abstract method 'public int qualityScore(double)' has been added
[ERROR] 7002: org.biojava.nbio.sequencing.io.fastq.IlluminaFastqWriter:
  Method 'protected void validate(org.biojava.nbio.sequencing.io.fastq.Fastq)' has been removed
[ERROR] 7002: org.biojava.nbio.sequencing.io.fastq.SangerFastqWriter:
  Method 'protected void validate(org.biojava.nbio.sequencing.io.fastq.Fastq)' has been removed
[ERROR] 7002: org.biojava.nbio.sequencing.io.fastq.SolexaFastqWriter
  Method 'protected void validate(org.biojava.nbio.sequencing.io.fastq.Fastq)' has been removed
```
Such wasn't strictly against release policy in the past, not sure what our policy is currently.

andreasprlic:
I don't see any problems with those tiny changes to the method signatures.

heuermh:
Thanks!

Issue #346, opened 2015-11-10
biojava-phylo module extension


lafita:
I would like to make some changes to the **phylo** module of biojava. I would like to factor out the pairwise distance calculation in [TreeConstructor.java](https://github.com/biojava/biojava/blob/master/biojava-phylo/src/main/java/org/biojava/nbio/phylo/TreeConstructor.java) to include easily more scoring functions for the evolutionary distance. I have in mind to include a structural similarity score, if the alignment comes from a structural alignment (some literature claims that the resulting tree is more accurate and has less uncertainty). Currently percentage of identity (PID) and substitution matrix scores are implemented.

I have also noticed that some code was ported from **Jalview** and is either not used or redundant with other biojava code. An example is the [ResidueProperties.java](https://github.com/biojava/biojava/blob/02952752ada4ddc0e03ce8a4eea4d78d74192aef/biojava-phylo/src/main/java/org/biojava/nbio/phylo/ResidueProperties.java) class, that contains hard coded substitution matrices (PAM250, BLOSUM62) and a lot of color assignments and options never used in biojava (I suppose they are used for the Jalview GUI). About the matrices, the **biojava-alignment** module contains already these two substitution matrices in the resources folder, and many more, but the phylo module does not depend on the alignment module, so they cannot be accessed. I think it would be good to add the alignment module as a dependency of the phylo module, since generating sequence alignments is a previous step to the phylogenetic tree construction.

Lastly, some names are confusing in the module. For example, the [TreeConstructionAlgorithm.java](https://github.com/biojava/biojava/blob/02952752ada4ddc0e03ce8a4eea4d78d74192aef/biojava-phylo/src/main/java/org/biojava/nbio/phylo/TreeConstructionAlgorithm.java) enum has BLOSUM and PID as its fields. By tree constructor, one understands one of the algorithms to build trees, like UPGMA or Neighbor Joining (NJ). Those are named in the enum [TreeType.java](https://github.com/biojava/biojava/blob/02952752ada4ddc0e03ce8a4eea4d78d74192aef/biojava-phylo/src/main/java/org/biojava/nbio/phylo/TreeType.java), which should stand for one of the three tree types (Distance Tree, Maximum-Likelihood Tree or Pasimony Tree). The BLOSUM and PID fields should be grouped into something named like **ScoreMatrixType** to be consistent with the literature.

Does anyone have comments on (or oppose to) these changes? 
Is there any important point I am missing?

Aleix

josemduarte:
[...]

Issue #350, opened 2015-11-17
Update the forester version (from 1.005 to 1.038)


lafita:
This is important, since the version we are currently using contains a **BUG** that allows for negative branch lengths in trees constructed with the Neighbor Joining algorithm and another bug that incorrectly calculates the branch lengths of the internal nodes. I discovered it with a simple Test that expects a tree of error zero.

This affects all the **phylo** classes (since the NJ is the only tree constructor algorithm) and the *GuideTree* class in the alignment package, used to construct progressive multiple sequence alignments (with the consequence that not necessarily the closest sequences are paired).

sbliven:
We control the forester package on maven central, so we need to wrap the latest release in a pom and upload it. @andreasprlic did this process for 1.005, or I also have permissions to do it.

lafita:
Now that we are discussing this, the **forester** library is mainly a tree visualization library, with very few tree inference and construction support (not very used for the inference features, thus prone to bugs like the one above).

There exist other Java libraries focused on the algorithmic part of phylogenetic tree construction. The most popular one is [Phylogenetic Analysis Library (PAL)](http://iubio.bio.indiana.edu/soft/molbio/java/pal/doc/) a general purpose library for molecular evolution and phylogenetics. Many other libraries use the code, and it covers the whole field (parsimony, distance, maximum likelihood, UPGMA, NJ, etc). 

At the bottom of the page they propose other alternatives for phylogeny in Java (and they cite BioJava as well).

josemduarte:
The PAL project looks good but very out of date, last release was 1.4 in 2002 (http://iubio.bio.indiana.edu/soft/molbio/java/pal/). Aren't there any other more up-to-date and maintained alternatives from the list they provide?

lafita:
I agree that the project is very old, but it looks complete and well organized. It also seems to be very popular, since other projects depend on it for the fundamental phylogenetic algorithms. 

The alternatives they provide (and that I could found to be maintained and up-to date) are either in another language or very specific to a single application. There seems to be a lack of a maintained and up-to date **general purpose phylogenetic library** in Java, which I think is a great opportunity for **BioJava**. 

An idea could be to import the code we consider relevant from the **PAL** library (and possibly other more specialized libraries) and provide support for it (testing and documentation), together with adaptation to the **BioJava** Data Structures. It seems a big project that should be well planned and with the opinion of potential users. Very interesting also!

josemduarte:
[...]

PR #351, opened 2015-11-18
Extend and Merge the biojava-phylo module

lafita:
The modifications are:
- Use biojava `SubstitutionMatrix` data structures and methods. Eliminate the Jalview code (named `ScoreMatrix`) that was intended for this (redundant).
- Make names consistent: `TreeType` contains the optimization criteria (score) and `TreeConstructorType` contains the clustering algorithm.
- Factor out the distance matrix calculation from the `TreeConstructor`. The user can use any distance matrix. The `TreeType` is also factored out, but only Neighbor Joining is available.
- Remove the `TreeConstructorListeners`, replaced by individual logging in each class (they were not listeners in the first place, because they only created logger messages when called).
- Use least squares error (LSE) instead of the L1 norm error in the `DistanceTreeEvaluator` (convention in the literature).

The new features are:
- Implement new functions to calculate evolutionary distances from an MSA (sequence) or MStA (structure). New functions are: Kimura and Poisson corrections, dissimilarity score (based on Substitution Matrix, it was already implemented in **alignment** module) and structural distance (based on RMSD and diffusion model of protein structures, Gishin 1995).
- Create Wrapper functions for forester-biojava conversion.
- Create a demo (obtained from the `CookBook`) and tests.

The tests will fail until the forester version is updated to the newest version #350, so they are skipped.
This closes issue #346, but it leaves open the issue whether **phylo** and **alignment** modules should be joined.


josemduarte:
Thanks, this is awesome! I'm going to look into fixing the forester issue. After that we really need to decide whether we join **phylo** and **alignment**.

andreasprlic:
we are waiting for #350 to be resolved before making this merge, right?

andreasprlic:
Actually, It is better to keep pull requests small, so we can be more atomic about moving forward. I think we should have discussed the merge between alignment and phylo modules some more. I'd prefer if the alignment module does not get dependent on forester. (see #346)

lafita:
The major part of the changes are renaming and reorganizing the existing code. I agree that I should have waited for commit 2f9c652. After my comment in #346 (alignment already depends on forester), do you @andreasprlic still think we need to take the commit out?

We need #350 for the tests, because they are now skipped. I tried with the new version and they all pass. So this could be merged and when #350 is resolved, we remove the *Skip* tags of the tests.

lafita:
This pull request has been refactored into extending #370 and merging phylo. The latter has still to be discussed.

Issue #319, opened 2015-08-14
Include Graph Dependency in structure Module


lafita:
As decided in #299, the current Graph classes and algorithms used in biojava will be replaced by an external dependency. The graph library proposed was jgrapht.

PR #299, opened 2015-07-14
Implement Graph Data Structures and Search Iterators

lafita:
Since there was already a Graph interface and an implementation for Undirected Graphs in biojava (symmetry project) and I was using a lot graph searching algorithms and directed graphs for symmetry, I created some new classes to represent different types of Graphs and the two basic searching algorithms as iterators for them. New classes:

- DirectedGraph
- DirectedAcyclicGraph (DAG)
- Depth First Searching (DFS) iterator
- Breath First Searching (BFS) iterator
- Test for their correctness

The idea is to have a set of basic and simple graph support, without importing a whole external library. The code is now in structure.symmetry.utils, would it make sense to create a new package for it? It is general enough so that it doesn't have to be inside the structure package.

Another possible extension for graph features would be to include Weighted Graphs, but I haven't done anything about that since I haven't need them until now.

PR #352, opened 2015-11-23
Include jgrapht dependency to biojava-structure

lafita:
As it was discussed in #299, the **BioJava** implementations and algorithms for **Graphs** have been replaced by an external library: [jgrapht](http://jgrapht.org). 

The only code replaced is related to the **Quaternary Symmetry** detection. Tests pass and demos work as before, but @pwrose might want to take a look and check that the changes are conceptually correct. 

As a summary, the **ComponentFinder** was checking if a **SubunitGraph** had more than one connected component. This is now done by using the **ConnectivityInspector** from jgrapht, that returns true if the graph is connected (one component) or not (more than one component). The constructor of the **SubunitGraph** is the same, and also the construction of a **SubGraph** given a set of vertices.

Merging this closes issue #319.


lafita:
It was actually a bad idea to replace all the **Lists** for **Sets**, when only a single conversion was needed to call the **jgrapht** connectivity algorithm. I have undone some of the changes so that the tests for local symmetry introduced by @josemduarte pass.

josemduarte:
Thanks!

Issue #342, opened 2015-10-28
mmCIF parsing support for missing SITES information.


larsonmattr:
BioJava should support any SITE information present in mmCIF files.  

The [Simple]MMCIFParser could read _struct_site and _struct_site_gen categories and trigger handlers in the [Simple]MMCIFConsumer to handle these loop entries to build Site classes in the BioJava Structure instance.

Supporting SITE information is present in parsing of PDB files but not yet in mmCIF.

larsonmattr:
Implementation was merged closing this issue.

josemduarte:
There seems to be a bug in matching the residue numbers and chains for the sites in struct_site_gen. Could you take a look? 

I've added a test here to demonstrate the issue: https://github.com/josemduarte/biojava/tree/sitesIssue

I came across this because I was seeing a lot of warnings like this while testing:
```
org.biojava.nbio.structure.io.mmcif.SimpleMMcifConsumer - comp_id doesn't match the residue at H221 - skipping
```

larsonmattr:
Jose,

 Thanks for finding this.  It is a problem with not using the insertion
code when searching for residues.  Also related to this, there is issue
with the Site.toPDB() method not formatting SITE entries correctly for DNA
residues like DT that are three-letter codes with whitespace.

 I'll submit a fix for both issues.



On Thu, Jan 14, 2016 at 7:55 PM, Jose Manuel Duarte <
notifications@github.com> wrote:

> There seems to be a bug in matching the residue numbers and chains for the
> sites in struct_site_gen. Could you take a look?
>
> I've added a test here to demonstrate the issue:
> https://github.com/josemduarte/biojava/tree/sitesIssue
>
> I came across this because I was seeing a lot of warnings like this while
> testing:
>
> org.biojava.nbio.structure.io.mmcif.SimpleMMcifConsumer - comp_id doesn't match the residue at H221 - skipping
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/biojava/biojava/issues/342#issuecomment-171844414>.
>



-- 
Matt Larson, PhD
Madison, WI  53705 U.S.A.

PR #357, opened 2016-01-04
Implementation of SITES parsing for mmCIF

larsonmattr:
Issue #342 describes missing functionality of parsing SITE equivalents from mmCIF files.

These are the _struct_site and _struct_site_gen records.

This pull request contains the implementation of parsing for these records to add the missing Site instances into the Structure being parsed.  The unit test checks that the correct # of sites and their descriptions are present in the parsed structure.  This submission would close issue #342.


josemduarte:
This looks great, thanks! I'll merge it

Issue #343, opened 2015-10-28
mmCIF parsing support for missing SSBonds information.


larsonmattr:
BioJava should support any disulfide bond information present in mmCIF files.

_struct_conn categories are already read by SimpleMMCIFConsumer and trigger handlers in a consumer to handle these loop entries.   The SimpleMMCIFConsumer could parse _struct_conn to build SSBond instances that are part of BioJava Structure instances.

Support SSBOND-like information parsed for PDB files but not yet in mmCIF.

larsonmattr:
Closing this issue.  When #220 is implemented (5.0.0?) then a better representation of "connections" (StructConn) and allowing bonds between different symmetry related molecules should be developed.

josemduarte:
There were a few issues remaining with ssbond parsing in both pdb and mmcif that should be fixed now with last commit

Issue #220, opened 2014-12-12
Better support for symmetry in the Structure model


sbliven:
PDB and mmCIF files use symmetry operators to reduce the number of atoms which need to be specified. This is used with NCS to reconstruct the asymmetric unit (e.g. in viruses), as well as for specifying biological assemblies (BA).

BioJava is able to read and understand the symmetry operations required to generate the full structure, but the data model is far from ideal for this. Most analysis applications require all atom positions to be stored in an array, so we would like to be able to store a representation of the full biological assembly. The problem is that the current model assumes that chainIDs are unique within a particular model. Thus, dealing with BAs requires one of these work-arounds:
1. **Use multiple Structure objects.** Cons: Can't use methods requiring a single `Structure` object. Some methods which take `Atom[]` assume that all atoms share a structure through the getParent() hierarchy, e.g. for cloning atom arrays. Structure metadata (e.g. header data) is lost, duplicated, or inconsistent.
2. **Rename chainIDs so that all are unique.** Workable, now that mmCIF officially supports 4 character chains. Cons: Difficult to map back to original chainID and symmetry operations. No write support (pending #188).
3. **Use multiple models within a single Structure.** Current approach for RCSB-supplied BA files and for structure alignment files. Cons: Only intended for NMR structures. Difficult to map back to original symmetry operation. Other tools (specifically pymol, but also jmol to a lesser extent) expect models to have identical contents and be superimposed.

A long-term solution would be to associate chains with a particular NCS and CS operator. These could be additional objects in the Structure hierarchy, or could just be fields in Chain. For instance:
1. Structure
2. Model
3. Unit Cell
4. Asymmetric Unit
5. Chain
6. Group
7. Atom

josemduarte:
I think the solution should come through introducing a more general identifier for molecules within the `Structure`. Currently this is the situation:
- the __chain id__  is an identifier within the asymmetric unit, i.e. an _asym id_
- the __model id__ is an identifier for multiple conformations

Those 2 have well-defined uses, so we shouldn't touch them, we need to introduce a new identifier, let's call it __biological assembly identifier (BA id)__ . The identifier would then be used in this hierarchy:

1. Structure (a biological assembly)
2. Model (a conformation)
3. Chain (identified by a biological assembly id)
4. Group
5. Atom

The hierarchy would roughly work like this:
[...]

PR #358, opened 2016-01-04
Implementation of SSBond parsing for mmCIF

larsonmattr:
Issue #343 describes missing functionality of parsing SSBOND equivalents from mmCIF files.

This implementation reads _struct_conn records for pairs of CYS residues that are bonded.  The parsed Structure instance will then have SSBond instances created to represent the disulfide bonds.  The unit test checks for the creation of the expected SSBond instances in a given structure.

This submission would close issue #343.


andreasprlic:
Thanks for this patch! 

Can you set the insertion code to null if it is '?' in the file? This is for consistency with other parts of the code base, where it is null by default.

Thanks,

larsonmattr:
Ok, I've made the change to set ? insertion codes as null rather than as an empty string.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/SimpleMMcifConsumer.java
```diff
@@ -1816,6 +1879,71 @@ public void newStructConn(StructConn structConn) {
 		this.structConn.add(structConn);
 	}
 
+	void createSSBonds() {
```
josemduarte: It is good practice to explicitly set the visibility of the method, it really helps others understanding the code and helps readability in general. We've tried to do it throughout BioJava as much as we could. I'd set it to private unless otherwise needed.

josemduarte:
One issue that is not treated here is ssbonds between symmetry partners. The mmCIF format does annotate them unlike the PDB format. They use the fields `ptnr1_symmetry` and `ptnr2_symmetry` that annotate which symmetry partner each side of the bond belongs to. `1_555` means asymmetric unit (1 is an identifier for the symmetry operator and 555 the cell translation using x=5, y=5, z=5 as origin).

We can't really annotate ssbonds between symmetry partners because the BioJava data structures don't really support symmetry partners properly (see #220) but at least we should make sure that we only parse those belonging to the asymmetry unit.

josemduarte:
Another issue is alt conformations when they occur in a CYS involved in the ssbond. See for instance 3dvf: in that case one of the partners in the ssbond has two alt conformations A and B. The mmCIF file annotates 2 ssbonds, one for each alt conformation. The current implementation of SSBond can't handle that either (it lacks an alt conf field).

andreasprlic:
I don't think there is much pain in parsing in all SSBonds. It might come in handy in the future, once #220 has been resolved.

josemduarte:
Actually parsing the symmetry-related bonds can be an issue: imagine you have a molecular viewer that simply takes all ssbonds that we parse (including symmetry related ones). If there is a ssbond between 2 chains related by a 2-fold symmetry axis, the molecular viewer would then draw a bond between 2 far-apart CYS in the same chain, since it has no knowledge that the ssbond actually refers to 2 symmetry-related chains.

I agree, once #220 is resolved we should then parse them all. But at the moment I don't see we can accommodate that info easily.

larsonmattr:
As a work-around for the issue presented by Jose, with an SSBond between
two symmetry-related chains could the parsing of the SSBonds be limited to
within the asymmetric unit?  I.e., build SSBonds but limit the scope to
avoid issues with symmetry partners?  In this case, we would need to modify
this implementation to check that ptnr1_symmetry and ptnr2_symmetry match,
for now.  Would it be acceptable to add that limited SSBond implementation
into BioJava while the issues of #220 are hammered out?

---

I've been reading through #220, and it isn't clear how much of the data
structures will be changed to support symmetry.  SSBonds are useful to
quickly query a common information about a structure.  However, like
Structures, Chains, etc these are derived from a PDB format model.
Long-term, will BioJava's Structure more resemble the mmCIF model and shed
its roots as a representation of a PDB?

Here, I didn't want to change the model of the SSBond class as the goal was
more parity with PDB parsing.  Longer term one might want to focus more on
a general representation of _struct_conn, from which SSBond-type
information is only a subset.  Again this diverts away from the existing
representation of a PDB format.

On Mon, Jan 11, 2016 at 1:55 PM, Jose Manuel Duarte <
notifications@github.com> wrote:

> Actually parsing the symmetry-related bonds can be an issue: imagine you
> have a molecular viewer that simply takes all ssbonds that we parse
> (including symmetry related ones). If there is a ssbond between 2 chains
> related by a 2-fold symmetry axis, the molecular viewer would then draw a
> bond between 2 far-apart CYS in the same chain, since it has no knowledge
> that the ssbond actually refers to 2 symmetry-related chains.
>
> I agree, once #220 <https://github.com/biojava/biojava/issues/220> is
> resolved we should then parse them all. But at the moment I don't see we
> can accommodate that info easily.
>
> —
> <https://github.com/biojava/biojava/pull/358#issuecomment-170670461>.
>



-- 
Matt Larson, PhD
Madison, WI  53705 U.S.A.

josemduarte:
Absolutely, that would be a great solution. As I said above one just needs to check that both partners have `1_555` in the `ptnr*_symmetry` fields. Ssbonds having other values can be discarded. That should solve the issue.

larsonmattr:
Ok, I will update the implementation so.  I will treat this as a general label to look for regardless of whether it is a structure with a defined crystallographic space group or derived by NMR method with no symmetry.  As a sanity-check I looked at an NMR structure and it used the convention of 1_555.

andreasprlic:
@larsonmattr about the question re. long term plans: The goal for the Structure object was always to provide a "scientifically correct" representation of structural data. The fact that the info comes from PDB files or mmCif is secondary. Ultimately the data model required to represent the information stored in both luckily is very similar.

larsonmattr:
Thanks Andreas and Jose for the helpful comments on this ticket and for
merging the changes in.

I would agree that the current Structure data model isn't that different
than the direction of the mmCIF format.  But the mmCIF format lends itself
towards database like representations rather than the hierarchical
structures of Structure->Model->Chain->Group->Atom.  Consistency of linking
ids is now more of an issue with reading/writing of mmCIF _loop sections.
On the plus side, the good organization of the BioJava mmCIF parser has
helped make it very easy to start hooking additional implementations.

On Mon, Jan 11, 2016 at 5:45 PM, Andreas Prlic <notifications@github.com>
wrote:

> @larsonmattr <https://github.com/larsonmattr> about the question re. long
> term plans: The goal for the Structure object was always to provide a
> "scientifically correct" representation of structural data. The fact that
> the info comes from PDB files or mmCif is secondary. Ultimately the data
> model required to represent the information stored in both luckily is very
> similar.
>
> —
> <https://github.com/biojava/biojava/pull/358#issuecomment-170734148>.
>



-- 
Matt Larson, PhD
Madison, WI  53705 U.S.A.

PR #363, opened 2016-01-11
Change AbstractFeature Qualifiers to a key:Array from key:value (#302)

emckee2006:
change qualifier to list in AbstractFeature. Change FeatureInterface. Change FeatureDbReferenceInfo because it also uses the FeatureInterface. Note: this was not intended and may cause unexpected issues


emckee2006:
Does this fix the issues from your merge request, Zach?

emckee2006:
Just a reminder that this still needs to be checked over :)

josemduarte:
So if I understand it correctly this is to fix issue #301 and addresses all problems raised in pull request #302. Is that right? (I haven't checked the changes yet, just asking and referencing the appropriate issues). This pull request would then override #302

emckee2006:
I hope so. If not it's hopefully a step in the right direction and can be
amended further, if needed...

paolopavan:
Just a comment: I see that you added location writing logic in a
specific writer
class. Why don't you include this in the apposite class InsdcParser and if
still needed wrap it into GenericInsdcHeaderFormat ? It may improve ease of
maintenance.

Il mercoledì 13 gennaio 2016, emckee2006 <notifications@github.com> ha
scritto:

emckee2006:
That is a good question.  I don't think i added new location writing
logic.  All I did was take the previous PR and improve the generic types,
as we had discussed.

paolopavan:
I see. I improved the InsdcParser class (actually I almost completely
rewrote it) but I was not the original author and it was already there when
I added support for nested location and split. So your part maybe was
already there even if it seems clear to me that the two procedures should
be unified. Ok, if you don't do it before maybe I will in the future.
Thanks!

Il giovedì 14 gennaio 2016, emckee2006 <notifications@github.com> ha
scritto:

andreasprlic:
I think this is at least a step in the right direction, so merging this in.

PR #370, opened 2016-01-14
Extend the biojava-phylo module without merging

lafita:
Split pull request #351 into phylo extension and phylo-alignment merge.


lafita:
Closes issue #346

PR #371, opened 2016-01-14
Some fixes to make BioJava classes serializable - so they work with Apache Spark

abradle:


heuermh:
Thank you, @abradle.  I'm curious, what you would like to do with Biojava on Apache Spark?

PR #373, opened 2016-01-15
Fixes for two SITES issues

larsonmattr:
These changes are for:

1) fixing an issue found with PDB:1A4W, that has a Site containing residues with insertion codes, that weren't properly handled by submission for issue (#342).  Now, if there is an insertion code, this is used in the residue lookup.  When there was no insertion code (?) null is used to match conventions.

2) fixing an issue with the Sites.toPDB() method.  When residues with whilespaces are in a site, the column formatting for a SITE line is broken.  This was because a String.trim() method was used on the line with the intent of removing trailing whitespace, but trim() removes proceeding whitespace as well.  I've also made a change to format the output PDBName as a three-letter field, in the case that the Group.PDBName() was set as a shorter string such as "DT".


josemduarte:
Thanks!

Issue #232, opened 2015-01-14
Auto-detecting file formats


sbliven:
It would be convenient to auto-detect file formats. This is a hard problem, and one where the other Bio* projects differ. BioPerl has pretty good support for guessing file formats based on extension, magic bytes, or contents. BioPython has adhered to the python philosophy of 'explicit is better than implicit', and required the user to provide the format type.

What's the BioJava philosophy? I ask because I would like to implement auto-detecting structure formats (e.g. PDB vs mmCIF vs XML).

sbliven:
There is a bit of magic happening already in `InputStreamProvider` to distinguish gzipped files.

josemduarte:
Auto-detecting PDB vs mmCIF would be a really nice feature indeed. We implemented a very rough one in https://github.com/eppic-team/owl, which would decide by checking the first line of the file against these signatures (regexes):

```java
String[] FILE_SIGNATURES  = {
		"(?:HEADER|TITLE|COMPND|SOURCE|EXPDTA|REMARK|DBREF|SEQRES|CRYST1|MODEL|HELIX|SHEET).*",
		"ATOM\\s+.*",
		"PFRMAT\\s+TS.*",
		"PFRMAT\\s+RR.*",
		// mmCIF files from PDB always start with "data_PDBcode". 
		// Phenix's cif files do start with data_ but then with no PDB code 
		"data_\\w+.*"
	};
```

pwrose:
+1

andreasprlic:
I added some checks for magic numbers at the [InputStreamProvider](https://github.com/biojava/biojava/blob/master/biojava3-core/src/main/java/org/biojava3/core/util/InputStreamProvider.java) to check what kind of compression is getting used.

The main use-case that I see is in user interfaces, where a user can upload a custom file format. This could allow to support both PDB and mmCif file formats. ( I think currently all UIs allow only PDB format).

phidias51:
It might be a good idea to separate the file sniffing code from the input
stream code.  This would let you use the file sniffer in different
contexts.  For example, you might use them as a key to select a file parser
(not just a compressor).

That said, it might be worth looking at Apache Commons Compress
CompressorStreamFactory which handles more file types.

Cheers,

Mark

agupta1989:
+1

email sent using phone app.

andreasprlic:
@sbliven can we close this one with you patches?

sbliven:
Yes. It should autodetect unless your file has the wrong suffix.

PR #376, opened 2016-01-20
StructureIdentifier framework

sbliven:
Add substantially new unified class for identify structures: StructureIdentifier

This unifies various prior methods for identifying a structure from a string, which was
previously spread over AtomCache methods and StructureTools/StructureIO utilities. It
includes PDB IDs, residue range specifications, URLs, SCOP/CATH/PDP domains, and files.

Major Classes:
* StructureIdentifier is an interface for methods which know how to load a structure from
  some resource and reduce it to the requested substructure. The most important property
  is the identifier, which can be an arbitrary string.
* SubstructureIdentifier is considered the canonical way to specify a structure and
  is the major implementation for chain and residue level substructures. All identifiers
  can be converted to a SubstructureIdentifier.
* StructureName is suitable for wrapping user-supplied identifiers, and it dispatches the
  request to a more specific class based on a guess as to the type (PDB, URL, file, etc).

StructureIdentifiers can represent arbitrary strings (e.g. domain IDs). These are converted
(possibly through some relatively slow process like downloading and parsing a file) into
standard-format SubstructureIdentifier instances, which should be easier to serialize and
recreate.

Detailed changes:

* Substantially changes StructureIdentifier, which existed but wasn't used anywhere.
  * Remove getPdbId() method, since not globally relevant. It's still present in most
    implementations, and can always be accessed via toCanonical().getPdbId()
* AtomCache
  * Accept chain indices ("4HHB.0") with a warning if there was not
    already a chain with that ID.
  * Removes support for the strictScop parameter. The new default is effectively strictScop=false. This can be overridden by getting a ScopDomain directly from a particular ScopDatabase, rather than using generic methods like AtomCache.getStructure(String), which use the StructureName class for fuzzy matching of domain names.
* Better guessing of PDB/CIF filetypes from filename (but not full autodetection from contents, as requested in #232)


sbliven:
I'd welcome a code review of this pull. Particularly, what should the return type for getCanonical() be? Ideally, it would be an identifier capable of representing arbitrary selections from a structure. Currently, the best we have is an identifier capable of representing residue ranges (SubstructureIdentifier).

josemduarte:
Travis is failing, is that some networking issue?

I'll give a look at the code when I get a bit of time. Thanks for the pull request, this looks awesome!

A question, this would support multi-letter chain identifiers, wouldn't it?

sbliven:
Tests pass on my laptop, but there are plenty of changes here. I'll restart Travis and double check.

lafita:
There is a Test in the core module that sometimes gives a network exception from Travis. Is it the case?

lafita:
The error is in the test files I created for MultipleAlignment serialization (XML, FASTA, etc), because the names of the structures do not match (that is expected from your changes @sbliven). We just need to update the files with the new names.

sbliven:
Ok, finally fixed the minor test regressions and got travis to finish the tests.

sbliven:
I'm going to merge this before master diverges too much. Comments about the new code are still appreciated.

Issue #341, opened 2015-10-28
mmCIF parsing support for missing CONECT ChemComp information.


larsonmattr:
The SimpleMMcifConsumer does not create use Structure.setConnections to create a CONECT record list during MMCIF parsing.  It does have addBonds(), but this will create bonds for all amino acids, nucleotides, etc.  

There is already existing functionality to pull ChemComp records for a ligand that can be used to build the connections information to replicate the CONECT records of PDB files.  

BioJava should add to the SimpleMMcifConsumer the ability to make CONECT bonds only for the ligands, build from ChemComp data, without needing to create bonds for standard residues like amino acids and nucleotides.

josemduarte:
So what about the inter-residue connections annotated in `_struct_conn` within the mmCIF file itself? I think we should also parse those and add them to the relevant bond data structures, shouldn't we?

larsonmattr:
For the issue 343 (ssbonds), the implementation that has been started parses _struct_conn limited to looking for disulfide bonds.  This task could more generally parse bond information of _struct_conn for the bond information stored within the mmCIF.  

Generally, it appears that mmCIF files in the protein databank are absent the ligand bonds that PDB files contained leading to the approach using ChemComp groups to grab this information.  

Parsing _struct_conn should be also implemented to get any additional bond information contained in the mmCIF.

andreasprlic:
about the ligand bonds: these are available through the chemical component dictionary. It can be accessed in BioJava by using the AllChemCompProvider, or the DownloadChemCompProvider (depending on your use case).

andreasprlic:
Thanks for the patch!

josemduarte:
I am revisiting this again and have just realised that we did have bonds parsing before this issue. The information was stored into the `Atom` objects (through `getBonds`) but wasn't available in the `Structure` object via `getConnections()`. The class `BondMaker` was doing all the work, so the new `LigandConnectMaker` introduced in #379 is redundant.

I would propose to simply deprecate `Structure.getConnections()` because it's a legacy representation too tied with the CONECT representation of PDB files. `Atom.getBonds()` does a much better job. I would then also remove `LigandConnectMaker` so that we are sure we don't do the bonds parsing twice.

If there's nothing against this, I'll submit a pull request to do as described above.

josemduarte:
[...]

PR #379, opened 2016-01-22
Adding CONECT information when parsing mmCIF files.

larsonmattr:
These changes are for addressing and closing issue #341 (creating CONECT information when parsing mmCIF).  

FileParsingParameters is updated to include an option setCreateLigandConects(true/false) that enables creating CONECT-style records for ligand groups in a Structure.

SimpleMMcifConsumer, when isCreateLigandConects() will build CONECT-style records and use a new LigandConnectMaker class to find and create CONECT records for all ligands in a Structure.  This relies on ChemComp groups to supply the missing bond information (mmCIF, unlike PDB, do not include this in the file normally).  Along with the mechanism of DownloadChemCompProvider or other ChemComp provider this provides an automatic way to annotate the ligand bonds.

The PDBFileParser likewise recognizes the isCreateLigandConects(), and fill in missing CONECT bonds for ligand groups based on the ChemComp information.  This can be useful for minimal ATOM-only PDB files that need to add bond information.

Also, there are two tests: 
1 unit test that mmCIF parses Bonds for common ligands like in PDB:4HHB
1 unit test that mmCIF parses Bonds for a ligand in PDB:3UCB that has alt. conformations.


larsonmattr:
All tests passed locally, and it is unclear why Travis has an error in biojava-integrationtest with BerkeleyScopInstallationTest>ScopDatabaseTest.testCategories

andreasprlic:
I see a org.biojava.nbio.structure.scop.ScopIOException: java.io.IOException: Can't parse line d3spbb_	
Caused by: java.io.IOException: Can't parse line d3spbb_

Are your locally cached files the latest version?

larsonmattr:
It may have tripped travis because it failed to download dir.com.scop.txt_2.05 from the SCOP_DOWNLOAD site.  

I've merged in the recent changes from the last pull, and checks pass.

andreasprlic:
yes, was prob a temporary networking issue at Travis..

andreasprlic:
I accepted the merge. One thing to think about still is if this behaviour should be turned on by default.

PR #381, opened 2016-01-25
Added functionality to write groups to MDL molblocks (as strings)

abradle:
Added function to Group interface
Added function to HetatomImpl implementation of that interface


andreasprlic:
thanks, nice utility method...

PR #382, opened 2016-01-25
Getting Junit test to work on installations that have empty spaces in…

andreasprlic:
… directory names.

Currently these junit tests don't work if there is a space character in one of the directory names of an installation. (not that I endorse having spaces in directory names! ;-)


andreasprlic:
This addresses some of the problems of #383

lafita:
I would recommend to undo the commit ce107ff, because now that we know what was causing the error it is better to throw an Exception if it happens again that a Group has null chain

PR #386, opened 2016-01-26
Updated the SecStrucCalc function and the Test

abradle:
1) Function is a lot faster for big structures. e.g. 3j3y goes from 30min+ to 40seconds.
This is by using grid and contactsets to find groups for which the calculation needs to happen
2) Class does not need to be recreated for each run through - Class global variables are reset when .calculate is called
The tests pass fine


josemduarte:
@lafita could you check this works well with some of the corner-case structures you've been working with?

abradle:
I've also added a fix for a resource leak in the DownloadChemCompProvider

andreasprlic:
awesome speedup!

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/DownloadChemCompProvider.java
```diff
@@ -264,9 +264,20 @@ public  ChemComp getChemComp(String recordName) {
...
 
 		}
+		finally{
+			// Now close it
+			if(inStream!=null){
+				try {
+					inStream.close();
+				} catch (IOException e) {
+					// This would be weird...
+					logger.error("COULD NOT CLOSE CHEMICAL COMPONENT FILE ->>>>> RESOURCE LEAK RESOURCE LEAK. MAYDAY!!!!!!!!! MAYDAY!!!!");
```
lafita: Why don't we throw the `IOException` if this ever happens? Or throw another type of Exception with the same message, if it is so weird that it happens.

lafita:
I have reviewed the code and everything looks fine! Tests pass so it is also a good sign. 
Thanks @abradle, this **speedup** will help a lot in any other algorithm using secondary structure (e.g, `CeSymm`).

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/DownloadChemCompProvider.java
josemduarte: We can't throw the IOException in there since the method signature (inherited from an interface) does not declare a "throws". Changing that would require some API changes. I agree that would be a better design, where we then would deal with the IOException further down the line.

Note that the actual resource leak happened because `inStream` wasn't being closed at all, whether an IOException happened or not. By putting the `close()` call in the `finally` we make sure it always closes (only in the unlikely case that the closing causes an exception then there could be resources left opened).

lafita: Ok, it makes sense! I did not notice the inheritance, sorry. I will look for non closed files similar to this one in `AtomCache` regarding #395

Issue #387, opened 2016-01-26
Demos are not compiled and can't be run with default set-up


josemduarte:
At the moment demos have been excluded from compiling in the pom.xml files. That means that with default set-up, they aren't compiled and can't be run. I believe the idea to exclude them was to avoid namespace collisions (demo package is the same in all modules). The fix should be excluding them from jar but not from compiling.

PR #388, opened 2016-01-26
Making demos compile but not package into jar #387

josemduarte:
This should solve the issue with demos not being compiled or run #387. They are excluded from the jar but not from compiling.


josemduarte:
I'll merge this in. We need to test at some point that this doesn't affect the signing of jars. I've already checked that jars created with this don't contain the demo packages. So it shouldn't affect anything in principle.

Issue #353, opened 2015-11-23
mmCIF parsing support for missing SEQRES information


darnells:
The current PDBParser supports parsing SEQRES records from a PDB file header; however, the MMCIFParser does not yet support parsing sequence information from a mmCIF file.

The current implementation in SimpleMMCIFConsumer is not using the mmCIF-equivalent SEQRES records; it uses ATOM group equivalents to construct the current component list.

The SimpleMMCIFParser should parse the analogous records from an mmCIF file (_pdbx_poly_seq_scheme)--the direct equivalent to the SEQRES record--and construct a component list representing the SEQRES sequence as does the PDBParser (this is required for passing existing integration tests: `HeaderOnlyTest>MMcifTest.testLoad:63->MMcifTest.comparePDB2cif`).

darnells:
We have implemented this feature in my company and are testing it internally. We will be submitting a PR.

josemduarte:
There is some support for SEQRES records for mmCIF files at the moment, but it will only work when setting `FileParsingParameters.setAlignSeqRes()` to true. When that flag is set, the SEQRES groups in the `Chain`s are populated. What other information from pdbx_poly_seq_scheme are you looking for? More that what's stored in `Chain.getSeqResGroups()` ?

Related to that and regarding testing, there is a very useful test in the **biojava-integrationtest** module that does a thorough comparison of the parsed data from PDB and mmCIF files to check the consistency of the parsers: [TestLongPdbVsMmCifParsing](https://github.com/biojava/biojava/blob/master/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/io/TestLongPdbVsMmCifParsing.java). It runs the parsers through 1000 randomly-selected PDB entries. The test is not enabled by default because it takes many minutes to run. But I've found it very useful whenever I modify anything related to the parsers. Run it manually from command line with:
```
mvn -Dtest=TestLongPdbVsMmCifParsing#testLongPdbVsMmCif test
```

larsonmattr:
Jose,

 If the current SimpleMMCIFConsumer populates the SEQRES groups when
setAlignSeqRes(true), then the parsing of the SEQRES is working as
intended.  We were looking at filling the SEQRES groups for each chain
based on the _pdbx_poly_seq_scheme.  Do you know if SEQRES groups are
parsed by default when reading PDB files?  Perhaps the seAlignSeqRes()
should be a default behavior?

 Another related feature we were looking at is to be able to parse SEQRES
[...]

PR #389, opened 2016-01-26
Improved headerOnly() performance and removing FileParsingParameters.storeEmptySeqRes.

larsonmattr:
These changes are to close out issue #353 (originally mmCIF parsing of SEQRES..).  Based on the design Steve Darnell suggested, these changes do the following:

Simplify the options:
* FileParsingParameters.storeEmptySeqRes option is dropped.
* FileParsingParameters.alignSeqRes is true by default.  When false, parsing will store seqres without aligning to AtomGroups.
* FileParsingParameters.headerOnly == true indicates that parsing should only take in header information and skip sections like Atoms, Sites, and Conects.  

Parsers:  
PDBFileParser and SimpleMMcifConsumer were updated for matching behaviors to the options.
  When in headerOnly mode, parsing non-header sections return immediately to speed the parsing.

Tests:
Existing unit tests were updated for this change in the FileParsingParameters intent.  Added a new unit test to specifically test the setHeaderOnly() option and parsing of SEQRES.

Performance:
The headerOnly parsing is now greatly faster than default mode.  Prior to these changes, a 10% speed difference was found for mmCIF or PDB parsing in headerOnly mode.  After these changes with headerOnly mode for the PDB:4HHB, PDB parsing is 95% faster and mmCIF parsing is 45% faster. This can greatly speed up bulk-parsing of PDB/mmCIF files for data mining and scanning of a large PDB/mmCIF library.  (see org.biojava.nbio.structure.io.TestHeaderOnly#testSpeed2 using local files).

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/io/FileParsingParameters.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/io/FileParsingParameters.java
@@ -94,8 +94,6 @@ public class FileParsingParameters implements Serializable
 	 */
 	boolean updateRemediatedFiles;
 
-	private boolean storeEmptySeqRes;
-
 	/** 
 	 * The maximum number of atoms that will be parsed before the parser switches to a CA-only
 	 * representation of the PDB file. If this limit is exceeded also the SEQRES groups will be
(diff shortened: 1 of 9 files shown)


josemduarte:
Great stuff thanks! I'll merge it so we can test it as thoroughly as possible before the release.

PR #394, opened 2016-01-29
More robust support for MMcif parsing non-polymeric chains

larsonmattr:
This pull request is for a change to support issue #305 (More robust PDB/mmCIF parsing for non-deposited files).  I've added a unit test for how PDBFileParser and SimpleMMcifConsumer parse non-polymeric chains to TestNonDepositedFiles.java.  The default behavior of the PDBFileParser is to accept chains that have no protein or nucleic acid residues and to create from these new Chain(s) in the Structure.  We should preserve the information in a Structure for a user's supplied input (if valid).  SimpleMMcifConsumer is modified to create Compounds for an orphaned non-polymer chain.


josemduarte:
I think this looks good. It's a better solution than discarding it. I've only found a small issue that I fixed in my fork and pull requested to your branch. If you accept my PR to your branch it will appear here too. Thanks!

larsonmattr:
Thanks Jose for taking the time to look through this and add in your fix.  This change will be useful for non-conventional mmCIF files like isolated ligands, etc.

josemduarte:
I've found a small issue with this, the PDB parser is not consistent with the mmCIF parser for entries like 4a10. Weirdly in the PDB file for 4a10, the waters are assigned to a separate chain F different from all the polymers. The PDB parser seems to throw the chain away, whilst with the above patch the mmCIF parser keeps the chain.

I found the problem by running this test manually (excluded from maven in normal builds):

```
mvn -Dtest=TestLongPdbVsMmCifParsing#testLongPdbVsMmCif test
```

larsonmattr:
I see in the PDBFileParser, line 2136:
// removing water-only chains, they don't follow the standard data modeling practices. 
// We have to remove them or otherwise they can cause problems down the line, 
// e.g. 3o6j has chain Z with a single water molecule
...

Are there certain BioJava API functions that having water-only chains is a problem?  

There could be a valid reason for having such information in the file.  For example, when making the models there might be many temporary chains containing only waters or ligands.  I would argue that parsers should do their best to create Structures that are faithful to the PDB/mmCIF being parsed.  It's easy to remove chains after the Structure has been parsed but frustrating if the parsers are simply throwing away data without a reason.

josemduarte:
I agree that the parser should not throw it away, so we should try to do something about it. I'm working on it in branch  https://github.com/josemduarte/biojava/tree/nonpoly-only-chains. 

The issue is that a Chain in Biojava currently is expected to have some polymer groups (nucleotides or aminoacids), if you don't have them at all many things go wrong: there's no sequence, no representative atoms (CA or phosphates), etc.

josemduarte:
Alright the issue should be fixed with my last commits. Now all chains are parsed and new compounds created if they don't exist in file. Both for PDB and mmCIF parsers

larsonmattr:
Jose,

 Thanks for doing the further work in both parsers!  This should now be a
much better parsing of non-deposited PDB/mmCIFs.

On Tue, Feb 2, 2016 at 4:23 PM, Jose Manuel Duarte <notifications@github.com
> wrote:

> Alright the issue should be fixed with my last commits. Now all chains are
> parsed and new compounds created if they don't exist in file. Both for PDB
> and mmCIF parsers
>
> —
> <https://github.com/biojava/biojava/pull/394#issuecomment-178860561>.
>



-- 
Matt Larson, PhD
Madison, WI  53705 U.S.A.

Issue #397, opened 2016-02-01
Change the DEFAULT ChemCompProvider to a complete one


lafita:
There have been a lot of discussion lately around which **ChemComp** Provider to set as **DEFAULT**. This issue is intended to summarize the other comments about it and promote more discussion.

On the one hand, the chemical component information is essential in many cases, and can produce unexpected behaviors like in #377. It is also used in defining the type of `Group` of a Structure (Nucleotide, AminoAcid or Hetatom), which is needed for obtaining the representative Atom array #313 and secondary structure calculation #337. Other discussions are in #314 and #294.

On the other hand, the `DownloadChemCompProvider` is slower than the `ReducedChemCompProvider` and more susceptible to Network settings (less robust), but it ensures that the chemical component information is complete (unless it fails to download it).

Now that the fallback behavior of the `DownloadChemCompProvider` has been implemented #315, setting it as the default would only mean a slower `Structure` parsing, which would only be noticed when downloading a single `Structure` (for larger sets the ChemComp information would be cached).

josemduarte:
I definitely support `DownloadChemCompProvider` as the default. That is the most sensible setting for most uses. Not having the full chemical description of molecules can lead to many problems as you explain, the non-expert user will be very confused if that happens. 

Only when the user is really concerned about performance or lack of network access then it makes sense to explicitly set the `ReducedChemCompProvider`. The user will then know that if something doesn't work it is because he set the reduced provider.

andreasprlic:
Let's review the new ZipChemCompProvider, perhaps that should be the new default...

andreasprlic:
To explain some more, I think the ZipChemCompProvider should be modified to fetch components.cif.gz at first init and convert it into a zip file. Essentially this can be a replacement for the AllChemCompProvider, without needing to store the whole dictionary in memory.

lafita:
I agree with any solution (implementation) that ensures having all the Chemical Component information when needed as the default. The actual default restricts the available information to improve the efficiency.

andreasprlic:
fixed

PR #399, opened 2016-02-02
Adding ZipChemCompProvider. 

larsonmattr:
Adding ZipChemCompProvider: pulls ChemComp files from a zip archive or download.
Adding ZipAppendUtil: manages updating the zip archive with new additions.

DownloadChemCompProvider: updated to add a check if the requested ChemComp
was found or threw FileNotFoundException.  downloadChemCompRecord() returns
true if it was successful.  getChemComp now checks if the download was
successful so that it won't throw an IOException if no chemcomp exists.

TestChemCompProvider unit test checks creation/addition with ZipChemCompProvider and includes 'bad' residues that would cause DownloadChemCompProvider to fail.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/DownloadChemCompProvider.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/DownloadChemCompProvider.java
@@ -20,20 +20,29 @@
  */
 package org.biojava.nbio.structure.io.mmcif;
 
-import org.biojava.nbio.structure.align.util.HTTPConnectionTools;
-import org.biojava.nbio.structure.align.util.UserConfiguration;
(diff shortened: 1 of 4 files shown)


andreasprlic:
Thanks! should be useful... We should test this and then see if this should be the default ...

larsonmattr:
Thanks Andreas for adding!  It's been mostly tested with a large
pre-assembled chemcomp.zip library though the unit test can build the
chemcomp.zip from scratch.

On Tue, Feb 2, 2016 at 3:34 PM, Andreas Prlic <notifications@github.com>
wrote:

> Thanks! should be useful... We should test this and then see if this
> should be the default ...
>
> —
> <https://github.com/biojava/biojava/pull/399#issuecomment-178834865>.
>



-- 
Matt Larson, PhD
Madison, WI  53705 U.S.A.

lafita:
#397

PR #401, opened 2016-02-03
CE-Symm Final Stable Version

lafita:
This PR cleans the code for `CE-Symm` by removing unused methods, making static functions and improving documentation.

New features have been added:
- Package internal symmetry results in an object `CeSymmResult` that takes care of evaluating the significance of the result, order of symmetry and other properties of symmetry.
- Show **phylogenetic** trees of the structural alignments (Internal Symmetry and `MultipleAlignment`).
- Calculate StructureIdentifiers of the symmetric units (use case for viewer coloring, comparing, etc).

The algorithm after this PR will be considered a **stable version**, since the program has been run against the benchmark (1007 domains) and SCOP protein family representatives (5K domains) with the most used parameter combinations and no Exceptions occurred. The benchmark results were also checked for consistency with the true values and the prediction accuracy was within reasonable values for such parameter combinations.


andreasprlic:
great stuff!

sbliven:
Well, as stable as an algorithm can be before the reviewers get a crack at it!

PR #402, opened 2016-02-03
Added class to find weekly update information from FTP site

abradle:
Returns a Map indicate the different updates


biojava-structure/src/main/java/org/biojava/nbio/structure/rcsb/RCSBUpdates.java
```diff
@@ -0,0 +1,60 @@
...
+			String line;
+
+			// read from the urlconnection via the bufferedreader
+			while ((line = bufferedReader.readLine()) != null)
+			{
+				outList.add(line);
+			}
+			bufferedReader.close();
+		}
+		catch(Exception e)
```
josemduarte: I would rather throw the IOException forward so that it can be dealt with properly at a later stage. Here you will end up with an empty list which is not showing clearly if it comes from an empty file.

biojava-structure/src/main/java/org/biojava/nbio/structure/rcsb/RCSBUpdates.java
```diff
@@ -0,0 +1,60 @@
...
+import java.net.URLConnection;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class RCSBUpdates {
+
+
+	public Map<String,String[]> getUpdates(){
```
josemduarte: Could you add some minimal javadoc here? It is not obvious what the return map contains

andreasprlic:
it would be also nice to get the updates from a specific week, and not just only the latest week...

andreasprlic:
Merged in for now. Would still be nice to be able to get specific weeks at some point in the future.

PR #403, opened 2016-02-03
Bumping up java version as discussed in mailing list

josemduarte:
In preparation for next release, this bumps up the java version to 1.7

See discussion in mailing list here: http://mailman.open-bio.org/pipermail/biojava-l/2016-January/011445.html


andreasprlic:
As discussed on the mailing list we are now on Java 1.7.

PR #405, opened 2016-02-09
Fix resource leaks with ChemComp

### LINKED ISSUE #395: Exception when too many files requested to AtomCache
When the same **AtomCache** is used to request for many structures (not at the same time) an Exception is thrown and the following message is given: **Too many open files**. 

It could indicate a resource or memory leak in **AtomCache**, by maintaining some files open (although the structure has already been downloaded). I get those Exceptions when downloading, in parallel 4 at the same time, more than 2K structures. I do not remember the Exception when I calculated the census the first time (6 months ago), although I am not sure if I was using exactly the same code. Any ideas?

This is the **stacktrace** of the Exception:

```java
java.io.FileNotFoundException: /scratch/cache/data/structures/divided/mmCIF/rl/1rlr.cif.gz (Too many open files)
	at java.io.FileInputStream.open(Native Method) ~[?:1.6.0_38]
	at java.io.FileInputStream.<init>(FileInputStream.java:140) ~[?:1.6.0_38]
	at org.biojava.nbio.core.util.FlatFileCache.addToCache(FlatFileCache.java:63) [classes/:?]
	at org.biojava.nbio.core.util.InputStreamProvider.getInputStreamFromFile(InputStreamProvider.java:247) [classes/:?]
	at org.biojava.nbio.core.util.InputStreamProvider.getInputStream(InputStreamProvider.java:163) [classes/:?]
	at org.biojava.nbio.structure.io.LocalPDBDirectory.getInputStream(LocalPDBDirectory.java:377) [classes/:?]
	at org.biojava.nbio.structure.io.LocalPDBDirectory.getStructureById(LocalPDBDirectory.java:343) [classes/:?]
	at org.biojava.nbio.structure.align.util.AtomCache.loadStructureFromCifByPdbId(AtomCache.java:1038) [classes/:?]
	at org.biojava.nbio.structure.align.util.AtomCache.getStructureForPdbId(AtomCache.java:1019) [classes/:?]
	at org.biojava.nbio.structure.scop.ScopDomain.loadStructure(ScopDomain.java:245) [classes/:?]
	at org.biojava.nbio.structure.align.client.StructureName.loadStructure(StructureName.java:349) [classes/:?]
	at org.biojava.nbio.structure.align.util.AtomCache.getStructure(AtomCache.java:394) [classes/:?]
	at demo.CeSymmMain$CeSymmWorker.run(CeSymmMain.java:1061) [classes/:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471) [?:1.6.0_38]
	at java.util.concurrent.FutureTask$Sync.innerRun(FutureTask.java:334) [?:1.6.0_38]
	at java.util.concurrent.FutureTask.run(FutureTask.java:166) [?:1.6.0_38]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1146) [?:1.6.0_38]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615) [?:1.6.0_38]
	at java.lang.Thread.run(Thread.java:701) [?:1.6.0_38]
``

### PR DESCRIPTION
**Pull request by sbliven**
Adds additional check that files and streams are being closed in the ChemComp providers.

Also changes the ChemCompGroupFactory's behavior so that it does cache "empty"/stub chemicals
(since they will still be empty for the next fetch), but clears the cache when the provider changes.
This ensures that stub chemicals from a Reduced provider don't poison future fetches with the full
Download provider.

This should fix #395. Waters were being treated as "empty" due to an unfortunate choice of
one-letter code, and therefor not cached, leading to thousands of open handles to HOH.cif.gz.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/ChemCompGroupFactory.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/ChemCompGroupFactory.java
@@ -57,10 +57,8 @@ public class ChemCompGroupFactory {
 		logger.debug("Chem comp "+recordName+" read from provider "+chemCompProvider.getClass().getCanonicalName());
 		cc = chemCompProvider.getChemComp(recordName);
 		
-		// If chemCompProvider fails don't try to cache null & one letter code may be null.
-		if (null != cc && !cc.isEmpty()){
-			cache.put(recordName, cc);
-		}
+		// Note that this also caches null or empty responses
+		cache.put(recordName, cc);
 		return cc;
 	}
 
@@ -79,6 +77,8 @@ public class ChemCompGroupFactory {
 	public static void setChemCompProvider(ChemCompProvider provider) {
 		logger.debug("Setting new chem comp provider to "+provider.getClass().getCanonicalName());
 		chemCompProvider = provider;
+		// clear cache
+		cache.clear();
 	}
 
 	public static ChemCompProvider getChemCompProvider(){
@@ -153,10 +153,5 @@ public class ChemCompGroupFactory {
 		}
 		return oneLetter;
 	}
-
-	public static SoftHashMap<String, ChemComp> getCache() {
-		return cache;
-	}
-
 
 }
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/ChemCompGroupFactory.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/ChemCompGroupFactory.java
@@ -58,7 +58,7 @@ public class ChemCompGroupFactory {
 		cc = chemCompProvider.getChemComp(recordName);
 		
 		// If chemCompProvider fails don't try to cache null & one letter code may be null.
-		if (null != cc && !"?".equals(cc.getOne_letter_code())){
+		if (null != cc && !cc.isEmpty()){
 			cache.put(recordName, cc);
 		}
 		return cc;
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/DownloadChemCompProvider.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/DownloadChemCompProvider.java
@@ -166,62 +166,57 @@ public class DownloadChemCompProvider implements ChemCompProvider {
 		int counter = 0;
 		InputStreamProvider prov = new InputStreamProvider();
 
-		InputStream inStream = prov.getInputStream(f);
+		try( BufferedReader buf = new BufferedReader (new InputStreamReader (prov.getInputStream(f)));
+				) {
+			String line = null;
+			line = buf.readLine ();
+			StringWriter writer = new StringWriter();
 
-		BufferedReader buf = new BufferedReader (new InputStreamReader (inStream));
+			String currentID = null;
+			while (line != null){
 
-		String line = null;
-		line = buf.readLine ();
-		StringWriter writer = new StringWriter();
+				if ( line.startsWith("data_")) {
+					// a new record found!
 
-		String currentID = null;
-		while (line != null){
+					if ( currentID != null) {
+						writeID(writer.toString(), currentID);
+						counter++;
+					}
 
-			if ( line.startsWith("data_")) {
-				// a new record found!
-
-				if ( currentID != null) {
-					writeID(writer, currentID);
-					counter++;
+					currentID = line.substring(5);
+					writer = new StringWriter();
 				}
 
-				currentID = line.substring(5);
-				writer = new StringWriter();
+				writer.append(line);
+				writer.append(NEWLINE);
+
+				line = buf.readLine ();
 			}
 
-			writer.append(line);
-			writer.append(NEWLINE);
+			// write the last record...
+			writeID(writer.toString(),currentID);
+			counter++;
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/DownloadChemCompProvider.java [hunk 1]
```diff
@@ -166,62 +166,57 @@ private void split() throws IOException {
...
 		InputStreamProvider prov = new InputStreamProvider();
 
-		InputStream inStream = prov.getInputStream(f);
-
-		BufferedReader buf = new BufferedReader (new InputStreamReader (inStream));
+		try( BufferedReader buf = new BufferedReader (new InputStreamReader (prov.getInputStream(f)));
+				) {
+			String line = null;
+			line = buf.readLine ();
+			StringWriter writer = new StringWriter();
```
josemduarte: Nice! I didn't know you could do this kind of thing with Java 7. This is the reference for those that don't know (like me!): https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/ReducedChemCompProvider.java
```diff
@@ -46,54 +46,46 @@
...
-		if ( inStream == null){
-			//System.out.println("Could not find chem comp: " + name + " ... using generic Chem Comp");
-			// could not find the chem comp definition for this in the jar file
-			logger.debug("Getting empty chem comp for {}",name);
-			ChemComp cc = getEmptyChemComp();
-			cc.setId(name);
-			return cc;
-		}
+		try {
+			try(InputStream inStream = this.getClass().getResourceAsStream("/chemcomp/"+name + ".cif.gz")) {
```
josemduarte: Shouldn't the try-with be inside the first try? The second try is redundant, isn't it?

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/DownloadChemCompProvider.java
```diff
@@ -364,29 +359,20 @@ private static boolean downloadChemCompRecord(String recordName) {
...
 			HttpURLConnection uconn = HTTPConnectionTools.openHttpURLConnection(url);
 
-			InputStream conn = uconn.getInputStream();
-
-			FileOutputStream outPut = new FileOutputStream(localName);
-
-			GZIPOutputStream gzOutPut = new GZIPOutputStream(outPut);
+			try( PrintWriter pw = new PrintWriter(new GZIPOutputStream(new FileOutputStream(localName)));
+					BufferedReader fileBuffer = new BufferedReader(new InputStreamReader(uconn.getInputStream()));
+					) {
```
josemduarte: This try is redundant, isn't it? The try-with could go above with the first try.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/DownloadChemCompProvider.java
sbliven: Yep! I'm going to milk the 1 month of Java 7 development for all its worth. I'm also using [type inference](https://docs.oracle.com/javase/tutorial/java/generics/genTypeInference.html) in all my code from now on.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/DownloadChemCompProvider.java
sbliven: In this case I think not, because `openHttpURLConnection` can throw the exceptions as well. I wanted the try-with to surround only the stream code.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/ReducedChemCompProvider.java
sbliven: In this case it is redundant. I wasn't sure if you could combine catch blocks with try-with, but apparently you can.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/ChemCompGroupFactory.java
```diff
@@ -79,6 +77,8 @@ public static ChemComp getChemComp(String recordName){
 	public static void setChemCompProvider(ChemCompProvider provider) {
 		logger.debug("Setting new chem comp provider to "+provider.getClass().getCanonicalName());
 		chemCompProvider = provider;
+		// clear cache
+		cache.clear();
```
sbliven: Are there any potential problems here? I know there was discussion that we should do this dating back to before github, so I was very surprised that it hadn't been done.

josemduarte: Yes this looks familiar... I don't really remember what was the problem exactly but I'm pretty sure we've had a lot of subtle issues around how the cache is initialised and cleared. See for instance #313 

Let's anyway merge it so that we can test it for a while before the release. I know that @abradle is doing a lot of parallel parsing of files and he would be able to find problems with this.

andreasprlic:
nice catch!

PR #409, opened 2016-02-19
Implement missing TER records in the Structure.toPDB()

larsonmattr:
This is pull request is to add "TER" records after chains when writing a PDB.  The PDB standard (http://deposit.rcsb.org/adit/docs/pdb_atom_format.html) describes that "The list of ATOM records in each polymer chain must be terminated by a TER record".  
* FileConvert writes "TER" line after each chain written out to the PDB file.
* Updated the PDBFileParserTest(s) to look also for the "TER" line when checking the PDB file output.

Also includes a check to prevent NPE in SimpleMMcifParser when trying to read a header for a zero-length file.


josemduarte:
Thanks! :)

PR #414, opened 2016-02-25
Add getAsymId on a group level

abradle:
Get the chain AsymId from the group level - as these are not available otherwise. (Chain AsymId is just the first AsymId of the groups in this AuthId chain).


andreasprlic:
Hi, not happy about having chain IDs exposed on the Group level. Let's discuss a better way of representing this relationship.

andreasprlic:
I'd vote for either a new "InternalChain" class that maps groups to internal chain IDs on the group level, or to resolve this at the file parsing level and request a Structure object that breaks up chain based on internal CHain IDs, and not on public chain IDs.

abradle:
Updated - so instead the function alters the file parsing param and allows for making chains on the basis of asym ids

PR #419, opened 2016-03-03
improved qualifier managemnt

stefanharjes:
This patch improves the organization of qualifiers. It stores all qualifier entries in a string[] in the qualifier itself, which also holds only one name. To improve readability a new map was implemented. This means it is now deprecated to say `feature.add(qualifierName, qualifier)` please use instead `feature.add(qualifier)`. Most of the map functions are mirrored in the `Feature` and the `FeatureInterface`, so you do not notice the map if you extend 'AbstractFeature'. Also included is an updated `DBReferenceInfo` (which expands `Qualifier`) which is further implemented in the above map and the `FeatureInterface` this rendering `DBReferenceInterface` deprecated. 

This patch simplifies the qualifier management, improves readability and performance. All tests have passed.


biojava-core/src/main/java/org/biojava/nbio/core/sequence/features/DatabaseReferenceInterface.java
```diff
@@ -30,7 +30,9 @@
  * has a list of cross reference id(s)
  * @author Scooter Willis <willishf at gmail dot com>
  */
+@Deprecated  //use feature interface
```
josemduarte: Can you give the tip for what to use without comments? Like that it won't show in javadoc, i.e. something like 
`@Deprecated Use Feature interface instead`

josemduarte: Sorry that's wrong, you should add that in the javadoc:
```java
/**
 * .....
 *
 * @Deprecated use Feature interface instead
 */
```

biojava-core/src/main/java/org/biojava/nbio/core/sequence/features/FeatureInterface.java
```diff
@@ -169,25 +169,35 @@
...
+	public String[] getFirstDatabaseReferenceInfo();
+	public String getFirstDatabaseReference();
+	public String getFirstDatabase();
+	public String getDatabaseReference(String database, int i);
+	public String[] getAllDatabaseReferences(String database);
+	public String getFirstDatabaseReference(String database);
+	public void setDatabaseReferenceInfo(String database, String reference);
+	public boolean addDatabaseReferenceInfos(String[][] entries);
+	//old stuff to be removed
+	@Deprecated //use addQualifier(Qualifier q)
```
josemduarte: Can you add some javadoc for the deprecated method to point to which one to use instead

biojava-core/src/main/java/org/biojava/nbio/core/sequence/loader/UniprotProxySequenceReader.java
```diff
@@ -638,26 +640,33 @@ public String getOrganismName() {
...
+        		//}
+        		//DBReferenceInfo dbreferenceInfo = new DBReferenceInfo(type, id);
         		ArrayList<Element> propertyElementList = XMLHelper.selectElements(element, "property");
         		for (Element propertyElement : propertyElementList) {
         			String propertyType = propertyElement.getAttribute("type");
         			String propertyValue = propertyElement.getAttribute("value");
-        			dbreferenceInfo.addProperty(propertyType, propertyValue);
+        			//dbreferenceInfo.addProperty(propertyType, propertyValue);
+        			
+        			System.out.println("here comes a dbxref, with: type: "+type+", id: "+id+", propertyType: "+propertyType+", propertyValue: "+propertyValue);
```
josemduarte: Use a logger here (from slf4j api). I guess it should be debug level here

stefanharjes: While addressing the above issues, I encounter another issue, which will take a bit, so patience please.

andreasprlic:
Can somebody please resolve the merge conflicts?

andreasprlic:
Thanks @stefanharjes ! Do you have any more commits coming for this, or are you done?

heuermh:
The commits here should be rebased and squashed down to a reasonable number of changes with good commit messages.  As it is there's no way to review this.

stefanharjes:
The main problem was, that I was testing the change and working on it. In the meantime there was a release which changed all spacings and tabs and thus caused hundreds of merge conflicts. Git seems to be unable to discriminate between simple line shifts and actual edits. Or at least I did not find the correct switch. I also did not find a macro how you handle the space/tab format. When I was done, I waited too long or maybe forgot to do the push and it all happened again. I was also unaware, that when pushing to my branch, it would automatically update the commit (which I intended to clean up before submition). 

We could in principle delete the commit and I try to do another pull request once I am satisfied. It would be very helpful, if there is one standard macro to enforce space/tab formatting.
Regards
 

    Michael L Heuer <notifications@github.com> schrieb am 2:54 Freitag, 17.Juni 2016:
 

 The commits here should be rebased and squashed down to a reasonable number of changes with good commit messages. As it is there's no way to review this.—

andreasprlic:
I agree about the code formatting macro. We used to have an eclipse code formatting template, however I could not quickly find that. Perhaps time to start a new one! (also would work in IntelliJ). Alternative, set up something like [this](https://github.com/michalrus/git-hooks-code-autoformat).

Issue #408, opened 2016-02-13
Add more `_struct_conn` parsing/consuming from mmCIF


arose:
As a followup to https://github.com/biojava/biojava/issues/343 which added SSBonds I suggest to add more bond data present in the mmCIF files.

Not covalent bonds, so ignore (`_struct_conn.conn_type_id`)?
```
                // hydrog - hydrogen bond
                // mismat - mismatched base pairs
                // saltbr - ionic interaction
```

Process (`_struct_conn.conn_type_id`)!
```
                // covale - covalent bond
                // covale_base -
                //      covalent modification of a nucleotide base
                // covale_phosphate -
                //      covalent modification of a nucleotide phosphate
                // covale_sugar -
                //      covalent modification of a nucleotide sugar
                // disulf - disulfide bridge
                // metalc - metal coordination
                // modres - covalent residue modification
```

josemduarte:
After some discussion we decided that the only important ones (consistent enough across the PDB archive) are:
```
covale
covale_base
covale_phosphate
covale_sugar
modres
```
Plus `disulf` which we already have.

PR #421, opened 2016-03-04
Improving bonds support

josemduarte:
- Fixes issues #341 and #408 .
- It includes an API change in `Structure.getSSBonds()` which now returns a `List<Bond>` instead of a `List<SSBond>`. 
- It deprecates `Structure.getConnections()` and `Structure.setConnections()` which were tied to the PDB file representation. The already existing `Bond` class and `Atom.getBonds()` are a much better way to access and handle the bonds. 
- The new parsing parameter introduced recently `isCreateLigandConnects` has been removed, since it was redundant. All bonds parsing is controlled by a single parameter `createAtomBonds` (false by default).
- Should deal correctly with alt locations in ssbonds
- Doesn't deal yet with bonds between symmetry partners, that can only be done after introduction of better symmetry support #220 
- All bonds parsing code has been unified into the (already existing) class `BondMaker`
- A few bug fixes


lafita:
`TestParseMmCIFLigands `is failing.

josemduarte:
Yes I saw that, but it's really weird. If I try locally it works... Let me see if I can tell travis to restart it

PR #423, opened 2016-03-06
Revised calls to multipleSequenceAlignment.getAlignedSequence()

kevinmcain:
Revised calls to multipleSequenceAlignment.getAlignedSequence(). Now passing in index plus 1 as the MultipleSequenceAlignment.getAlignedSequence() function uses sequences.get(listIndex - 1). There is a comment on the related function header "bioIndex starting at 1 instead of 0." Let me know if I'm missing something, but it seems pretty clear if a function is called passing in a value of index = 0, then that function is going to blow up when accessing a collection using index -1.


kevinmcain:
I guess I should have checked the master repository first. It looks like a lot has changed since I initially identified the index out of range issue...

lafita:
I noticed this Exception when restructuring the **phylo** module and I fixed it. See ths issue #346 and pull request #370. 

The **phylo** module was not very extensively developed and contained copied pieces of code from different sources, in addition to a buggy **forester** version as core dependency. The code has been unified with the rest of **biojava**, the dependencies have been updated and many additional methods for phylogenetic analysis have been created. All changes will be included into the coming 4.2 release this week.

PR #424, opened 2016-03-07
Zipchem

larsonmattr:
Jose suggested a clean up of ZipChemCompProvider for exception handling in general.  The ZipFileSystem now available in Java 7 can manage the archiving and appending without backup copies of the archive, so I've switched to using this instead of ZipOutputStream and simplified the exception handling and error reporting.  Unit testing with ZipChemCompProvider will no longer remove .cif.gz so as to avoid repeat downloading when running the tests involving ZipChemCompProvider and DownloadChemCompProvider.  The class is threadsafe with synchronized methods for reading or appending to the archive.


biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/ZipChemCompProvider.java
```diff
@@ -38,59 +38,83 @@
...
+ * updated 3/5/2016 for Java 7 ZipFileSystem
  */
 public class ZipChemCompProvider implements ChemCompProvider{
-
-	private static final String FILE_SEPARATOR = System.getProperty("file.separator");
-	private String tempdir = "";  // Base path where $zipRootDir/ will be downloaded to.
-	private String zipRootDir = "chemcomp/"; 
-	private ZipFile zipDictionary;
-	private String zipFile;
 	private static final Logger s_logger = Logger.getLogger( ZipChemCompProvider.class.getPackage().getName() );
```
josemduarte: This should be slf4j Logger, not the java.util.logging one.

biojava-structure/src/test/java/org/biojava/nbio/structure/io/mmcif/TestChemCompProvider.java
```diff
@@ -7,14 +7,19 @@
...
+import org.biojava.nbio.structure.StructureException;
+import org.biojava.nbio.structure.StructureIO;
+import org.biojava.nbio.structure.align.util.AtomCache;
 import org.biojava.nbio.structure.align.util.UserConfiguration;
 import org.biojava.nbio.structure.io.FileParsingParameters;
 import org.biojava.nbio.structure.io.PDBFileReader;
 import org.junit.Test;
 
 public class TestChemCompProvider {
+	private static final Logger s_logger = Logger.getLogger( TestChemCompProvider.class.getPackage().getName() );
```
josemduarte: Same here, use slf4j logger as we do elsewhere in biojava. Otherwise the logging config won't work well for this.

josemduarte:
Looks good! If you can fix the logging and make it slf4j logging, we can then merge it in. Just in time for 4.2 :)

larsonmattr:
Oops! Good point -  I've switched these two classes to use the slf4j logger
abstraction instead. Thanks!

On Mon, Mar 7, 2016 at 11:40 AM, Jose Manuel Duarte <
notifications@github.com> wrote:

> Looks good! If you can fix the logging and make it slf4j logging, we can
> then merge it in. Just in time for 4.2 :)
>
> —
> <https://github.com/biojava/biojava/pull/424#issuecomment-193364442>.
>



-- 
Matt Larson, PhD
Madison, WI  53705 U.S.A.

PR #427, opened 2016-03-09
Making classes serializble and fixing issues with parsing

abradle:


biojava-alignment/src/main/java/org/biojava/nbio/alignment/SimpleGapPenalty.java
```diff
@@ -30,9 +32,13 @@
  *
  * @author Mark Chapman
  */
-public class SimpleGapPenalty implements GapPenalty {
+public class SimpleGapPenalty implements GapPenalty, Serializable {
```
heuermh: wish this pull request hadn't been merged so quickly, making non-final public classes Serializable is somewhat dangerous, see e.g. the chapter on serialization in Effective Java.

biojava-alignment/src/main/java/org/biojava/nbio/alignment/SimpleGapPenalty.java
```diff
@@ -30,9 +32,13 @@
...
 
-    private static int dgop = 10, dgep = 1;
+    /**
+	 * 
+	 */
+	private static final long serialVersionUID = 3945671344135815456L;
```
heuermh: ...and useless javadoc and odd formatting

abradle: Great spot. I've made another pull request to fix this. Thanks.

biojava-alignment/src/main/java/org/biojava/nbio/alignment/SimpleGapPenalty.java
abradle: Thanks for this. I'm afraid it's necessary in this case. But yes I agree with your caution.

heuermh: I saw the commit messages "Make classes serializable for Spark" earlier and reached out to see if might help (I currently work for the UC Berkeley AMPLab on genomics applications for Apache Spark).

Making these classes serializable in this way is not only unsafe, and may cause future problems with API compatibility, it is not necessarily the most efficient way to use Biojava classes in Spark.  A lot of static and non-threadsafe state including matrices and caches will also be serialized, leading to significant overhead and potential concurrency problems.

abradle: Hi Michael,

Thanks for reaching out. And you're right there are potential issues with
doing this.

I'd be interested to hear (offline or on here) how you've tackled this in
AMPLab.

Best wishes,

Anthony

andreasprlic:
Why does the BondMaker need to be part of this?

abradle:
It's part of the fixing issues with parsing.

josemduarte:
The BondMaker changes were to fix an issue introduced in #418

andreasprlic:
Ok, got it. This is actually unrelated to serialisation in Spark. In any case, as Michael says, let's slow down, review everything tomorrow and make sure all these patches don't break everything. 

@heuermh , we should have a Skype call and share Spark experiences!

heuermh:
That sounds great, or better yet an excuse to visit!

andreasprlic:
Visit would be best! when are you in SD next time?

heuermh:
I could come out in April or early May, will start an email thread.

PR #428, opened 2016-03-09
Update to remove docstring

abradle:


heuermh:
I know we're trying to lock down before a release, but can we please slow down a bit to allow for review?  This pull request only fixes one of five of the same problem with #427.

josemduarte:
There was not much to review in this one, was there? The fix for the other issue can still come in separate commits. Apologies in any case for merging #427 so quickly.

Anyway regarding that Serializable issue, it's a good point that non-final public classes shouldn't be serializable, but the fact is that it is by far not the only instance in biojava of such thing. We could discuss that in a separate issue and correct all instances of it if that's deemed important enough.

heuermh:
Agreed.

I find it easier to discuss while reviewing a pull request than having to roll back or otherwise fix things that were merged too quickly.

andreasprlic:
The code free was planned for Monday. I think we'll delay the release for a day since we had too many patches AFTER the freeze. This will give us more time for housekeeping tasks and testing the latest changes. Better make sure that nothing got broken in the last moment.

Issue #337, opened 2015-10-19
Using hasAminoAtoms() for modified AA


lafita:
Since the pull request #314 the `ReducedChemCompProvider` is the default behaviour. This generated some problems in parsing structures with modified aminoacids, because they are stored in `Hetatom` objects instead of `AminoAcids`. This was a problem when trying to extract the representative `Atom` Array of a `Structure`, the array being of different length when using the `DownloadChemCompProvider` (includes the modified AA) or the `ReducedChemCompProvider` (does not include them). 

A possible solution to this issue could be to use the method **Group.hasAminoAtoms()**, instead of **instanceof AminoAcid**, since the method is independent of the Group implementation. From my point of view, the use of the `ChemCompProvider` should only affect the information of the aminoacid properties, but not fundamental methods like the representative array of a Structure.

I propose to use the method **Group.hasAminoAtoms()** consistently every time a Group needs to be checked for AminoAcid type, instead of the **instanceof AminoAcid** (which should only be used if the specific methods of the AminoAcid interface need to be used). This is currently done in the secondary structure annotation. Anyone sees a problem in the change?

andreasprlic:
One concern is C-alpha only protein structures. The proposed behavior would not work for those.

pwrose:
Proper handling of polymer chains (and non-polymers) will also be very
important for our data compression project. Perhaps we can discuss the
various issues around handling polymers and come up with a solution in our
meeting on Wed.

On Mon, Oct 19, 2015 at 10:21 AM, Andreas Prlic <notifications@github.com>
wrote:

> One concern is C-alpha only protein structures. The proposed behavior
> would not work for those.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/biojava/biojava/issues/337#issuecomment-149287288>.
>



-- 
Peter Rose, Ph.D.
Site Head, RCSB Protein Data Bank West (http://www.rcsb.org)
San Diego Supercomputer Center (http://bioinformatics.sdsc.edu)
University of California, San Diego
+1-858-822-5497

josemduarte:
I'd also have some other concerns about that:
- C-terminal aminoacids won't have all of the amino atoms as defined in `Group.hasAminoAtoms()` (C, N, CA, O) since O is replaced by OXT
- Non-standard aminoacids must not necessarily follow the convention of naming the amino atoms as C, N, CA, O
- Other compounds that are not aminoacid-related might use the names just by chance (nothing prohibits that)

[...]

Issue #145, opened 2014-08-11
ChemCompGroupFactory not respecting the ChemCompProvider if value is already cached


josemduarte:
This is a subtle issue with singletons in the structure module. I've only found it because of a strange inconsistency while parsing both mmCIF and PDB files in same JVM. 

In `ChemCompGroupFactory` the `ChemCompProvider` can be set with `setChemCompProvider` (e.g. it is set to `ReducedChemCompProvider` in `PDBFileParser` when `isLoadChemCompInfo` is false), however if the cache in `ChemCompGroupFactory` is already initialised then the `ChemCompProvider` setting is ignored and the value from the cache is taken instead. 

For instance the PDB file parser sets it by default to `ReducedChemCompProvider` while the mmCIF parser sets it by default to `DownloadChemCompProvider`. If one uses the mmCIF parser first and then the PDB parser the setting to `ReducedChemCompProvider` is ignored because the cache contains already the value coming from the `DownloadChemCompProvider` (i.e. from the file in `/tmp/chemcomp/xxx.cif.gz`)

One way to reproduce it is with this code:

```
String pdbCode = "1oew";
Structure structure = null;

AtomCache cache = new AtomCache();  
cache.setUseMmCif(true);
structure = cache.getStructure(pdbCode);

cache.setUseMmCif(false);
FileParsingParameters params = new FileParsingParameters();
params.setAlignSeqRes(true);
cache.setFileParsingParams(params);

structure = cache.getStructure(pdbCode);

```

That produces an error because of an issue in the alignment of SEQRES to ATOM. 

However this doesn't produce an error:

```
String pdbCode = "1oew";
Structure structure = null;

AtomCache cache = new AtomCache();  
cache.setUseMmCif(false);
FileParsingParameters params = new FileParsingParameters();
params.setAlignSeqRes(true);
cache.setFileParsingParams(params);

structure = cache.getStructure(pdbCode);

```

PR #429, opened 2016-03-09
Remove load chem comp info

### LINKED ISSUE #426: loadChemCompInfo system property is used inconsistently
A `loadChemCompInfo` system property (see constant in `PDBFileReader.LOAD_CHEM_COMP_PROPERTY`) is used mostly internally to sign that we should load chemical component descriptions. The property interferes with the chemical component provider mechanism and in it looks like it's doing more bad than good. 

For instance in `ChainImpl.getAtomSequence()` and `ChainImpl.getSeqResSequence()` the chemical components are used only if the property is true, whilst the behaviour should depend on the chemical component provider used. In fact if one removes [this condition](https://github.com/biojava/biojava/blob/ff66c26573022cfba2cfa6a025b58435b5c9a960/biojava-structure/src/main/java/org/biojava/nbio/structure/ChainImpl.java#L608), other bugs surface that were hidden by `loadChemCompInfo` being false.

This relates to issues #313 and #397 

I'm working on a pull request. Maybe it can still go into the 4.2 release.

josemduarte:
This fixes #426 . A few issues were hidden by the loadChemCompInfo setting that I've fixed in the commits below.


andreasprlic:
This comes in after the code freeze. We can delay the release by a day so this can still get in. However let's review tomorrow and make sure that these changes don't break anything in the last moment:

sbliven:
Seeing as this removes methods, would it be more suited for the 5.0 release anyways?

josemduarte:
The one method that has been removed with this pull request is `FileParsingParameters.setLoadChemCompInfo()`. The effect of calling that was inconsistent (it would switch on/off some chemcomp behavior but not other). I thought it was better to get a compile error so that the BioJava user knows that the method doesn't exist anymore and doesn't do what they'd expect.

Of course I can just put it back and add a deprecation tag to it saying that it doesn't do anything. Would that be preferred?

josemduarte:
To be clear this issue is related to the many we've had with the chemical component framework (#313, #337, #397, #145 ). We've fixed many of them for 4.2 but I feel without this fix the thing would be incomplete and more issues would come in the future. The framework is essential for the functioning of the whole structure module so I think it's worth trying getting this right.

lafita:
I agree with @josemduarte, the behavior of this method was never very clear. There was also an issue reported long ago in the mailing list and a temporal fix was introduced that changed static objects of the **ChemCompFactory** through this method.

However, how would the user deactivate the loading of chemical component information? By manually setting the system property, like:
```java 
System.setProperty(PDBFileReader.LOAD_CHEM_COMP_PROPERTY, "false");
```

andreasprlic:
Travis fails since it seems EBI ftp is not working well. I'll merge this anyways, since  I compiled the branch locally and verified that everything else works as expected.

josemduarte:
One note: when merging #431 into this I had conflicts in all files that I had touched. To solve them quickly I had to accept my versions of the files. That means that those files will not be re-formatted with @sbliven's #431 changes. If we are really concerned about that, the re-formatting scripts would have to be rerun on those.

andreasprlic:
Let's release tomorrow, this is not worth delaying the release.

PR #431, opened 2016-03-09
Major Housekeeping

sbliven:
Formatting and cleaning commits.
- Convert all files to unix line endings
- Run addlicense.sh
- Remove whitespace at the end of lines
- Add missing @Override
- remove unneeded imports
- Reformat leading spaces to tabs (semimanual process using development/spacestotabs.py)


sbliven:
I'll try to squeeze in one more change, but if I don't feel free to merge this if needed

andreasprlic:
Thanks Spencer!

sbliven:
Thanks for merging Andreas.

I also considered just letting eclipse reformat everything according to our formatting files. This is very easy (select all projects, Source>Clean Up...). Unfortunately it touches nearly every file, does a lot of line wrapping in inconvenient places, and makes many comments much uglier. So I went with the much more tedious approach above.

I did push 38934ac, which goes along with this pull request

aalhossary:
My understanding is that Eclipse controls the formatting through XML files.

Couldn’t we just disable line wrapping and apply the autoformat?

sbliven:
@aalhossary Yes, we could do that, although line wrapping is enabled in the current profile. I think a bigger problem was places where comments were formatted as fixed-spaced <pre> blocks, where auto indenting would lose the alignment. I cleaned up all these cases and formatted them as normal javadocs, so we could probably be a bit more aggressive now with the autoformatter.

I'm satisfied with the current state of the code and am disinclined to spend any more time on this issue.

PR #433, opened 2016-03-11
Upgrading minimum requirement to java 8

josemduarte:
Upgrading to Java 8 for next developing cycle towards the 5.0 release, [as discussed in the mailing list](http://mailman.open-bio.org/pipermail/biojava-l/2016-January/011445.html)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,5 @@
 language: java
+jdk:
+  - oraclejdk8
 sudo: false
+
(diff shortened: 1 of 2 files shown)


heuermh:
In addition to bumping the java version in pom.xml, we may also want to update the travis build to jdk8-only and add a maven enforcer plugin configuration that lets the user know early if they try to build with earlier than jdk8, e.g. https://github.com/heuermh/dishevelled-bio/blob/master/pom.xml#L129

josemduarte:
Great suggestions! I'll add that to the PR

PR #437, opened 2016-03-16
Add parameterized constructor to NCBIQBlastService to support BLAST on…

christiam:
… the cloud service providers


josemduarte:
Thanks! This should go with Biojava 5 (possibly a few months from now). Do you think it is worth having it for 4.2.1? In that case we can still cherry pick it. We should be able to release 4.2.1 pretty quickly.

christiam:
@josemduarte IMHO yes, as BioJava users may benefit from it.

PR #441, opened 2016-03-23
Refactored compound to entity and altered behaviour to consider non-polymer compounds.

abradle:
1) Refactored Compound to EntitiyInfo -> it seemed a better descriptiton.

2) Altered behaviour to consider non-polymer entities. 
This was required by MMTF and is I believe desired.

3) Added extra information (description and type).
These were required by MMTF and also generally useful.

### DIFF
--- a/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/StructureTest.java
+++ b/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/StructureTest.java
@@ -119,7 +119,7 @@ public class StructureTest {
 		assertEquals(64, c2.getAtomGroups(GroupType.HETATM).size());
 		assertEquals(0, c2.getAtomGroups(GroupType.NUCLEOTIDE).size());
 
-		List<EntityInfo> compounds= structure.getEntityInformation();
+		List<EntityInfo> compounds= structure.getEntityInfos();
 
 		// from Biojava 4.2 on we are creating compounds whenever an entity is found to be without an assigned compound in the file
 		// see issues https://github.com/biojava/biojava/issues/305 and https://github.com/biojava/biojava/pull/394
@@ -203,7 +203,7 @@ public class StructureTest {
 		assertEquals("the technique in the Header is " + technique, techShould, technique);
 
 
-		List <EntityInfo> compounds = structure.getEntityInformation();
+		List <EntityInfo> compounds = structure.getEntityInfos();
 
 		// from Biojava 4.2 on we are creating compounds whenever an entity is found to be without an assigned compound in the file
 		// see issues https://github.com/biojava/biojava/issues/305 and https://github.com/biojava/biojava/pull/394
--- a/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/StructureTest.java
+++ b/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/StructureTest.java
@@ -119,13 +119,13 @@ public class StructureTest {
 		assertEquals(64, c2.getAtomGroups(GroupType.HETATM).size());
 		assertEquals(0, c2.getAtomGroups(GroupType.NUCLEOTIDE).size());
 
-		List<Compound> compounds= structure.getCompounds();
+		List<EntityInfo> compounds= structure.getEntityInformation();
 
 		// from Biojava 4.2 on we are creating compounds whenever an entity is found to be without an assigned compound in the file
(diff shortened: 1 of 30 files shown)


biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/StructureTest.java
```diff
@@ -119,13 +119,13 @@ public void testReadPDBFile() throws Exception {
 		assertEquals(64, c2.getAtomGroups(GroupType.HETATM).size());
 		assertEquals(0, c2.getAtomGroups(GroupType.NUCLEOTIDE).size());
 
-		List<Compound> compounds= structure.getCompounds();
+		List<EntityInfo> compounds= structure.getEntityInformation();
```
andreasprlic: Shouldn't this be structure.getEntityInfo() ? getter and setter methods should use the same names as the classes, to make it easier to work with Reflection.

biojava-structure/src/main/java/org/biojava/nbio/structure/Chain.java
```diff
@@ -195,15 +195,15 @@
...
 
 	/**
 	 * Returns the Compound for this chain.
 	 *
 	 * @return the Compound object
-	 * @see #setCompound(Compound)
+	 * @see #setCompound(EntityInfo)
 	 */
-	public Compound getCompound();
+	public EntityInfo getCompound();
```
andreasprlic: should be getEntityInfo()

biojava-structure/src/main/java/org/biojava/nbio/structure/ChainImpl.java
```diff
@@ -217,7 +217,7 @@ private static int findMathingGroupIndex(List<Group> atomGroups, Group g) {
 	 *
 	 */
 	@Override
-	public void setCompound(Compound mol) {
+	public void setCompound(EntityInfo mol) {
```
andreasprlic: should be setEntityInfo

biojava-structure/src/main/java/org/biojava/nbio/structure/Structure.java
```diff
@@ -526,12 +526,12 @@ public Chain getChainByPDB(String chainId, int modelnr)
 	 *
 	 * @return a list of Compounds
 	 */
-	public List<Compound> getCompounds();
+	public List<EntityInfo> getEntityInformation();
```
andreasprlic: should be getEntityInfos()

abradle:
Yeah - good point Andreas. Thanks,

biojava-structure/src/main/java/org/biojava/nbio/structure/EntityInfo.java
```diff
@@ -80,21 +81,21 @@
 	private Map<String, Map<ResidueNumber,Integer>> chains2pdbResNums2ResSerials;
 
 	private String refChainId;
-
-	private String molName = null;
+	private String description = null;
 	private String title = null;
+	/**
+	 * The type of group (polymer, non-polymer, water)
```
josemduarte: Should be 'type of entity'

biojava-structure/src/main/java/org/biojava/nbio/structure/EntityInfo.java
```diff
@@ -988,4 +989,18 @@ public void addChain(Chain chain){
 	public void setChains(List<Chain> chains){
 		this.chains = chains;
 	}
+
+	/**
+	 * @return the type
```
josemduarte: Would be good to have javadocs for this one. This is an essential part of the refactoring and of the Entity concept.

Also should we have the return type as an Enum `EntityType` instead of a String?

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/SimpleMMcifConsumer.java
```diff
@@ -1125,7 +1126,7 @@ private void addCompounds(StructAsym asym) {
 			// we only add the compound if a polymeric one (to match what the PDB parser does)
 			if (e!=null) {
 				c.setDescription(e.getPdbx_description());
-				c.setType(e.getType());
+				c.setType(EntityType.entityTypeFromString(e.getType()));
```
josemduarte: This will produce a null pointer if entity string is not recognised

josemduarte:
A couple of other important things: 
- the entity type is not specified in PDB format. Could we at least try to guess an entity type for them?
- the `CompoundFinder` class should set entity types for its heuristically found entities. Additionally we'd need to refactor `CompoundFinder` to `EntityFinder`

PR #445, opened 2016-03-29
Entity features

abradle:
Merge Jose's fixes


josemduarte:
This should have been part of #441 but we forgot it then

Issue #438, opened 2016-03-18
JavaDoc linting (java 8)


abradle:
By default Java 8 is very strict with doclinting. It will fail unless the docs are appropriately documented. See:
http://blog.joda.org/2014/02/turning-off-doclint-in-jdk-8-javadoc.html

We have two options to get round this for builds. 
1) Alter the pom.xml to not be so strict
http://stackoverflow.com/questions/15886209/maven-is-not-working-in-java-8-when-javadoc-tags-are-incomplete

2) Convene to the standards...

My suggestion would be 1) (and it's what I'm doing for the alpha release). But I thought it should be discussed.

sbliven:
Both?

sbliven:
Maybe a third option would be to switch to a more powerful doc parser. The [pegdown-doclet](https://github.com/Abnaxos/pegdown-doclet) lets you use markdown and PlantUML, in addition to the standard html syntax. It even has IntelliJ support (although not Eclipse, afaik).

andreasprlic:
I like both. 

Can the alternative doc parser get hooked up with mvn release? 

A

josemduarte:
I'd vote against doing non-standard docs. As good and easy-to-use as the markdown thing can be, the fact that it is non-standard makes it difficult for other people (possibly used to normal javadocs elsewhere) to contribute javadocs.

Also there are tons of tools to deal with javadoc following the standard, if we move away from that, we won't be able to use those. A standard might be bad, but its power resides precisely in that it's standard and everyone has agreed to use it.

I'd say that realistically we'd need to keep the non-strict parsing mode as @abradle suggests, while trying to fix as many as possible of the broken docs.

heuermh:
I vote for 2) Convene to the standards...

josemduarte:
I'll close this one for now. We'll have to keep improving the javadocs going forward.

PR #446, opened 2016-04-01
Merging of MMTF functionality

abradle:
**MMTF Reading support**
MMTF is the new binary file format for macromolecule structure data. It is still under development. More information (also under development) can be found here: http://mmtf.rcsb.org/, here: https://github.com/rcsb/mmtf and here: https://github.com/rcsb/mmtf-java
1. Added the MMTF package (org.biojava.nbio.structure.io.mmtf). Includes a decoder from MMTF and a test. 
2. Added dependency to mmtf-api and mmtf-common (alpha3) from the rcsb/mmtf project.  (https://github.com/rcsb/mmtf-java)

**Alternative location behaviour has been updated**
The following commits:
82b23d3
9730025
cecf434
0ba8656
c1ea01e
  1. All alternate location groups (except with microheterogenity) contain one copy of each atom in  the group (not just the alt loc atoms).
  2. Fixed a bug where the main group for microheterogenous groups was containing all atoms (e.g. Tryptophan would contain the aromatic nitrogens of Histidine)

**Refactoring and cleanup**
1. More changes from Compound to EntityInfo 
Following on from these:
ed563e31da46f9af5e7ec5a40bdf1453da4ada1d
c84e6ec1ee28ec4138b663497036ff2a730bcee3
2. Doclint set to none  <additionalparam>-Xdoclint:none</additionalparam> at least for the time being:
https://github.com/biojava/biojava/issues/438
3. Refactoring newS -> newStructure and removed unnecesary Exceptions


biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/SimpleMMcifConsumer.java
```diff
@@ -920,6 +929,28 @@ public void documentEnd() {
...
+								if (altLocGroup.getPDBName().equals(group.getPDBName())) {
+									altLocGroup.addAtom(groupAtom);
+								}
+							}
+						}
+					}
+				}
+			}
+		}
+	}
```
josemduarte: This block is a duplicate of one above in PdbFileParser. We should extract it to some utility class.

josemduarte:
Thanks, it looks great! Let's keep this open for a while to see if there are any more comments.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/SimpleMMcifConsumer.java
abradle: 814d836
Deals with this

heuermh:
Could these commits be rebased and squashed?  A lot of the commits could be left out via cherry pick, e.g. all the alpha release pom changes.  As far as I know those were not official releases, right?  Although they do show up in Maven Central, so what do I know.  :)

abradle:
Hi Michael,

Yes they were official alpha-releases (as you say pushed to maven central). So I suppose it would be a shame to remove the tags at which the release were made.

Best wishes,

Anthony

josemduarte:
I'll proceed to merge this in later today. I guess there will be some issues still to be solved but we can do that on the master branch.

PR #450, opened 2016-04-06
Fix #449 SeqRes missing from substructures

### LINKED ISSUE #449: SeqRes missing from substructures
The seqres of a partial substructure is always empty. For instance,

```{java}
Structure s = StructureIO.getStructure("1hh0.A");
List<Group> seqres = s.getChain(0).getSeqResGroups();
assertEquals("Wrong seqres length",46,seqres.size());
```

fails with  `expected:<46> but was:<0>`.

The problem is that `SubstructureIdentifier.reduce()` doesn't handle the seqres properly.

sbliven:
Fix for #449 

Incidentally also merges in the bugfix branch.


sbliven:
I should have done the merge after the fix. Now I'll have to cherry-pick to the right branch.

Issue #457, opened 2016-04-21
Groups mapping to multiple residues in sequence, how to deal with them?


josemduarte:
Some groups in PDB structures can map to more than 1 residue in the sequence. The chemical component dictionary gives several letters for them in `_chem_comp.one_letter_code` instead of the usual single letter. This happens for instance in many chromophores appearing in GFP proteins.

Some examples:
- `CRO` residue in 1ema (chain A)
- `CR2` residue in 3dqj (chain A)
- `SUI` residue in 1oew (chain A)
- `NRQ` residue in 3cfh (chain A)

In my opinion this is badly modelled in the mmCIF files: they should be modelled as multiple groups not to break the 1-to-1 relationship between groups and residues in the sequence. 

Anyway for the moment this is an issue in biojava that can lead to inconsistencies when going from seqres groups to sequence and viceversa. At the moment `Chain.getSeqResSequence` will produce the sequence with multiple letters for these groups, thus resulting in a sequence with longer length than the size of `Chain.getSeqResGroups`.

andreasprlic:
I'd say this is a feature. BioJava is representing the data as it is encoded in the mmcif files and chemical component dictionary. 

https://github.com/biojava/biojava-tutorial/blob/master/structure/special.md

PR #458, opened 2016-04-22
Mmtf 0.1.0 update

abradle:
- Updated code to be compatible with the new beta release (0.1.0) of the mmtf-java (encoder/decoder)

https://github.com/rcsb/mmtf-java
http://mmtf.rcsb.org/javadoc/api-0.1.0/

- Introduced mmtf writer into biojava
- mmtfActions allows for simple reading and writing of mmtf data
- Added demo functions
- Added tests for the mmtf suite
- Added an example mmtf file

Biojava changes
- fixed a bug/feature where bonds were only made for the first model.
- if a single seqResGroup has a multi-letter single_letter_code (e.g. CRO of 1EMA) this is returned as X in the sequence from getSeqResSequence in ChainImpl. #457


biojava-structure/src/main/java/org/biojava/nbio/structure/ChainImpl.java
```diff
@@ -592,8 +592,10 @@ public String getSeqResSequence(){
 					PolymerType.POLYNUCLEOTIDE_ONLY.contains(cc.getPolymerType())){
 				// an amino acid residue.. use for alignment
 				String oneLetter= ChemCompGroupFactory.getOneLetterCode(cc);
-				if ( oneLetter == null || oneLetter.isEmpty() || oneLetter.equals("?"))
+				// AB oneLetter.length() should be one. e.g. in 1EMA it is 3 and this makes mapping residue to sequence impossible.
+				if ( oneLetter == null || oneLetter.isEmpty() || oneLetter.equals("?") || oneLetter.length()!=1) {
```
josemduarte: FYI this changes the behaviour discussed in #457, with this `getSeqResSequence` and `getSeqResGroups` have the same size

biojava-structure-gui/pom.xml
```diff
@@ -1,111 +1,113 @@
...
+			<url>http://www.gnu.org/licenses/old-licenses/lgpl-2.1.txt</url>
+			<distribution>repo</distribution>
+		</license>
+	</licenses>
+	<repositories>
+		<!-- Used for javaws:javaws -->
+		<repository>
+			<id>biojava-maven-repo</id>
+			<name>BioJava repository</name>
+			<url>https://github.com/biojava/maven-repo/raw/master/</url>
```
andreasprlic: I just noticed that structure-gui still has the link to the biojava specific Maven repo. Let's fix this. After this pull request has been accepted, we should replace this with 
```xml
<dependencies>
    <dependency>
      <groupId>javax.jnlp</groupId>
      <artifactId>jnlp-api</artifactId>
      <version>7.0</version>
      <scope>system</scope>
      <systemPath>${java.home}/lib/javaws.jar</systemPath>
    </dependency>
  </dependencies>
```

see also http://stackoverflow.com/questions/4645104/how-to-bundle-jnlp-api-with-maven-project

biojava-structure/src/main/java/org/biojava/nbio/structure/ChainImpl.java
andreasprlic: I don't think we should just change this in this way. We have documented this as a feature in the tutorial. Until the chemical component representation of chromophores has been fixed, let's add a utility method somewhere which masks groups with oneLetter length > 1 as X.

PR #465, opened 2016-04-27
pmd:ConsecutiveAppendsShouldReuse - Consecutive Appends Should Reuse

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule pmd:ConsecutiveAppendsShouldReuse - Consecutive Appends Should Reuse

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=pmd:ConsecutiveAppendsShouldReuse

Please let me know if you have any questions.

M-Ezzat

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/util/IOUtils.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/util/IOUtils.java
@@ -312,17 +312,16 @@ public class IOUtils {
 	 * @return legend String
 	 */
 	public static String getPDBLegend() {
-		StringBuilder s = new StringBuilder();
-		s.append("</pre></div>");
-		s.append("          <div class=\"subText\">");
-		s.append("          <b>Legend:</b>");
-		s.append("          <span class=\"m\">Green</span> - identical residues |");
-		s.append("          <span class=\"sm\">Pink</span> - similar residues | ");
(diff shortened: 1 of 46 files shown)


andreasprlic:
Does this code change really offer any performance difference? I'd say that the compiler would take care of doing things like that.

m-ezzat:
Based on the PMD plugin, it has a performance impact, and I see that it's also more clearer for readability.
Here's another reference http://pmd.sourceforge.net/pmd-5.2.0/pmd-java/rules/java/strings.html

andreasprlic:
Thanks. Looking at the PMD page, it claims that this 'can' improve performance, but I could see no further reference, where that information is coming from. Do we have any other sources that describe this? 

As I said above, I'd be surprised if the compiler does not deal with this at compile time.

andreasprlic:
This suggests that there is no difference:

http://programmers.stackexchange.com/questions/173377/java-performance-of-stringbuilder-append-chains

m-ezzat:
No, I actually don't have, and I've seen this resource before, but I'm not sure if this is really true or not, but don't think that it's a fake rule by PMD.

andreasprlic:
I just checked the PMD mailing list archives for a discussion of this topic but I could not find anything there. Perhaps the mailing search interface there is not very useful.

andreasprlic:
I don't think this rule adds much value. However let's review and apply the other rules as discussed in #484 .

PR #466, opened 2016-04-27
Fix #455. Use parsing params for AtomCache

### LINKED ISSUE #455: Default parsing params for AtomCache
I recently got tripped up by the fact that AtomCache's constructor disables alignSeqRes and parseSecStruc specifically.

```{java}
// we don't need this here
params.setAlignSeqRes(false);
// no secstruc either
params.setParseSecStruc(false);
```

The current default for FileParsingParams.alignSeqRes is true. I would have expected `cache.setFileParsingParams(new FileParsingParams())` to reset a cache to its default state, but this is not true.

Would anyone object if I deleted the above two lines, making `alignSeqRes=true` the default?

sbliven:
Specifically, set default alignSeqRes parameter in AtomCache  to true.


sbliven:
Looks like Genbank is down again for the tests.

PR #475, opened 2016-05-03
NCS operator field non-numeric value handling.

abradle:
Updated MMCif Parser to handle non-numeric data in the NCS operator field


josemduarte:
I'll cherry pick this to bugfixes too

Issue #469, opened 2016-04-28
Add better support for asym chain ids as well as auth ids


abradle:
Meeting notes from a discussion between @josemduarte @arose @andreasprlic @abradle : 

We can improve the support for working with both public (authId) and private (asymId) chain IDs assigned by PDB by adding several new methods to the Structure interface:

* `getChainByPdb(String authId);` (the current behavior)
* `getPolyChain(String asymId);`
* `getNonPolyChain(String asymId);`
* `getPolyChainByPdb(String authId);`
* `getNonPolyChainByPdb(String authId);`

Underlying we can store both asymId as well as authId in chain.name and chain.id;

`chain.getId()` will now respond with asymId; (note: currently getId() returns a Long, first take out from interface, then refactor chain.getInternalChainId() to getId() )
`chain.getName()` will now respond with authId; (note: currently called getChainID() , needs to be refactored to getName();

Internally the Structure object will contain a list of Chain objects, where polymeric and non-polymeric chains will be represented separately. Each chain will have a unique Id (asymId) but one polymeric and one non-polymeric chain can share the same name (=authId, =public PDB chain Id).

`Structure.getChainByPdb()` currently returns one chain. With this change, we will introduce a new `ChainCollection` class that implements `Chain` and contains a collection of polymeric and nonpolymeric ChainImpls that share the same authId (name).

`Structure.getChains()` will return a List of `ChainCollection`s.
`Structure.getPolyChains()` will return all polymeric chains.
`Structure.getNonPolyChains()` will return all non-polymeric (ligand) chains.

josemduarte:
A note: the model above __is essentially the mmCIF data model__. When parsing from PDB files (which don't naturally adhere to this data model) we would have to find the different parts of the chains that are polymeric/non-polymeric to be able to create the data structure as above (including creating some fake asym_ids).

andreasprlic:
I have an initial version of this working in my branch. However there are now still a lot of problems with unit tests that need to be addressed.

andreasprlic:
I have fixed many unit tests that got broken because of this (mostly minor data changes due to ligands now being in their own chains). Currently 27 / 297 unit tests are failing. @abradle  can you take a look a the test related to bonds? @josemduarte can you take a look at the ASA test?

andreasprlic:
fixed with #479

PR #479, opened 2016-05-11
Refactoring of structure data model

josemduarte:
This implements what we discussed in issue #469, with commits from @andreasprlic, @abradle and myself.

Some highlights of the changes:
- new data model for the structural data: chains are split into polymeric, non-polymeric and water chains, following the mmCIF data model. Chains can be accessed using both asymIds and author chain ids.
- rewrite of the PDB parser so that the data parsed adheres to the new mmCIF data model now used internally in the biojava data structures. Chains are split into their polymeric and several non-polymeric parts and asym ids are assigned to each of them following the same rules used in mmCIF.
- entities are now created heuristically for all compounds in the files even if entity annotation is not present in PDB or mmCIF files
- discontinued support for: parsing of old PDB files with blank chain ids, parsing of LINK records in PDB files (we do have CONECT records still)
- it is important to note that all downstream code will need to adapt to this whenever operations with chains are performed. Especially calls to `Structure.getChainByPDB` (now deprecated) need to change to `Structure.getPolyChainByPDB` or `Structure.getNonPolyChainByPDB`

Tests are passing but I'm sure many issues still remain, let's try to iron them out slowly as we move forward.


andreasprlic:
Thanks Jose! Just to add, `Structure.getChainByPDB` should not get removed until we are in the BioJava 6 series (at least)...

andreasprlic:
need to change the signature of all new get*ChainByPDB methods. Response should be a List<Chain> rather than a Chain object.

josemduarte:
I've reverted `List<Chain> getPolyChainsByPDB(authId)` to `getPolyChainByPDB(authId)`, but kept `List<Chain> getNonPolyChainsByPDB(authId)`. The one-to-many relationship only exists for non-polymer chains.

Other than that I've fixed a few more issues with PDB parsing and I think this is ready to be merged.

andreasprlic:
There don't seem to be any major comments on this. I think this one is an important step forward and I will merge this in shortly.

PR #481, opened 2016-05-13
Automatically re-download ECOD if the cache is >2weeks old

sbliven:
Since ECOD releases every two weeks, we shouldn't cache version "latest" longer than that.
With this the cache is at most one version behind.


josemduarte:
Let's continue with cherry-picking to bugfixes-4.2. I think this one should be there

PR #482, opened 2016-05-13
Fix #480: Allow CIF labels to differ from the java field name

### LINKED ISSUE #480: Misspelled spacegroup field in generated CIFs
CIF files should contain `_symmetry.space_group_name_H-M`, but the hyphen is getting converted into an underscore. This may occur for other mmCIF fields containing hyphens as well:

```
$ sed -En 's/^\s*(save_)?(_[^ ]+-[^ ]+)/\2/p' mmcif_pdbx_v40.dic|sort -u
_diffrn_source.take-off_angle
_em_2d_crystal_entity.space_group_name_H-M
_pdbx_nmr_constraints.NA_alpha-angle_constraints_total_count
_pdbx_nmr_constraints.NA_beta-angle_constraints_total_count
_pdbx_nmr_constraints.NA_chi-angle_constraints_total_count
_pdbx_nmr_constraints.NA_delta-angle_constraints_total_count
_pdbx_nmr_constraints.NA_epsilon-angle_constraints_total_count
_pdbx_nmr_constraints.NA_gamma-angle_constraints_total_count
_pdbx_nmr_constraints.NA_other-angle_constraints_total_count
_pdbx_nmr_force_constants.non-bonded_inter_conf_db_potential_term
_pdbx_nmr_force_constants.non-bonded_inter_radius_of_gyration_term
_pdbx_nmr_force_constants.non-bonded_inter_radius_of_gyration_term_units
_pdbx_nmr_force_constants.non-bonded_inter_van_der_Waals_term
_pdbx_nmr_force_constants.non-bonded_inter_van_der_Waals_term_type
_pdbx_nmr_force_constants.non-bonded_inter_van_der_Waals_term_units
_pdbx_point_symmetry.H-M_notation
_space_group.name_H-M_alt
_symmetry.pdbx_full_space_group_name_H-M
_symmetry.space_group_name_H-M
```

See also https://github.com/arose/ngl/pull/111

sbliven:
This provides a general solution to the problem in #480, which is that some CIF labels contain characters not permitted in java identifiers (like hyphens). Thus, a mechanism is needed for mapping java fields to the correct identifier for cases like `_symmetry.space_group_name_H-M` when writing to mmCIF.

This is implemented by adding two annotations. `@CIFLabel(label="str")` causes the cif property to be `_category.str`. `@IgnoreField` causes the field to be ignored entirely.

This pull also changes the MMCIFFileTools.toMMCIF method to be generic,
so that we enforce at compile time that all loop rows must be of the same type. The old method is deprecated (and should be backwards compatible if I understand type erasure correctly).

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/contact/StructureInterface.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/contact/StructureInterface.java
@@ -20,7 +20,20 @@
  */
 package org.biojava.nbio.structure.contact;
 
-import org.biojava.nbio.structure.*;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+import org.biojava.nbio.structure.Atom;
+import org.biojava.nbio.structure.Chain;
+import org.biojava.nbio.structure.Element;
+import org.biojava.nbio.structure.EntityInfo;
+import org.biojava.nbio.structure.Group;
+import org.biojava.nbio.structure.GroupType;
+import org.biojava.nbio.structure.ResidueNumber;
+import org.biojava.nbio.structure.Structure;
 import org.biojava.nbio.structure.asa.AsaCalculator;
 import org.biojava.nbio.structure.asa.GroupAsa;
 import org.biojava.nbio.structure.io.FileConvert;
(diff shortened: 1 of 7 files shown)


josemduarte:
This is awesome! Couldn't we use these annotations also while parsing the files? (not only when writing them). 

For instance there's a few fields that contain square brackets, e.g. `_pdbx_struct_oper_list.matrix[1][1]`. Those are handled in `SimpleMMcifParser` in special ways (instead of with the generic `buildObject` method), but it should be possible to handle them like the others via using these annotations, shouldn't it?

josemduarte:
I'm working in using the annotations also from the cif parser. I'll pull request some commits to your branch when I'm ready.

josemduarte:
I'll merge this in. I'll then add a new pull request to deal with reading mmcif files using the new tags.

PR #493, opened 2016-05-25
Prepare Biojava to use MMTF as default

abradle:
**mmCIF is still the default** - but MMTF could now be used as default if we wanted. (All tests pass if I put MMTF as default).

- Altered mmCIF specific tests to be forced to use mmCIF
- Fixed bugs found in MMTF 
- Parse more data from mmCIF files (e.g. R-work)
- Secondary structure is calculated for all models
- Deuterium bonds are now created - if the atom name for the corresponding H is found in the CCD (e.g. D1 vs H1).
- Added some more bond testing functionality


andreasprlic:
Great stuff!

PR #497, opened 2016-06-03
Version 0.2.0 of MMTF support

abradle:
1) **Sets default atomCache to MMTF**
2)  Updates the Reader and Writer to work with 0.2.0 MMTF
3) Added tests for the new MMTFUtils class
4) Refactoring of docs and code
5) Fixed bugs in the generation of SeqResGroups


andreasprlic:
nice, however unfortunately does not improve the speed of the unit tests. Still too much mmcif parsing, I suppose.

PR #513, opened 2016-06-15
MMTF roundtripping in Biojava

abradle:
1. Fix for the MmtfStructureReader to handle altLoc being a null byte.
2. Ensure groupType is upper case
3. SeqResGroups get a working chemComp
4. Added test that Biojava can roundtrip structures

### DIFF
--- a/biojava-structure/src/test/java/org/biojava/nbio/structure/io/mmtf/TestMmtfRoundTrip.java
+++ b/biojava-structure/src/test/java/org/biojava/nbio/structure/io/mmtf/TestMmtfRoundTrip.java
@@ -1,7 +1,19 @@
 package org.biojava.nbio.structure.io.mmtf;
 
-import java.io.IOException;
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+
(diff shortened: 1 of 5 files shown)


biojava-structure/src/test/java/org/biojava/nbio/structure/io/mmtf/TestMmtfRoundTrip.java
```diff
@@ -0,0 +1,33 @@
...
+	 * @throws StructureException an error parsing the structure
+	 */
+	@Test
+	public void testRoundTrip() throws IOException, StructureException {
+		Structure structure = StructureIO.getStructure("4CUP");
+		AdapterToStructureData writerToEncoder = new AdapterToStructureData();
+		new MmtfStructureWriter(structure, writerToEncoder);
+		MmtfStructureReader mmtfStructureReader = new MmtfStructureReader();
+		new StructureDataToAdapter(writerToEncoder, mmtfStructureReader);
+		mmtfStructureReader.getStructure();
```
josemduarte: It would be good to assert a few things about the roundtripped structure so that we actually check we are getting back the same thing. Or are we testing that elsewhere?

abradle: Yeah I test that elsewhere. Feel free to add it in. This is just a simple test to check we can round trip.

sbliven:
_She was a round tripper / a one way ticket no more_
_It took me so long to find out / and I found out_

PR #514, opened 2016-06-16
squid:S1206 - "equals(Object obj)" and "hashCode()" should be overrid…

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule squid:S1206 - "equals(Object obj)" and "hashCode()" should be overridden in pairs

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:S1206

Please let me know if you have any questions.

M-Ezzat

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/secstruc/SecStrucCalc.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/secstruc/SecStrucCalc.java
@@ -29,6 +29,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


biojava-structure/src/main/java/org/biojava/nbio/structure/secstruc/SecStrucCalc.java
```diff
@@ -679,6 +680,14 @@ public String printFASTA() {
...
 
 	@Override
+	public int hashCode() {
+	    final int prime = 31;
+	    int result = 1;
+	    result = prime * result + Arrays.hashCode(atoms);
+	    return result;
+	}
+
+	@Override
```
josemduarte: This is not acting on the same fields as the `equals` below (`equals` checks for sec structure elements to be the same).

The comment is a bit late, now that it's been merged. But we should try to solve this...

biojava-structure/src/main/java/org/biojava/nbio/structure/symmetry/core/PermutationGroup.java
```diff
@@ -137,6 +137,21 @@ public String getGroupTable() {
 
 	@Override
 	public int hashCode() {
-		return getGroupTable().hashCode();
+	    return getGroupTable().hashCode();
```
josemduarte: Again this is not on same fields as the `equals`

PR #518, opened 2016-06-20
Fixing issues in squid:S1206 - "equals(Object obj)" and "hashCode()" should be …

m-ezzat:
Referenced by #514

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/secstruc/SecStrucCalc.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/secstruc/SecStrucCalc.java
@@ -683,7 +683,7 @@ public class SecStrucCalc {
 	public int hashCode() {
 	    final int prime = 31;
 	    int result = 1;
-	    result = prime * result + Arrays.hashCode(atoms);
+	    result = prime * result + Arrays.hashCode(groups);
 	    return result;
 	}
 
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/symmetry/core/PermutationGroup.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/symmetry/core/PermutationGroup.java
@@ -142,16 +142,16 @@ public class PermutationGroup {
 
 	@Override
 	public boolean equals(Object obj) {
-	    if (this == obj) {
-		return true;
-	    }
-	    if (obj == null) {
-		return false;
-	    }
-	    if (this.getClass() == obj.getClass()) {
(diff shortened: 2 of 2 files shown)


biojava-structure/src/main/java/org/biojava/nbio/structure/secstruc/SecStrucCalc.java
```diff
@@ -683,7 +683,7 @@ public String printFASTA() {
 	public int hashCode() {
 	    final int prime = 31;
 	    int result = 1;
-	    result = prime * result + Arrays.hashCode(atoms);
+	    result = prime * result + Arrays.hashCode(groups);
```
josemduarte: Again this is not the same as what we have in the `equals`

m-ezzat: I see both of them deal with getGroupTable() methods
What do you recommend it to be ?

josemduarte: The `equals` compares `SeqStrucInfo` objects are equals per `Group`, whilst above the `hashCode` implementation is based on equality of groups. Instead one would need to build an array of all `SecStrucInfo`s and then compute their hashCode (making sure that `SeqStrucInfo` does implement a hashCode too).

biojava-structure/src/main/java/org/biojava/nbio/structure/symmetry/core/PermutationGroup.java
```diff
@@ -142,16 +142,16 @@ public int hashCode() {
...
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (this.getClass() == obj.getClass()) {
+            return getGroupTable().equals(((PermutationGroup) obj).getGroupTable());
+        }
+        return false;
+    }
```
josemduarte: I'm not familiar with this class, but it would be good if someone (@sbliven, @lafita, @pwrose  ?)  could check this is alright.

sbliven: I think it's correct but unnecessary. `getGroupTable()` depends only on the `permutations` member and fully represents all information in it. I guess you could argue that using the `getGroupTable()` function in both `hashCode` and `equals` makes their correctness more obvious, but I'm not sure it's worth the additional computation to convert `permutations` into a string.

andreasprlic:
results are not equivalent

PR #520, opened 2016-06-24
Major Refactoring of the Quaternary Symmetry Detection

lafita:
Description
--------------
The refactoring is motivated by a limitation of the previous implementation: it was using `Chains` as the **units** to calculate the symmetry. 
Some `Chains` can have internal domains, some participating in the symmetry and others not. Some `Chains` can also have internal symmetry (domains or repeats) that expand the quaternary symmetry point group. Finally, `Chain` definition is dependent on the file format (author, PDB), and there might be cases where the `Chain` boundaries or definition has to be changed, without changing the entire `Structure` object.
To overcome the limitation a **new unit** of symmetry, the `Subunit`, which is just defined as an array of `Atoms`, has been introduced. This is a more general approach, since `Chains` are a special case of `Subunit`.

In addition, the previous implementation was using a **clustering** algorithm to cluster `Chains` into groups of identical (same entity) or similar (sequence and/or structural) `Chains` to calculate the equivalent unit groups for determining the stoichiometry, calculating symmetry alignments and coloring the `Chains`. 
The clustering algorithm was entangled with the **symmetry detection** algorithm. They shared the same parameter bean, they were located in the same package and they used internal objects to handle the information.
The organization of the code has been changed to split the **clustering** from the **symmetry detection** algorithm, making their code completely independent. This allows the functionality of the **clustering** algorithm to be accessible for other purposes other than detecting the symmetry (for example, to match the `Chains` from two different `Structures`).

Changes
-----------
The major changes introduced are: 

1. Move the clustering algorithm and classes to **structure.cluster** package.
2. Implement equivalent classes of `Chain` clustering for `Subunit` clustering.
3. Split the `QuatSymmParameters` into `SubunitClustererParameters` and `QuatSymmParameters`.
4. Make `SubunitClusterer` and `QuatSymmetryDetector` static classes.
5. Create tests for all `Subunit` clustering methods and expand test cases for `QuatSymmetryDetector`.
6. Use **ENUMS** to designate clustering and symmetry detection methods.
7. Unify the code with the rest of **BioJava**. Use `StructureTools` and `Calc` where possible.
8. Document extensively methods and classes.
9. Use the **logger** instead of the **verbose** parameter.
10. Use **CE-Symm** to divide `Subunits` in internal symmetry domains.
11. Move `ScanSymmetry` analysis script to the **symmetry** project.
12. Move **bioassembly** handling to the highest level: the user should take care to download a biological assembly if he/she wants to analyze the assembly instead of the AU. No special cases in the code should exist for biological assemblies.

A schematic figure of the previous (left) and new (right) code structure is below:

sofla_112_216062414400_0001

Remarks
-----------
We decided to keep the pull request open for some weeks, so that the code can be reviewed and tested, specially by @pwrose. 

In the **structure.cluster** package you will find the old and the new classes for clustering. The old classes have been deprecated and are no longer used, but I kept them to be able to compare the equivalent methods (like `SubunitExtractor` vs `ProteinChainExtractor`, or `SubunitClusterer` vs `ChainClusterer`). Before merging the pull request, the old classes have to be removed.


lafita:
There is a problem with **MMTF** file downloads, they time out. Travis is failing because of that, and the tests do not pass locally either.

abradle:
Yes there is a problem with our cloud. I have e-mailed Cole about it (this
morning UK time) - awaiting response (probs not till tomorrow now)

andreasprlic:
There are some hardware problems in the server room, actively being worked on. There is no east-coast mirror as of yet, but coming soon.

PR #534, opened 2016-07-13
Adding tools to map from chromosome positions to mRNA coding sequence…

andreasprlic:
… position.

The genome module now also depends on the Guava library. This suggests that we should probably merge the Genome and Sequencing modules, since there is less and less difference between the two.


sbliven:
Can this be merged, or does it depend on #536?

andreasprlic:
does not depend on #536

biojava-genome/src/main/java/org/biojava/nbio/genome/util/ChromosomeMappingTools.java
```diff
@@ -0,0 +1,1394 @@
...
+                        System.out.println(codingLength + " | " + tmp);
+                        System.out.println(" -> found position in coding exon #" + (i + 1) + ":  " + format(chromPos) + " " + format(start + 1) + " " + format(tmp) + " " + cdsStart + " " + codingLength);
+                    }
+
+                    return codingLength + tmp ;
+                }
+                // full exon is coding
+                codingLength += (end - start);
+                if (debug)
+                    System.out.println("     Exon        :" + format(start) + " - " + format(end) + " | " + format(end - start) + " | " + codingLength + " | " + (codingLength % 3));
```
josemduarte: All these should be done with `logger.debug` and not use a `debug` constant at all.

biojava-genome/src/main/java/org/biojava/nbio/genome/util/ChromosomeMappingTools.java
```diff
@@ -0,0 +1,1394 @@
+package org.biojava.nbio.genome.util;
+
+
+import com.google.common.collect.Range;
+
+import org.biojava.nbio.core.sequence.ProteinSequence;
```
josemduarte: This import is not used

biojava-genome/src/main/java/org/biojava/nbio/genome/util/ChromosomeMappingTools.java
```diff
@@ -0,0 +1,1394 @@
...
+
+    public static String formatExonStructure(GeneChromosomePosition chromosomePosition ){
+        if ( chromosomePosition.getOrientation() == '+')
+            return formatExonStructureForward(chromosomePosition);
+
+        return formatExonStructureReverse(chromosomePosition);
+
+    }
+
+    public static String printHTMLExonStructure(GeneChromosomePosition chromosomePosition ){
```
josemduarte: Any documentation for this? What kind of html output does this produce? Perhaps we shouldn't have this specific output thing here in biojava but elsewhere

biojava-genome/src/main/java/org/biojava/nbio/genome/util/ChromosomeMappingTools.java
```diff
@@ -0,0 +1,1394 @@
...
+public class ChromosomeMappingTools {
+
+    public static final boolean debug = false;
+    private static final String newline = System.getProperty("line.separator");
+
+
+    public static final String CHROMOSOME = "CHROMOSOME";
+    public static final String CDS = "CDS";
+
+    public static String formatExonStructure(GeneChromosomePosition chromosomePosition ){
```
josemduarte: Any docs for this? What kind of formatting does this produce?

biojava-genome/src/main/java/org/biojava/nbio/genome/util/ChromosomeMappingTools.java
```diff
@@ -0,0 +1,1394 @@
...
+    }
+
+
+
+    /** Extracts the exon boundaries in CDS coordinates. (needs to be divided by 3 to get AA positions)
+     *
+     * @param chromPos
+     * @return
+     */
+    public static List<Range> getCDSExonRanges(GeneChromosomePosition chromPos){
```
josemduarte: Range should be parameterised, the compiler gives warnings for it. Same for all other occurrences below

andreasprlic: Interesting! IntelliJ did not show this warning. Will fix...

andreasprlic:
this addresses all of @josemduarte 's comments.

josemduarte:
Thanks!

Issue #517, opened 2016-06-17
Quaternary Symmetry Detection Broken with MMTF


lafita:
Using structures downloaded with the `MMTF` format break the `QuatSymmetryDetector` algorithm. It is related to the chemical component information. The tests explicitly set the format to `MMCIF`, so that is why the bug was not noticed when changing `MMTF` to default. 

Here the stacktrace when the line `cache.setUseMmCif(true);` in `TestQuatSymmetryDetection` or `DemoOrientBioAssembly` is commented out (`MMTF` used):

<pre>
java.lang.NullPointerException
	at org.biojava.nbio.structure.io.mmcif.ChemCompGroupFactory.getChemComp(ChemCompGroupFactory.java:47)
	at org.biojava.nbio.structure.HetatomImpl.getChemComp(HetatomImpl.java:429)
	at org.biojava.nbio.structure.ChainImpl.getSeqResSequence(ChainImpl.java:576)
	at org.biojava.nbio.structure.symmetry.core.ProteinChainExtractor.extractProteinChains(ProteinChainExtractor.java:121)
	at org.biojava.nbio.structure.symmetry.core.ProteinChainExtractor.run(ProteinChainExtractor.java:95)
	at org.biojava.nbio.structure.symmetry.core.ProteinChainExtractor.getCalphaTraces(ProteinChainExtractor.java:60)
	at org.biojava.nbio.structure.symmetry.core.ProteinSequenceClusterer.extractProteinChains(ProteinSequenceClusterer.java:98)
	at org.biojava.nbio.structure.symmetry.core.ProteinSequenceClusterer.run(ProteinSequenceClusterer.java:88)
	at org.biojava.nbio.structure.symmetry.core.ProteinSequenceClusterer.getSequenceAlignmentClusters(ProteinSequenceClusterer.java:56)
	at org.biojava.nbio.structure.symmetry.core.ClusterProteinChains.run(ClusterProteinChains.java:84)
	at org.biojava.nbio.structure.symmetry.core.ClusterProteinChains.<init>(ClusterProteinChains.java:42)
	at org.biojava.nbio.structure.symmetry.core.QuatSymmetryDetector.run(QuatSymmetryDetector.java:104)
	at org.biojava.nbio.structure.symmetry.core.QuatSymmetryDetector.hasProteinSubunits(QuatSymmetryDetector.java:74)
[...]

PR #538, opened 2016-07-15
Fix issue 517 - debug mmtf reader

abradle:
Remove the useMmcif from test
Fix the bug and clean up the logic of adding seq res groups
https://github.com/biojava/biojava/issues/517


abradle:
This is down http://release.cathdb.info/v4.0.0/CathDomainList - so the pull request's won't work.

josemduarte:
I've tried locally and it all works... That's funny, maybe some tests run in travis that don't run locally? Anyway all that is unrelated to this PR. I'll merge it, thanks @abradle !

PR #540, opened 2016-07-27
Fix #537 - handle missing and empty Element column in PDB files

### LINKED ISSUE #537: PDB File Models Missing Element Column
We have recently noticed that some structural bioinformatics programs (structure refinement or modelling) generate **PDB files** where the `Element` column is missing. The element column is the last column, where the periodic table element of the `Atom` is indicated.

Parsing these files with **BioJava** currently does not allow the calculation of structural alignments or symmetry (and any other analysis using C-alpha atoms), because to extract the C-alpha atoms of a structure the name (**CA**) and element (**C**) of the `Atoms` is checked (in `StructureTools.getRepresentativeAtoms()`).

The `Element` column is not completely redundant, because in case of a modified aminoacid with calcium bound to it, the name **CA** alone does not distinguish the **calcium** from the **C-alpha** carbon and the element column is needed to do so.

On the other hand, we could print a **warning** when parsing such models and guess and fill the `Element` column from the `Atom` names (at least for the `Atoms` in aminoacids), in order to support the incomplete files.

**Question**: _is there any drawback in guessing and filling the `Element` of the `Atoms`? Can we use the `Chemical Components` for that?_

lafita:
The element information is filled using the **Chemical Component Dictionary** information, as discussed in #537.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/io/PDBFileParser.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/io/PDBFileParser.java
@@ -70,6 +70,7 @@ import org.biojava.nbio.structure.StructureException;
 import org.biojava.nbio.structure.StructureImpl;
 import org.biojava.nbio.structure.StructureTools;
 import org.biojava.nbio.structure.io.mmcif.ChemCompGroupFactory;
+import org.biojava.nbio.structure.io.mmcif.model.ChemCompAtom;
 import org.biojava.nbio.structure.secstruc.SecStrucInfo;
 import org.biojava.nbio.structure.secstruc.SecStrucType;
 import org.biojava.nbio.structure.xtal.CrystalCell;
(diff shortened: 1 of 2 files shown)


biojava-structure/src/main/java/org/biojava/nbio/structure/io/PDBFileParser.java
```diff
@@ -1825,30 +1826,48 @@ private void  pdb_ATOM_Handler(String line)	{
...
+		// element from the chemical component.
 		Element element = Element.R;
+		boolean guessElement = true;
 		if ( line.length() > 77 ) {
 			// parse element from element field
-			try {
-				element = Element.valueOfIgnoreCase(line.substring (76, 78).trim());
-			}  catch (IllegalArgumentException e){}
+			String elementSymbol = line.substring(76, 78).trim();
+			if (elementSymbol.equals("")) {
```
josemduarte: Minor comment: better use isEmpty() here

biojava-structure/src/main/java/org/biojava/nbio/structure/io/PDBFileParser.java
```diff
@@ -1825,30 +1826,48 @@ private void  pdb_ATOM_Handler(String line)	{
...
+				logger.warn("Element column was empty. Assigning atom element "
+						+ "from Chemical Component Dictionary information", elementSymbol);
```
josemduarte: The `elementSymbol` parameter is superfluous here. In any case `elementSymbol` is empty so there's not point in printing it out. It would be nice if some more info like atom name and serial could be output in the warning.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/PDBFileParser.java
```diff
@@ -1825,30 +1826,48 @@ private void  pdb_ATOM_Handler(String line)	{
...
+			} else {
+				try {
+					element = Element.valueOfIgnoreCase(elementSymbol);
+					guessElement = false;
+				}  catch (IllegalArgumentException e){
+					logger.warn("Element {} was not recognised. Assigning atom element "
+							+ "from Chemical Component Dictionary information", elementSymbol);
```
josemduarte: Same here, it would be good to have atom name and serial (and even residue number)

biojava-structure/src/main/java/org/biojava/nbio/structure/io/PDBFileParser.java
```diff
@@ -1825,30 +1826,48 @@ private void  pdb_ATOM_Handler(String line)	{
...
+					logger.warn("Atom name {} was not found in the Chemical Component Dictionary information of {}. "
+							+ "Assigning generic element R to it", fullname.trim(), currentGroup.getPDBName());
+				} else {
 					element = Element.valueOfIgnoreCase(elementSymbol);
-			} catch (IllegalArgumentException e){
-				logger.warn("Element {} was not recognised. Assigning generic element R to it", elementSymbol);
+				}			
+			} else {
+				logger.warn("Chemical Component Dictionary information was not found for Atom name {}. "
+						+ "Assigning generic element R to it", fullname.trim());
```
josemduarte: Same comment for this warning as for the others

josemduarte:
Nice stuff! I think we could try to cherry pick this one into the bugfix branch. I think it shouldn't conflict too much despite the parser was changed a lot in 5.0. I'd say let's try cherry picking and if it isn't easy to merge then we can drop it.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/PDBFileParser.java
lafita: Yes, you are right! I forgot to remove it. Yes this is possible to output, I will include it.

lafita:
I made the changes

josemduarte:
Thanks!

josemduarte:
Alright I've tried cherry picking back to bugfixes-4.2 but it was complicated. Let's just keep this for the master branch

PR #541, opened 2016-07-27
Fix for issue 517

josemduarte:
As mentioned in the discussion in #517 the issue was with seqres groups not having set the chem component. The mmtf inflator was actually setting them correctly (without a 3-letter code though, a known limitation), but the problem was in cloning of groups (needed to create the bioassembly) where the chem comp wasn't cloned properly.


josemduarte:
So the tests have been failing a few times and still are. I've solved some of the issues related to the changes introduced in #537 and pushed the commits. However the CATH server is not working at the moment and the build still fails because of that. 

In any case this solves issue #517, so the related tests that were failing for the last week or so are now fine with this.

I've also back-ported the cloning fix to bugfixes-4.2 (see 46b30b35f409f2b5b705190f523cf47542537a91)

lafita:
Thank you @josemduarte! I checked out your branch and the tests work fine. About the CATH problem I will create a new issue.

andreasprlic:
merging although unit tests fail (related to CATH, so not related to this pull request(

Issue #542, opened 2016-07-27
Remove SCOP related unit tests


andreasprlic:
SCOP has been shut down at LMB. As such any SCOP related unit tests will not work going forward. We need to take them out.

andreasprlic:
Actually it is easier. We just need to change the default SCOP download location to point to Berekeley-SCOP and then the tests are passing again.

PR #543, opened 2016-07-27
#542 changing default SCOP download location to Berkeley- SCOP.

andreasprlic:


josemduarte:
Tests are failing still because of #517, but that's unrelated. I'll merge

PR #546, opened 2016-07-28
Fix for issue 544

josemduarte:
This should fix the test problems for the moment. I didn't get a reply from cath yet, so don't know if the old URL is gone for good. In any case the new server layout seems to be well thought and solid, it looks like it is there to stay. One problem is I couldn't find the new file names for CathDomainDescriptionFile and CathNames, so that's a TODO for later.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathFactory.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathFactory.java
@@ -37,7 +37,8 @@ public class CathFactory {
 
 	public static final String VERSION_3_5_0 = "3.5.0";
 	public static final String VERSION_4_0_0 = "4.0.0";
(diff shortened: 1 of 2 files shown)


andreasprlic:
thanks!

lafita:
Thank you @josemduarte!

**CathDomainDescriptionFile**: http://download.cathdb.info/cath/releases/latest-release/v4_1_0/cath-classification-data/cath-domain-description-file-v4_1_0.txt
**CathNames**: http://download.cathdb.info/cath/releases/latest-release/v4_1_0/cath-classification-data/cath-names-v4_1_0.txt

I will make a direct commit to master with the names.

Should we include these changes to 4.2-bugfixes branch too?

andreasprlic:
Yes, I think that would be useful

josemduarte:
Thanks @lafita , I've just cherry-picked it back to bugfixes-4.2. There should be a release with all this later today.

PR #549, opened 2016-08-03
Fix a bug reading in Deuterated files using MMTF

abradle:
Minor bugfix - for reading Deuterated 'nanoheterogenity' cases using MMTF


biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmtf/MmtfStructureReader.java
```diff
@@ -282,9 +282,16 @@ public void setAtomInfo(String atomName,
...
 			// If it's not a microheterogenity case
 			if (group.getPDBName().equals(atom.getGroup().getPDBName())) {
-				group.addAtom(atom);
+				// And it's not a deuterated case.
+				if(StructureTools.hasNonDeuteratedEquiv(atom,group)){
+					
+				}
+				else{
+					group.addAtom(atom);
+				}
```
josemduarte: Could you inverse the logic here? It's not good practice to leave empty `if` blocks... If I read that in 2 years time I would think someone forgot to add some code in the if block.

josemduarte:
Thanks!

Issue #430, opened 2016-03-09
Bug in display of alignment identity and similarity


andreasprlic:
Reported by Art Nasamran (email):

 If you try DemoSW3DAligner.java from biojava-structure-gui and align 1CDG.A and 1TIM.A, the console says:
"Align 1CDG.A.pdb 686 with 1TIM.A.pdb 247

Twists 0 ini-len 0 ini-rmsd 0.00 opt-equ 56 opt-rmsd 13.21 chain-rmsd 13.21 Score 39.00 align-len 71 gaps 15 (21.13%)

Afp-num 0 Identity 37.50% Similarity 58.93%"

while View -> View Text Only from the text window shows: 

"Align 1CDG.A.pdb Length1: 686 with 1TIM.A.pdb Length2: 247

Equ: 56 
RMSD: 13.21 
Score: 39.00 
Align-len: 71 
Gaps: 15 (21.13%)
TM-score: 0.00

Identity: 29.58% 
Similarity: 46.48%"

Why are the identity and similarity percentages different?

lafita:
_Identity_ is the percentage of aligned residues that are **equal** (identity scoring matrix) and _Similarity_ is the percentage of aligned residues that are similar, where similarity is assessed by a scoring matrix (and a similarity threshold is set by default).

Similarity will always be higher than identity.

emckee2006:
It sounds like the two views disagree about the actual values, though.
Each reports similarity and identity, but the views disagree  (e.g.
identity in view 1 is different from identity in view 2).

lafita:
Sorry I misunderstood the issue. There is definitely a bug (or the calculation method is different) between the **Alignment Panel** and the **Text Only** views. 

I remember that some display options recalculated the superposition and alignment from the AFPChain.

altayhunoglu:
As far as I understand, while finding percentage of similarity and identity; code on the console side uses this: 
 	       

>  afpChain.setIdentity(nrIdent*1.0/pos);
> 		afpChain.setSimilarity(nrSim*1.0/pos);

 In this case pos= 56 (aligned positions)

 But in the code of **"View -> View Text Only**" the algorithm uses 71 (Align-len) instead of 56 (aligned positions)

josemduarte:
This was fixed, right @andreasprlic ?

andreasprlic:
based on the history of this ticket, yes!

PR #551, opened 2016-08-05
fix for #430

andreasprlic:


biojava-structure-gui/src/test/java/org/biojava/nbio/structure/gui/TestSimilarityCalc.java
```diff
@@ -0,0 +1,98 @@
+package org.biojava.nbio.structure.gui;
```
josemduarte: I think this test should go to the integrationtest module. It doesn't seem to be specifically related to structure-gui.

biojava-structure-gui/src/test/java/org/biojava/nbio/structure/gui/TestSimilarityCalc.java
```diff
@@ -0,0 +1,98 @@
...
+import org.biojava.nbio.structure.align.model.AfpChainWriter;
+import org.biojava.nbio.structure.align.seq.SmithWaterman3DParameters;
+import org.biojava.nbio.structure.align.seq.SmithWaterman3Daligner;
+import org.biojava.nbio.structure.align.util.AtomCache;
+import org.junit.Test;
+
+/**
+ * Created by andreas on 8/3/16.
+ */
+public class TestSimilarityCalc extends TestCase{
```
josemduarte: No need to extend TestCase, the @Test annotation in the individual methods is enough

biojava-structure-gui/src/test/java/org/biojava/nbio/structure/gui/TestSimilarityCalc.java
```diff
@@ -0,0 +1,98 @@
...
+            AFPChain afpChain = algorithm.align(ca1, ca2, params);
+
+            afpChain.setName1(name1);
+            afpChain.setName2(name2);
+
+
+
+            // get the scores
+            int ca1Length = afpChain.getCa1Length();
+            int ca2Length = afpChain.getCa2Length();
```
josemduarte: These 2 plus a bunch of other variables are not used, let's remove them

biojava-structure-gui/src/test/java/org/biojava/nbio/structure/gui/TestSimilarityCalc.java
```diff
@@ -0,0 +1,98 @@
...
+
+
+
+
+
+
+
+        } catch (Exception e){
+
+            fail(e.getMessage());
```
josemduarte: Better to throw the exception so that it's reported as an error instead of a failure

biojava-structure/src/main/java/org/biojava/nbio/structure/HetatomImpl.java
```diff
@@ -327,6 +329,71 @@ public boolean hasAminoAtoms(){
...
+
+		if ( rt.equals(ResidueType.nonPolymer))
+			return false;
+
+		PolymerType pt = rt.getPolymerType();
+
+		return pt.equals(PolymerType.peptide) ||
+				pt.equals(PolymerType.dna) ||
+				pt.equals(PolymerType.rna) ||
+				pt.equals(PolymerType.dnarna);
```
josemduarte: In here we could use the `PolymerType.PROTEIN_ONLY` and `PolymerType.POLYNUCLEOTIDE_ONLY` predefined sets. This is for instance what we have in some methods in StructureTools:

```java
 PolymerType.PROTEIN_ONLY.contains(cc.getPolymerType())
	|| PolymerType.POLYNUCLEOTIDE_ONLY.contains(cc.getPolymerType()
```
Also in StructureTools and other places we check this too:
```java
ResidueType.lPeptideLinking.equals(cc.getResidueType())
```
Would that be useful here?

josemduarte:
I'll merge it in, let's work on those details later. Thanks @andreasprlic !

PR #554, opened 2016-08-09
Struct conn record bonds now made for all models

abradle:
- StructConn bonds now added
 - Added test for this using 1CDR

https://github.com/rcsb/mmtf/issues/15


josemduarte:
Thanks!

PR #555, opened 2016-08-11
Using the new checks for polymer type in Group, ...

andreasprlic:
... instead of repeating similar checks in StructureTools.


sbliven:
Were the pom changes intended to belong to this PR?

biojava-structure/src/main/java/org/biojava/nbio/structure/StructureTools.java
```diff
@@ -1612,12 +1610,9 @@ public static Structure removeModels(Structure s) {
 
 			ChemComp cc = g.getChemComp();
 
-			if (ResidueType.lPeptideLinking.equals(cc.getResidueType())
```
josemduarte: This check is not included in `g.isPolymeric()`. Do we need to add it too? I really can't remember why this is important or not. Ideas anyone?

sbliven: What would be lPeptideLinking but not PROTEIN_ONLY? Non-ribosomal peptides?

andreasprlic:
To comment on the status of this: During the wednesday skype call we agreed that the checks for polymer type should get added also on the ```Chain``` level.

biojava-structure/src/main/java/org/biojava/nbio/structure/Model.java
```diff
@@ -71,7 +71,7 @@ public void addChain(Chain c) {
         EntityInfo info = c.getEntityInfo();
         
         if ( info == null || info.getType() == null) {
-        	logger.warn("No entity info could be found while adding chain with asym id {} (author id {}). Will consider it a polymer chain.", c.getId(), c.getName());
+        	logger.debug("No entity info could be found while adding chain with asym id {} (author id {}). Will consider it a polymer chain.", c.getId(), c.getName());
```
josemduarte: I'd keep this at least on INFO level, this is quite a strong indication that there's something wrong in reading the structure.

andreasprlic: done

PR #563, opened 2016-08-16
Version 1.0.0 of MMTF changes

abradle:
- Fixed issues with reading a few MMTF files (related to hetatm vs amino)
- Fixed a bug in generation of disulphide bonds - including a test case.
- Fixed a null pointer exception from getAtomFromRecord
- Fixed handling of entities with no chains
- Upgrade to v1.0.0
- Fixed the formatting and added some javadocs

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmtf/MmtfStructureReader.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmtf/MmtfStructureReader.java
@@ -29,6 +29,7 @@ import org.biojava.nbio.structure.PDBHeader;
 import org.biojava.nbio.structure.Structure;
 import org.biojava.nbio.structure.StructureImpl;
 import org.biojava.nbio.structure.StructureTools;
+import org.biojava.nbio.structure.io.mmcif.chem.PolymerType;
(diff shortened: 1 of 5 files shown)


biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmtf/MmtfStructureReader.java
```diff
@@ -411,7 +411,13 @@ private int getGroupTypIndicator(String currentGroupType) {
 		}
 		if(currentGroupType.toUpperCase().contains("D-GAMMA-PEPTIDE")){
 			return 0;
-		}		
+		}
+		if(currentGroupType.toUpperCase().contains("D-BETA-PEPTIDE")){
```
andreasprlic: Would be better to use the ResidueType enum here.

andreasprlic:
I think it would be better to replace getGroupTypIndicator(String) with the ResidueType.getResidueTypeFromString. Then it is possible to use the built-in ResidueType and PolymerType classes to identify if a group is an AminoAcid, Nucleotide, or Hetatom. In addition this ensures consistency with the CCD.

abradle:
Thanks Andreas! Good point - fixed.

andreasprlic:
Great thanks!

PR #564, opened 2016-08-16
Added MMTF encoded function return as OutputStream 

abradle:


biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmtf/MmtfActions.java
```diff
@@ -46,6 +48,24 @@ public static void writeToFile(Structure structure, Path path) throws IOExceptio
...
 		WriterUtils.writeDataToFile(writerToEncoder, path);
 	}
+	
+	/**
+	 * Write a Structure object to an {@link OutputStream}
+	 * @param structure the Structure to write
+	 * @return the {@link OutputStream} of the MMTF structure
+	 * @throws IOException an error transferring the byte[]
+	 */
+	public static OutputStream writeToOutputStream(Structure structure) throws IOException{
```
andreasprlic: I was expecting a method where you would pass in an OutputStream as an argument, rather than getting it as an output. ```	write(byte[] b, int off, int len)``` is defined on the OutputStream interface level, as such it should be supported by any of the implementing classes.

abradle:
Yeah good point Andreas - thanks!

PR #566, opened 2016-08-21
GenericInsdcHeaderFormat: Fixed formatting codes.

andviane:
* %r is not a valid formatting code and will result UnknownFormatConversionException. %s is the correct code to format the String as returned by your getStrand() method.

* Do not compare NaNs with ==. NaN is not equal to any number,   self including.

* Do not compare enum with String,  will never be equals.


andreasprlic:
thanks

Issue #355, opened 2015-12-16
org.biojava.nbio.genome.parsers.gff.Location union() and intersect() do not work as described in API


josemduarte:
From Christopher Gillies in biojava-l:

I am using version 4.1.0 and I am getting some unexpected behavior from the Location class.

Please see the unit test below. The union operator returns 51227381 as the bioEnd() but it should be 51227382.

```java
Location l1 = Location.fromBio(51227320, 51227381, '+');
Location l2 = Location.fromBio(51227323, 51227382, '+');

Location union = l1.union(l2);
assertEquals(51227320,union.bioStart());
assertEquals(51227382,union.bioEnd());
```

The documentation for the intersection method says that it will return null when there is no overlap for two locations, but the code appears to fail for boundary conditions. The following unit test fails. Shouldn’t the intersection method return null for this test case?

```java
Location l1 = Location.fromBio(100, 200, '+');
Location l2 = Location.fromBio(1, 99, '+');
Location intersection = l1.intersection(l2);
assertNull(intersection);
```

josemduarte:
I've added the test to branch https://github.com/josemduarte/biojava/tree/issue355. The first test does work, the second doesn't and I suspect it is a bug in the `intersection` method.

christopher-gillies:
Thanks Jose. You are correct. The first issue was my mistake.

josemduarte:
What happens is that `Location.fromBio` converts `start`, `end` coordinates to `start-1` and `end` (in case above it would be `l1=99,200`, `l2=0,99`). Based on that then there is an overlap on 99 for your test case above. Could that be the desired behaviour? I don't really know enough about this part of Biojava to judge.

christopher-gillies:
While it is possible that is the desired behavior, I do not think it is. However, I am new to biojava so I could be wrong. But I think this is not the correct behavior because of what the documentation says:

"Internally, location indices are stored in Java "half-open" format: the start is the (origin 0) index of the first symbol in the range; the end is the origin 0 index of the first symbol PAST the end of the range, so that end - start == length. "

So the end coordinate is the first symbol PAST the end range. So l2=[0,99) and 99 is the index of the first symbol PAST the end range. For l1=[99,200) the start is the first symbol in the range. So l2 contains no elements in l1 because the end value is non-inclusive. Therefore, there is no overlap between the two intervals. The length of the intersection result also validates this interpretation. See the code below:

Location l1 = Location.fromBio(100, 200, '+');
[...]

PR #567, opened 2016-08-24
Proposed fix for #355

andviane:
Interval intersection is wrongly computed. Here is the proposed alternative implementation. It passes the new TestIssue355 Unit test that exists in another branch.


andviane:
The second 'Cleanup' commit was only intended to restore the by chance applied formatting, without any changes in the code itself. Really no idea why it failed. Maybe something system dependent? There are messages about the PDB_DIR property being accessible during the build.

josemduarte:
Thanks! Could you merge in the TestIssue355 branch? or otherwise simply add the test manually? We need to make sure that the test always runs. I suspect Travis failures are related to some download problem. Let´s see if it builds after you push another commit.

andviane:
Ok done.

josemduarte:
Thanks! I'll merge it

josemduarte:
We'll need to add this to the bugfixes branch too

PR #568, opened 2016-08-25
Add protein alias and gene alias to uniprot extract

emckee2006:


andreasprlic:
Just to add: for a complete representation of UniProt in a Java datamodel, see here:

https://github.com/rcsb/uniprot-or-mapping

emckee2006:
Doesn't that only work if one is persisting it in a database in that format?

andreasprlic:
You can just read from UniProt XML, without a DB.

andreasprlic:
Just added some more docu there to provide an example how to read a UniProt XML into a Java object. For example this can be done:

```java

            URL u = UniProtTools.getURLforXML(accession);
            InputStream inStream = u.openStream();
            Uniprot up = UniProtTools.readUniProtFromInputStream(inStream);
```

emckee2006:
Any examples on how to do that?  That may be easier then using the proxy
reader (currently i download the full tar.gz from uniprot, read through it
record by record, and pass each to the proxy reader:
                        UniprotProxySequenceReader<AminoAcidCompound> upsr
= UniprotProxySequenceReader
                                .parseUniprotXMLString(
                                        "<uniprot> " + xml.toString()
                                                + " </uniprot>",
                                        new AminoAcidCompoundSet());

I could see an advantage of having it all in our db, though.  However, we
are an oracle shop and would have to create the schema ahead of time (and
change the hibernate settings appropriately).  Does is also remove obsolete
entries/update entries which have changed since last load, or is it just
lead each record once?

andreasprlic:
We can reload all of UniProt (and a small subset of Trembl) over night, so we never bothered with writing an incremental update strategy.

emckee2006:
so you just tend to dump it all and reload?  With some sort of script which
truncates the tables?

emckee2006:
Also, does everything from the uniprot record currently go in the database,
or just some of it?

andreasprlic:
- We just start with an empty database, let hibernate create all the tables and the we can load whatever we want (see LoadMissing class). We have not tried to load all of Trembl yet, but it works fine for all of SwissProt plus the subset of Trembl that is linked to PDB.

- Yes it is a complete mapping of all of the UniProt data into the database/ Java data model.

We should move this thread over to the other project :-)

emckee2006:
How do we do that?
Is there a way to get a dump of the schema so i can create it locally?
Does that schema change from time to time?

andreasprlic:
UniProt schema changes quite regularly. The project has a built in schema version that is used as a default (see file uniprot.xsd in the resources folder).  I'll set up a documentation page for how to run a database load.

emckee2006:
So it's not feasible to have the db schema before hand, and always just
truncate and load the db?

andreasprlic:
No, the schema gets generated at compile time. (The datamodel is auto-generated by the schema). As such this is available also before hand. That's why the database in principle is not needed and the code works also when only UniProt XML files are available. Truncating an existing db and just loading is possible.

andreasprlic:
here some documation for how to load a DB
https://github.com/rcsb/uniprot-or-mapping

PR #569, opened 2016-08-26
Update Jmol

sbliven:
Also uses mmCIF for communication with Jmol.

This was an important step for #462 to make chain renaming more robust. It also takes a step towards #201, although MMTF (or direct Java data structures) might make more sense.

Issue #201, opened 2014-11-05
BiojavaJmol transmits structures naively


sbliven:
BiojavaJmol transmits structures to Jmol by converting them to a PDB string. This is very simplistic and limits some features:
* Unit cells don't display, since the header records get stripped out, including crystal information
* large structures are going to break, since they don't have a PDB representation (#156)

lafita:
Another problem has popped up: with the current communication it is impossible to display two (or more) aligned *Biological Assemblies*, since the **symmetry generated chains** are also stored in different models. See #220 for a more detailed description of the problem.

sbliven:
Officially solved with #569.

andreasprlic:
great!

PR #570, opened 2016-08-30
Switch Jmol interaction to use MMTF

sbliven:
MMTF write support is much better than mmCIF. I'm assuming it will be faster as well.

### DIFF
--- a/biojava-structure-gui/src/main/java/org/biojava/nbio/structure/align/gui/jmol/JmolPanel.java
+++ b/biojava-structure-gui/src/main/java/org/biojava/nbio/structure/align/gui/jmol/JmolPanel.java
@@ -24,12 +24,33 @@
 
 package org.biojava.nbio.structure.align.gui.jmol;
 
-import org.biojava.nbio.structure.*;
+import java.awt.Color;
+import java.awt.Dimension;
+import java.awt.Graphics;
+import java.awt.Rectangle;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import java.io.BufferedInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
+import java.text.DecimalFormat;
+import java.util.List;
+
+import javax.swing.JComboBox;
+
+import org.biojava.nbio.structure.Atom;
+import org.biojava.nbio.structure.Calc;
+import org.biojava.nbio.structure.Group;
+import org.biojava.nbio.structure.Structure;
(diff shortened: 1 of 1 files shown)


biojava-structure-gui/src/main/java/org/biojava/nbio/structure/align/gui/jmol/JmolPanel.java
```diff
@@ -114,11 +127,28 @@ public void executeCmd(String rasmolScript) {
...
+			new Thread((Runnable)() -> {
+				try {
+					MmtfActions.writeToOutputStream(s,out);
+				} catch (Exception e) {
+					logger.error("Error generating MMTF output for {}",
+							s.getStructureIdentifier()==null ? s.getStructureIdentifier().getIdentifier() : s.getName(), e);
+				}
+			}).start();
+			viewer.openReader(null, in);
+		} catch (IOException e) {
```
josemduarte: How about throwing the exception? In the end of the day this will only produce a blank jmol screen (and a message in the logs), but a user will not be able to know if there was a problem in the loading. If we throw it then a user can catch it down the line and act upon it.

josemduarte:
Nice thanks a lot! This would finally solve #201 right?

sbliven:
Actually, #201 was already solved by PR #569, which switched from PDB to mmCIF.

biojava-structure-gui/src/main/java/org/biojava/nbio/structure/align/gui/jmol/JmolPanel.java
sbliven: The problem is that the interesting exception is the one from writeToOutputStream, and that has to run in its own thread if I want to use pipes like this, so it can't throw an exception.

The other IOExceptions would indicate a broken pipe to Jmol, which would very likely be from program termination anyways. Trying to throw an IOException here requires adding the exception to maybe a dozen other functions that call this, some of which don't make much sense.

I guess I could write the MMTF to a byte[] and then pass it to jmol (avoiding creating a new thread).

I guess ultimately I caught the exceptions here because they seemed very unlikely to result from user input (assuming the MMTF encoder doesn't have bugs).

josemduarte: Thanks for the explanation, that makes sense now.

> I guess ultimately I caught the exceptions here because they seemed very unlikely to result from user input (assuming the MMTF encoder doesn't have bugs).

Well in theory if the structure is "malformed" in some way the encoder should throw an exception saying it doesn't like it. In real life I'm not sure if such a "malformation" is a possibility.

PR #571, opened 2016-08-30
Quaternary Structural Alignment Algorithm (QsAlign)

lafita:
Create a **structural alignment** algorithm for **macromolecular complexes** in **polynomial time** with the number of chains.

The algorithm builds on top of the `Subunit` clustering code written to refactor the quaternary symmetry detection code in **PR** #520 and #524.

The entry class is called `QsAlign`, which has methods to align a pair of `Structures` or a pair of `Subunit` collections. The output is a `QsAlignResult`, which contains the `Subunit` equivalencies between the pair and their full residue alignment.

Unlike other popular macromolecular complex alignment algorithms like [MM-align](http://zhanglab.ccmb.med.umich.edu/MM-align) and [VAST+](http://www.ncbi.nlm.nih.gov/Structure/vastplus/vastplus.cgi), `QsAlign` does not require the combinatorial ennumeration of all possible `Subunit` matchings. 
This is achieved by building the alignment from each possible `Subunit` equivancy between the two `Structures` `O(N^2)` and adding iteratively other `Subunit` equivalencies `O(N^2)` that fulfill a quality measure threshold. The quality measure thresholds implemented so far are: **centroid distance**, **orientation angle deviation** and **RMSD**. In the end the longest and lowest RMSD alignment is chosen as the optimal. The total running time is `O(N^4)`.

In addition, `QsAlign` is much faster for negative cases (when the alignment is not possible because `Subunits` are structurally non-equivalent), because it performs an initial clustering of the `Subunits` between `Structures` `O(N^2)`, before starting to build the alignment. In cases where the two `Structures` do not have any structurally equivalent `Subunit`, the algorithm returns an empty alignment, thus avoiding trying the macromolecular complex alignment.

Another advantage of the algorithm is that it is independent of chains, because a `Subunit` can also be a domain of a chain, like in the quaternary symmetry detection.

What the algorithm assumes is that the individual `Subunit` pairwise alignments are equal to their pairwise alignment in the complexes alignment, and that one `Subunit` of a `Structure` can only have one matching to the `Subunits` of the other `Structure` (assumed also by the other algorithms). It could potentially also have borderline cases for the definition of the quality measures, since they are hard thresholds, while the other algorithms try all possible combinations and choose the best.

I also created a test `TestQsAlign` with some examples from papers about macromolecular complexes alignment, and the results are the expected. I did not compare, however, the running time of the algorithm.


biojava-structure/src/main/java/org/biojava/nbio/structure/Calc.java
```diff
@@ -763,7 +764,11 @@ public static final void shift(Group group, Atom a) {
 	}
 
 	/**
+<<<<<<< HEAD
+	 * Returns the centroid of the set of atoms.
+=======
```
josemduarte: This looks like something went wrong in merging

abradle: Nah - it's a novel bit shift operation. See Effective Java - Chapter 72,
Section 4. ;-)

lafita: Fixed! My Effective Java edition only has 11 chapters (I had to check to realize it was a joke).

lafita:
I have moved long tests for **quaternary symmetry detection**, **quaternary structure alignment** and **subunit clustering** to the `integrationtest` module.

Just to explain my experience so far with testing in **biojava**, when creating tests I usually tend (and prefer) to use fake **Atom arrys** (like poly-alanine). 
However, this only works for testing very low level parts of the algorithms, where only coordinates are used. When higher level parts are tested to check for the overall correctness, more information (like parent `Groups` with its **chemcompinfo**) is required, and creating this fake (unbiased) information requires a lot of work. In addition, the data created this way can be oversimplified to the complexity of real protein structures.

I usually tend to select a structure, or a few, that have the desired property the algorithm is intended for and use it to test the overall correctness (meaning that the algorithm has the correct output). This type of tests are not unit tests in practice, because they require downloading and parsing the structures, extracting the representative atom array, etc., but a user that wants to use the algorithm will have to go through these same steps. So in some way it is a test that a user will be able to run the algorithm and obtain the correct output, although a failure might be due to the parser.

josemduarte:
Agree with what you say about tests. In general we should strive to have actual unit tests (e.g. the tests you describe with fake coordinates) in the individual modules and then integration tests that test realistic cases in the `integrationtest` module. Right now both on mvn command line and travis CI all tests run always, but at some point we could decide to do something special with `integrationtest`, so it's good to maintain a separation between both.

The problem is that we do still have a lot of integration tests in the structure module. With time we should try to convert them into unit tests and extract the more comprehensive ones into `integrationtest`.

lafita:
I did a schematic representation of an example alignment to demonstrate how the algorithm works.

sofla_112_216090211460_0001

The essential additional assumption compared to other macromolecular complex alignment algorithms becomes evident here: transformation matrix **T1** (the pairwise alignment of two single subunits used as the starting point) has to be a good approximation of **T4**, the transformation matrix of aligning the whole macromolecular complex (all subunits). 

This implies that the pairwise alignment of two given subunits equivalent in the macromolecular complex alignment is the same (or very similar) to their alignment in the whole macromolecular complex alignment.

The algorithm is exploiting this independence between each subunit alignment to calculate more efficiently the possible subunit equivalencies. Without this assumption, a combinatorial ennumeration of all possible subunit equivalencies is needed.

Issue #572, opened 2016-08-30
Switch to flattened biounits instead of multimodel biounits


josemduarte:
At the moment we use multimodel biounits in all algorithms related to quaternary structure. That is an issue for multimodel structures (e.g. NMR or some X-rays) where the multiple models would be taken as extra subunits when they aren't. 

Biojava can produce flattened biounits since #490, i.e. symmetry partners are expanded as new chains with new chain ids. I'd advocate moving to flattened biounits for the 5.0 release. We'd need to make all quaternary structure code compatible with that (simply by not looking at all models). I would then keep the possibility of producing biounits as multiple models but without using them internally.

Anything against that?

lafita:
At the moment there is a flag in a `Structure` object to determine if the multiple models are symmetry partners of a biological assembly. All models are only taken if the `Structure.isBiologicalAssembly()` method returns **true**. What should we do with this flag? Does it still make sense?

josemduarte:
I didn't realise we had the flag: it's true that using it we can know about multiple models being part of the biounit or not. In any case doing the flattened biounits still has the advantage that one could in theory look at different biounits of different models (impossible now). So I'd still advocate for the cleaner solution of flattening which in the long run will work better.

> What should we do with this flag? Does it still make sense?

I think it wouldn't hurt to keep the flag, we would simply not need it anymore for this purpose.

PR #573, opened 2016-09-03
Solving #572

josemduarte:
This switches the biounit generation by default to multichain biounits as discussed in #572


andreasprlic:
can we update the tutorial for this? https://github.com/biojava/biojava-tutorial/blob/master/structure/bioassembly.md

lafita:
@andreasprlic I created an issue in the tutorial to remember updating it.

josemduarte:
I've updated the tutorial now

PR #576, opened 2016-09-09
making SVDSuperimposer accessible from the outside

andreasprlic:
A small extraction of a block of code into a method, to allow the code to be a little more re-usable.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/io/FileConvert.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/io/FileConvert.java
@@ -408,6 +408,10 @@ public class FileConvert {
 
 
 		Character  altLoc = a.getAltLoc()           ;
+		
+		if ( altLoc == null)
+			altLoc = ' ';
+		
 		String resseq = "" ;
 		if ( hasInsertionCode(pdbcode) )
 			resseq     = String.format("%5s",pdbcode);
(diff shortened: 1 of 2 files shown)


andreasprlic:
I think we have a concurrency issue. Some junit tests are failing with ArrayIndexOutOfBounds exceptions. When re-running them locally they run fine.

lafita:
I am currently working on issue #558, and part of it is to make the superposition algorithms more easily accessible. The branch I am working on is https://github.com/lafita/biojava/tree/superpose.

For this particular method I think it would be better to create a method that calculates the two arrays of aligned atoms given an `AFPChain` and the two arrays of **original atoms**, and then use the `SuperPosition` interface to calculate their superposition. 

I think that a method that returns a **SVDSuperimposer** object is against the decisions we agreed on the discussion in #558.

josemduarte:
@lafita, following your comments and your PR #579, I guess we need a different solution for this, don't we? Could you have a look?

biojava-structure/src/main/java/org/biojava/nbio/structure/io/FileConvert.java
```diff
@@ -408,6 +408,10 @@ public static void toPDB(Atom a, StringBuffer str, String chainID) {
 
 
 		Character  altLoc = a.getAltLoc()           ;
+		
+		if ( altLoc == null)
+			altLoc = ' ';
+
```
josemduarte: Good catch @altaite ! It would be better if you commit this bug fix to the master branch directly instead of having it as part of an unrelated pull request. It would also be good to cherry-pick it to the bufixes-4.2 branch so that it goes to next bug fix release. Thanks!

lafita:
@josemduarte yes we need a different solution. I will make another PR to substitute this one.

josemduarte:
Alright, then we will have a new PR for the superimposer issue. The bugfix by @altaite will have to be committed to master. I'll close this one.

PR #578, opened 2016-09-10
Test atom cache parsing flags

andreasprlic:


andreasprlic:
This fixes a bug with setting the correct file type to use in AtomCache. Previously: both mmtf and mmcif flags were set to true.

josemduarte:
This seems to contain PR #576 too, should we close that one and keep both fixes together here?

andreasprlic:
Ah yes, you could just merge this one and close both...

andreasprlic:
I reverted the changes in AlignmentTools to the version of the master branch.

Issue #303, opened 2015-07-20
Replace Calc methods with Vecmath, instead of JAMA


lafita:
The Calc class contains methods to calculate geometric operations on the Structure interface objects (basically Atoms). 
Now that the Vecmath dependency has been added to the structure module, some methods have been added to convert from the old JAMA matrices and vectors to the Vecmath objects. 

Some problems have already occurred in the conversion, basically because the JAMA rotation matrices are pre-multiplication, and Vecmath uses post-multiplication matrices to transform points and vectors (which is the convention in 3D geometry). See 38c1e33. 
This means that to obtain the same transformation the matrices have to be transposed (inverted). The current Calc methods that convert between the two matrix representations ensure that this conversion is made (since 38c1e33).

However, it would be a step forward to replace completely all JAMA matrices in the Calc class to the Vecmath ones, so that no conversion methods or different matrix conventions are used. 
This would require work ensuring that the basic algorithms, like SVD calculation, return the appropriate transformation matrix in the post-multiplication form (as Vecmath requires).

Another TODO would be to create some tests for the Calc class (if possible, before changing to the vecmath implementation of the algorithms) that ensure the same behavior as the old code.

andreasprlic:
To be more backwards compatible, it would be great if Calc would initially support both Vecmath and JAMA...

lafita:
Both are supported right now in Calc, but there is this confusion with the transposition of the rotation Matrix. The bug should be fixed now.
The only thing to be done is a test, then?

josemduarte:
From my understanding the key piece of code is in `Calc.rotate(Atom,Matrix)` and `Calc.rotate(Atom,double[][])`. There the vector is pre-multiplied. Since the usual convention is to post-multiply, this means that all rotation matrices that use `Calc.rotate` are in effect transposed with respect to the usual rotation matrices found in textbooks or elsewhere. 

That really can cause some confusion and should be avoided. I'd definitely advocate for unifying the code as much as possible, in my opinion the best solution would be moving everything to vecmath, making sure that there's nothing needed from JAMA (SVD perhaps?). As @lafita says, provided we have good tests for it, doing this change should not be too hard. Surely the hardest part is to write good tests that cover well the different cases.

andreasprlic:
[...]

PR #579, opened 2016-09-13
Implement a Superposition Interface

lafita:
This PR addresses #558, with the following changes:

- Implement a `SuperPosition` interface to unify the three superposition algorithms in **biojava**.
- Implement a `SuperPositions` Class with a singleton superposition object, so that the default superposition algorithm used in **biojava** can be easily changed and to provide an easy access to single superpositions.
- Move some methods in `Calc` operating on transformation and rotation matrices to a new Class `Matrices`.


biojava-structure/src/main/java/org/biojava/nbio/structure/geometry/SuperPositionQCP.java
```diff
@@ -107,20 +107,13 @@
...
-	 * The required eigenvector precission
-	 */
-	private static final double EVEC_PREC = 1E-6;
-	
-	/**
-	 * The required eigenvalue precission
-	 */
-	private static final double EVAL_PREC = 1E-11;
+	private double EVEC_PREC = 1E-6;
+	private double EVAL_PREC = 1E-11;
```
josemduarte: These 2 should be lower case if they are variable fields

josemduarte:
This looks fantastic! 

Should we close PR #576 then? If I understand it, this PR would solve the issue discussed there, wouldn't it?

andreasprlic:
#576 is a small patch of alignment tools. I can redo this later once this has been merged in.

biojava-structure/src/main/java/org/biojava/nbio/structure/geometry/SuperPositionQCP.java
lafita: Fixed! Thanks

lafita:
I think we can close #545 with this PR. The only geometry issue that has still to be addressed is #303.

andreasprlic:
can we make AtomImpl use an internal Point3D to store coords?

josemduarte:
I'll merge this. I've opened a new issue for the AtomImpl thing #582

PR #580, opened 2016-09-14
Get one more type of protein alias

emckee2006:


josemduarte:
It looks like this could go to the next bugfix release. Could you cherry pick it back to bugfixes-4.2? Thanks!

Issue #582, opened 2016-09-14
Change AtomImpl to use Point3d for coordinates


josemduarte:
The main advantage of this is that moving between Atom[] and Point3d[] representations would be a lot lighter.

PR #583, opened 2016-09-15
Solving issue 582

josemduarte:
This solves #582 enabling a faster and less memory intensive translation between `Atom[]` and `Point3d[]`. I've also made `AsaCalculator` and `Grid` (contact calculation algorithm) more independent of `Atom[]` following the comments of @pwrose in #545, now it's possible to use both algorithms with either just `Point3d[]` or with `Atom[]`.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/Atom.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/Atom.java
@@ -25,6 +25,8 @@ package org.biojava.nbio.structure;
 
 import java.util.List;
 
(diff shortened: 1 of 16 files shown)


andreasprlic:
I changed the project config to now require travis to build successfully before merges are possible.  This was not enabled before, however github was showing a strange message that it was still waiting for checks to pass and did not allow to merge. I think that is  a GitHub bug. Let's keep an eye on this config and change this back, once the issue has been resolved.

andreasprlic:
not sure why this still shows "Required statuses must pass before merging". :-/

PR #591, opened 2016-09-28
Add ligands to subselections by proximity

### LINKED ISSUE #560: Structures with residue ranges don't load atoms
Currently, if you specify a residue range for `getStructure` and related methods then it won't load any ligands. Examples: 4hhb.A:1-140 (no heme), 3hz2.A:1-84 (no Calciums)

In one sense this is correct; you have specified a polymer property, so it makes sense to only get polymer residues back. However, more often than not you do want to include any surrounding ligands, particularly since we don't have a full selection language that would allow ligands to be specifically included or excluded.

I propose that residue ranges automatically include any ligands within 7Å of the selection. The simplest rule would be to include all non-polymer atoms within 7Å, regardless of their chain. This sidesteps any issues with the new policy of assigning ligands to their own chains, although it could lead to duplicate ligands when loading several disjoint pieces of a structure independently.

sbliven:
Fix #560.

SubstructureIdentifier now expands its selection to any ligands within 5A of the selection.
This also gets done for CE-Symm when creating a new structure.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/SubstructureIdentifier.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/SubstructureIdentifier.java
@@ -352,7 +352,8 @@ public class SubstructureIdentifier implements Serializable, StructureIdentifier
 	protected static void copyLigandsByProximity(Structure full, Structure reduced, double cutoff, int fromModel, int toModel) {
 		// Geometric hashing of the reduced structure
 		Grid grid = new Grid(cutoff);
-		grid.addAtoms(StructureTools.getAllAtomArray(reduced,toModel));
+		Atom[] nonwaters = StructureTools.getAllNonHAtomArray(reduced,true,toModel);
+		grid.addAtoms(nonwaters);
 
 		
 		// Find ligands from full structure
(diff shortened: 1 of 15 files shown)


sbliven:
I need to fix some tests

biojava-structure/src/main/java/org/biojava/nbio/structure/SubstructureIdentifier.java
```diff
@@ -316,4 +321,97 @@ public Structure loadStructure(AtomCache cache) throws IOException, StructureExc
 		return cache.getStructureForPdbId(pdb);
 	}
 
+	/**
+	 * Supplements the reduced structure with ligands from the full structure based on
+	 * a distance cutoff. Ligand groups are moved (destructively) from full to reduced
+	 * if they fall within the cutoff of any atom in the reduced structure.
+	 * The {@link StructureTools#DEFAULT_LIGAND_PROXIMITY_CUTOFF default cutoff}
+	 * (currently 7Å) is used.
```
lafita: It was changed to 5A. Maybe it is better no to specify in the javadoc what the cutoff value is.

sbliven:
This PR has some somewhat undesirable side effects in cases with PTMs. In BioJava 4 these are treated as ligands, so they get included with nearby selections. In BioJava 5 I think this should not be a problem since we handle PTMs correctly.

Question: should this be merged into bugfixes-4.2, or should I include it only in the 5.0 line?

sbliven:
Consensus was to only add this to 5.0 due to the ligand annotation problems in 4.2.

PR #595, opened 2016-10-26
Add some handling of "?" in seq align position. E.g. for 5dn6

abradle:
One entry in the archive has a "?" in the seq_align_beg field. This causes a number format exception. According to mmCIF this is legitimate. So Biojava should handle this gracefully.


josemduarte:
I'll merge it now that the build is finally passing.

PR #597, opened 2016-11-07
Improve ligand handling of substructures

### LINKED ISSUE #560: Structures with residue ranges don't load atoms

sbliven:
Fixes #560.

Clearly define ligand handling:
- Full-chain selections ('1abc.A') use all groups with that chain name, including waters and ligands
- Residue selections (1ab.A:1-100) do not include waters, and they may include ligands from other chains if they appear within the cutoff distance (5A default)

This PR also bundles some structural changes:
- Use StructureTools.addGroupsToStructure where ever you want to create a new structure. It handles the chain creation as needed.

This code was originally written for the 4.2 branch, but then moved to the 5.0 branch when it became apparent that it would change structure loading too much. As a result, during code review please pay special attention to places where the new chain model is used (e.g. ligands get their own chain), or where the new polymer/nonpolymer/water dicotomy comes into use. For instance, "ligand" is just defined as nonpolymer, and relies on the input structure for correct classification of PTMs. Another tricky place is figuring out when cloning groups and other data structures is necessary.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,0 +1,4 @@
+BioJava 5.0.0
+=============
+
+- For short structure selections (e.g. 1abc.A:1-100), ligands within 5A will be included
--- a/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/StructureToolsTest.java
+++ b/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/StructureToolsTest.java
@@ -105,21 +105,18 @@ public class StructureToolsTest extends TestCase {
 		Structure hivA = cache.getStructure("1hiv.A");
 		Atom[] caSa = StructureTools.getRepresentativeAtomArray(hivA);
 		Atom[] caCa = StructureTools.getRepresentativeAtomArray(hivA.getChainByIndex(0));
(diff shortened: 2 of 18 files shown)


josemduarte:
The travis build seems to be stuck, any ideas why? I can't restart it or anything.

andreasprlic:
another Travis bug related to approving pull requests? @sbliven can you do another push to this branch, if that does not help, I'll contact github support again.

sbliven:
Ok, Travis is re-running

PR #598, opened 2016-11-09
Fix #596. Remove test for GenBank identifier parsing

### LINKED ISSUE #596: Genbank files changed format: GenbankProxySequenceReaderTest is broken
We found out today that the genbank tests are failing, specifically `GenbankProxySequenceReaderTest`. After digging a bit into it I found this was related to a change just implemented in genbank, this is the relevant extract from ftp://ftp.ncbi.nih.gov/genbank/gbrel.txt :

```
 1.4.3 GI sequence identifiers to be removed from GenBank/GenPept/FASTA formats
      and FASTA header to be simplified

                        **NOTE**
  The removal of GI identifiers from GenBank (and RefSeq) Release and
  Update products has been postponed until March of 2017. NCBI will initially
  implement this change only in the Entrez system, in order to better gauge
  the impact that it will have on users. We expect that the Entrez change will
  take place in late October or early November of 2016.
                        **NOTE**
```

Basically the VERSION line of the gb files used to look like:

```
VERSION     U49845.1  GI:1293613
```

and now it looks like this:

```
VERSION     SCU49845.1
```

sbliven:
Starting Nov 2016, Entrez stopped including the GI identifier in
GenBank files. In such cases, getAccession().getIdentifier() will
return null.

This updates the tests to reflect that null identifiers are acceptable.

Users who need the GI identifier should store it manually after fetching
the sequence.

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/sequence/AccessionID.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/sequence/AccessionID.java
@@ -133,6 +133,8 @@ public class AccessionID {
 
 	/**
 	 * In case if {@link #getID() } in not unique keeps the alternative id, eg. NCBI GI number.
+	 * 
+	 * This may null.
 	 *
 	 * @return
 	 */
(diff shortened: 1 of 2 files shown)


sbliven:
Link #596

sbliven:
I'm going to merge this through without review since it's blocking additional development.

Issue #557, opened 2016-08-11
Accept overly long residue ranges?


sbliven:
Sometimes I find myself annoyed with having to look up the valid range of residues for a structure. For instance, with symmetric and circularly permuted proteins I often run parallel analyses on each half of a protein (say, `3hz2.A:1-42` vs `3hz2.A:43-84`). Although ResidueRange allows missing residues within a range, the endpoints must be valid residues, so I'm forced to look up that 2hz2.A contains 1-84.

Is there a good way to extend our selection syntax to allow this?

Possibilities:
1. Accept invalid residues, and silently (or with a warning) trim them to the first/last valid residues. Example: `3hz2.A:-999-999`
2. Add special symbols to indicate first and last residues. Example: `3hz2.A:^-$`
3. Allow the sides of the range to be empty. Examples: `3hz2.A:43-` (second half), `4hz2.A:-+42` (first half), `4hz2.A:-` (all)

The first option has some subtleties regarding insertion codes, but these are already handled by `AtomPositionMap.trimToValidResidues` in a reasonable manner. A bigger problem is that this could mask errors in cases that should have valid ranges, such the recent problems with ECOD ranges. It also incurs a small performance penalty on loading ranges.

Would this feature even be worth incorporating, or is it unnecessary complexity for BioJava?

sbliven:
c5be5bc simply uncomments some existing code for option 1. It's not necessarily an endorsement of that option.

sbliven:
I think consensus was to implement both 2 and 3 (but not 1)

PR #600, opened 2016-11-10
Allow partial residue ranges (#557)

sbliven:
Allow partial residue ranges (#557)

- Introduce '^' and '$' symbols to stand for the beginning and end of the
  chain. E.g. "A:^-100"
- Accept partial ranges for cases where they could not be confused with
  negative residue numbers. E.g. "A:-5" refers to a single residue, while
  "A:-+5" gives all residues up to and including 5.


sbliven:
This is a pretty simple change. Can I get a review please?

PR #604, opened 2016-11-22
Implementing parsing of atom_sites category in mmcif parser/consumer

josemduarte:
A small patch to be able to parse the atom_sites category in mmcif. The category contains the same data as SCALE matrix records in PDB format. This can be useful to check the scale matrix is consistent with the standard reference frame, see https://github.com/eppic-team/owl/issues/4


josemduarte:
@sbliven agree about having tests not so spread. I had a quick look and we really have quite a lot of related tests spread over a number of files. All that could be rationalised a bit. I think we could leave that for another PR, perhaps before the 5.0 release. It would also be good to move to the integration-tests module those tests that belong there.

josemduarte:
BTW I will cherry pick this commit to the `bugfixes-4.2` branch too.

PR #605, opened 2016-11-22
Implementing new crystallographic metadata fields.

josemduarte:
A follow up of #604. A minor addition: 2 new methods in PDBCrystallographicInfo to check for non-standard space groups and coordinate frame convention. I've created a more comprehensive test and moved it from structure module to integrationtest module.


josemduarte:
I'm intending to add this to bugfixes-4.2 if there's nothing against it.

josemduarte:
Well one can store the SG string and the matrix or store the booleans instead. Storing the booleans seems better in terms of simplicity of code and memory footprint. It only adds a (minimal) overhead of checking things at parse time.

josemduarte:
Regarding PDB format files: I've now pushed a commit that handles non-standard space groups when loading from PDB format file. But I didn't do the same for the scale matrix check because it requires implementing the parsing of that record. Anyway this checks only apply to old PDB-deposited files and we should only use mmCIF format for those archival files. For non-deposited files coming from users this non-standard things are not a concern at all, since the softwares that produced them are long dead.

sbliven:
Merged to master, but I'll leave the cherry-pick to @josemduarte

josemduarte:
Thanks! I've done the cherry-pick.

andreasprlic:
cool, thanks!

PR #607, opened 2016-11-25
Perameterize the search.io package

### LINKED ISSUE #354: Hsp class is not typesafe
The Hsp class and classes that use it (e.g. BlastXMLParserTest) give lots of 'Raw type' warnings. In this case these are not benign warnings, but flag a real possibility for some type errors. The problem is that we're trying to instantiate a sequence from a string without knowing what type of sequence that is. For instance, if you built an Hsp<ProteinSequence, AminoAcidCompound> with a string like "ACTG" then it would actually create a DNASequence, potentially causing errors.

I'm not sure what the best pattern to avoid this is. Maybe the constructor to Hsp needs to take a pair of SequenceReader objects, which can be supplied by the calling code (e.g. BlastHspBuilder).

Original author: @paolopavan

sbliven:
Fixes #354 by adding generic types to basically all classes in the search.io package.

The potentially risky operation of extracting a Sequence from a String is abstracted to a function that gets passed in, allowing the full parsed data structure to have internally consistent type. For instance,

```java
BlastXMLParser<DNASequence,NucleotideCompound> parser = new BlastXMLParser<>((seq) -> {
	try {
		return new DNASequence(seq,AmbiguityDNACompoundSet.getDNACompoundSet());
	} catch (CompoundNotFoundException e) {
		return null;
	}
});
```

It is also possible to do this without knowing a concrete sequence type at compile time, but it requires using a helper function for wildcard capture. For instance, `SequenceTools.getSequenceFromString(String)` returns a `Sequence<?>`. The helper to get results looks like:

```java
private <S extends Sequence<C>, C extends Compound>
List<Result<S, C>> getResults(File file, Function<String,S> buildSeq) throws IOException, ParseException {
	BlastXMLParser<S,C> instance = new BlastXMLParser<>(buildSeq);
	instance.setFile(file);
	List<Result<S,C>> result = instance.createObjects(1e-10);
	return result;
}

List<? extends Result<?,?>> results = getResults(file,SequenceTools::getSequenceFromString);
```

This is a bit tedious and verbose, but it is possible to create type safe parsers even without knowing the concrete type.

The one remaining part of the package that is not type safe is `SearchIO`, which sets up the package to be extensible via SPI. This is perhaps a nice feature, but the reflection required in SPI dangerously violates the parametric types in several places. I have deprecated the class and recommend that we drop SPI support.


sbliven:
Travis is reporting a compilation failure. I'm able to compile fine using both eclipse and maven (which uses javac), so this surprised me. I am using a slightly newer javac version than travis (javac 1.8.0_101 compared with _31), so I guess it's possible that Oracle fixed a compiler bug since then. Generics are pretty much the only place I would even consider looking for a compiler bug anymore. I will try to replicate the error.

sbliven:
Also compiles fine with 1.8.0_20.

andreasprlic:
That's the current error:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.5.1:testCompile (default-testCompile) on project biojava-core: Compilation failure
[ERROR] /home/travis/build/biojava/biojava/biojava-core/src/test/java/org/biojava/nbio/core/search/io/blast/BlastXMLParserTest.java:[65,17] method testCreateObjectsUntypedHelper in class org.biojava.nbio.core.search.io.blast.BlastXMLParserTest cannot be applied to given types;
[ERROR] required: java.util.function.Function<java.lang.String,S>
[ERROR] found: SequenceTo[...]tring
[ERROR] reason: inferred type does not conform to upper bound(s)
[ERROR] inferred: org.biojava.nbio.core.sequence.template.Sequence<?>
[ERROR] upper bound(s): org.biojava.nbio.core.sequence.template.Sequence<capture#1 of ?>
```

sbliven:
I've emailed travis about this error. I'd at least like to reproduce it locally.

paolopavan:
Hello Spencer,
I cannot check your solution since your unit test doesn't work.
I'm having this very same error, even with the most recent JDK.

Any suggestion?
Paolo

2016-11-28 10:32 GMT+01:00 Spencer Bliven <notifications@github.com>:

sbliven:
@paolopavan That's actually good, since I couldn't reproduce it. Can you post your javac version and details about your environment? Is it failing with eclipse/intelleJ, or just with maven?

If you want this to compile you can just comment out the testCreateObjectsUntyped test. All the functionality is tested by testCreateObjects; I was just trying to demonstrate that it is possible to write code that is agnostic of the type of sequence.

paolopavan:
java -version
java version "1.8.0_111"
Java(TM) SE Runtime Environment (build 1.8.0_111-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.111-b14, mixed mode)

I'm using netbeans ide and this is the maven command that it uses:
cd
C:\Users\pavanpa\Documents\Programming_workspace\sbliven-biojava\biojava-core;
"JAVA_HOME=C:\\Program Files\\Java\\jdk1.8.0_111" cmd /c "\"\"C:\\Program
Files\\NetBeans 8.1\\java\\maven\\bin\\mvn.bat\"
-Dmaven.ext.class.path=\"C:\\Program Files\\NetBeans
8.1\\java\\maven-nblib\\netbeans-eventspy.jar\" -Dfile.encoding=UTF-8
install\""




2016-11-28 16:05 GMT+01:00 Spencer Bliven <notifications@github.com>:

sbliven:
I'm baffled by this. I even created a docker image to test this in isolation, and it compiles fine with every version I can find. For now I guess I will remove that test.

josemduarte:
I can also reproduce the compilation failure on Ubuntu Linux 16.04 and javac (open jdk) 1.8.0_111 (like @paolopavan ). Eclipse's (4.6.1) own compiler doesn't complain at all and can compile the test and run it without problems.

sbliven:
Ok, so it seems like even a wildcard helper isn't sufficient here if you want to accept `Sequence<?>` objects, and it probably legitimately shouldn't compile. That said, I think that parameterizing the Blast*Parser classes is probably the correct solution, and it mirrors the FastaParser approach. 

Rather than using a lambda function for constructing the sequence, should I use a SequenceCreaterInterface instead?

paolopavan:
Dear all,
I am working at this solution: I have increased general type checking and
made data structures invariants with the purpose of enforcing the safety of
the framework. I mean that in this way the type checking can be done only
in the actual parser.
I am trying to avoid to parametrize the Hit and Result class because they
don't deal with any sequence. The purpose is to safely deal with Spencer
doubt when he mentioned SPI in the main SearchIO class. In my opinion this
can be safely done because at the parser side the HSP collection is typed
during its construction and then it is plugged as it is in the Hit object.
Since the data structure is invariant I think that this should be enough to
enforce the whole system safety. As a second enforcement level I have also
package restricted the access to the BlastHsp, BlastHit and BlastResult
classes (and their own Builders) so that no other class except the parser
may construct these data structures and may become the only delegated
entity to correctly assemble the whole.

2016-12-07 16:59 GMT+01:00 Spencer Bliven <notifications@github.com>:

sbliven:
Closing due to lack of consensus on how to best solve #354.

PR #611, opened 2016-11-29
fix for broken Phosphosite parser

andreasprlic:
Also contains a few javadoc improvements.

### DIFF
--- a/biojava-modfinder/src/main/java/org/biojava/nbio/phosphosite/Site.java
+++ b/biojava-modfinder/src/main/java/org/biojava/nbio/phosphosite/Site.java
@@ -65,7 +65,8 @@ public class Site {
 
 
 		while ((line = buf.readLine()) != null){
-			if ( line.startsWith("GENE")) {
+			if ( line.startsWith("GENE") ||
+					line.startsWith("PROTEIN")) {
 
 				headerFields = parseHeaderFields(line);
 
--- a/biojava-modfinder/src/main/java/org/biojava/nbio/phosphosite/Site.java
+++ b/biojava-modfinder/src/main/java/org/biojava/nbio/phosphosite/Site.java
@@ -56,25 +56,25 @@ public class Site {
 
 		int proteinIndex = -1;
 		int uniprotIndex = -1;
-		int modTypeIndex = -1;
 		int residueIndex = -1;
 		int orgIndex     = -1;
 		int groupIndex   = -1;
+		int geneIndex    = -1;
 
 		boolean inHeader = true;
 
 
 		while ((line = buf.readLine()) != null){
-			if ( line.startsWith("PROTEIN")) {
+			if ( line.startsWith("GENE")) {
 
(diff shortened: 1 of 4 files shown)


biojava-modfinder/src/main/java/org/biojava/nbio/phosphosite/Site.java
```diff
@@ -56,25 +56,25 @@ public Site(){
...
 		int orgIndex     = -1;
 		int groupIndex   = -1;
+		int geneIndex    = -1;
 
 		boolean inHeader = true;
 
 
 		while ((line = buf.readLine()) != null){
-			if ( line.startsWith("PROTEIN")) {
+			if ( line.startsWith("GENE")) {
```
sbliven: Would it be worth accepting both the old and new keywords, for backwards compatibility?

andreasprlic: Easy enough.

andreasprlic: About your other comment: I don't think there are user generated phosphosite files. The file format is very simple and it is only used for distribution as far as I know.

sbliven: In that case we don't need to worry about backwards compatibility

sbliven: Also, I intended this comment to apply to the other headers that changed as well

biojava-modfinder/src/test/java/org/biojava/nbio/protmod/phosphosite/TestAcetylation.java
```diff
@@ -0,0 +1,95 @@
...
+            for (Site s : sites) {
+
+                assertTrue(s.getResidue() != null);
+
+            }
+
+
+        } catch (Exception e) {
+            e.printStackTrace();
+            fail(e.getMessage());
```
josemduarte: It would be better to throw the exception here. Like this an exception will be reported as a test failure by junit.

Junit reports either errors or failures, and both are different: failures indicate that an assertion did not work, errors indicate that some unexpected exception occurred (see http://stackoverflow.com/questions/3425995/whats-the-difference-between-failure-and-error-in-junit)

biojava-modfinder/src/main/java/org/biojava/nbio/phosphosite/Site.java
andreasprlic: The other change was that they dropped a column (modification type) in the latest version. I think it can be inferred from the annotation in the "MOD_RSD" column (which the latest version of the parser is doing).

andreasprlic:
@sbliven please re-review so this can get merged. Thanks!

sbliven:
Sorry, I didn't realize I was blocking here. I just wanted to confirm that we can still parse the old files.

Issue #610, opened 2016-11-29
Phosphosite file format changed


andreasprlic:
Hi,

the latest file format of Phosphositeplus.org has changed and the parser in BioJava can not parse the current version of the files.

Andreas

lafita:
Was this entirely fixed by #612? Should we close?

josemduarte:
Indeed, it looks like it was solved already. I'll close it, @andreasprlic if you think it wasn't solved please reopen.

PR #612, opened 2016-11-29
fix for #610

andreasprlic:


sbliven:
Same comments as #611

josemduarte:
Let's make a new release once this is merged. There's a few commits in the bugfixes branch already.

PR #615, opened 2016-12-04
Fixed bug in parameter input

cristinamartins:
For a course at university (ESOF) we were supposed to find a bug in your project and solve it.
In the Alignment GUI (pairwise structure alignment), when choosing the parameters for the alignment algorithm, if the user writes letters instead of numbers in any of the parameters, after pressing “Apply”, the program retrieves a number format exception. This happens because the setValue method has an if cycle to convert the input string (value) to the variable type that the corresponding parameter is supposed to have (type). However, it is not prepared to deal with an invalid output, so we added a try-catch statement to the parsing and now in the console it only appears the message saying "Could not set value "value" for field "field"".

### DIFF
--- a/biojava-structure-gui/src/main/java/org/biojava/nbio/structure/align/gui/ParameterGUI.java
+++ b/biojava-structure-gui/src/main/java/org/biojava/nbio/structure/align/gui/ParameterGUI.java
@@ -285,15 +285,45 @@ public class ParameterGUI extends JFrame{
 			Object data = null;
 
 			if ( type == Integer.class){
-				data = Integer.parseInt(value);
+				try  
+				  {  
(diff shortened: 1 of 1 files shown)


andreasprlic:
Hi,

It is great that you need to find and fix bugs as part of your course! I wish more universities would do that. However, are you also discussing your solutions with your teacher? The problem with your patch is that catching an exception and not doing anything with it is considered to be bad programming behavior.

Please take a look at the chapter for _Exceptions_ in the book `Effective Java`, by Joshua Bloch. In general this is a great book to read for any Java developer and I highly recommend it.

A better fix for the described behavior would be not to allow the user to enter non-numerical values in the UI.

Hope that makes sense,

Andreas

PR #618, opened 2016-12-09
Fix354 paolo

paolopavan:
#354 #607 
Dear all,
I am working at this solution: I have increased general type checking and made data structures invariants with the purpose of enforcing the safety of the framework. I mean that in this way the type checking can be done only in the actual parser.
I am trying to avoid to parametrize the Hit and Result class because they don't deal with any sequence. The purpose is to safely deal with @sbliven doubt when he mentioned SPI in the main SearchIO class. In my opinion this can be safely done because at the parser side the HSP collection is typed during its construction and then it is plugged as it is in the Hit object. Since the data structure is invariant I think that this should be enough to enforce the whole system safety. As a second enforcement level I have also package restricted the access to the BlastHsp, BlastHit and BlastResult classes (and their own Builders) so that no other class except the parser may construct these data structures and may become the only delegated entity to correctly assemble the whole.

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/search/io/Hsp.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/search/io/Hsp.java
@@ -32,6 +32,7 @@ import org.biojava.nbio.core.sequence.ProteinSequence;
 import org.biojava.nbio.core.sequence.RNASequence;
 import org.biojava.nbio.core.sequence.compound.AminoAcidCompoundSet;
 import org.biojava.nbio.core.sequence.compound.DNACompoundSet;
+import org.biojava.nbio.core.sequence.compound.RNACompoundSet;
 import org.biojava.nbio.core.sequence.template.Compound;
 import org.biojava.nbio.core.sequence.template.Sequence;
 import org.slf4j.Logger;
@@ -137,7 +138,7 @@ public abstract class Hsp <S extends Sequence<C>, C extends Compound> {
             if (sequence instanceof DNASequence) 
                 returnSeq = new DNASequence(sequenceString, DNACompoundSet.getDNACompoundSet());
             else if (sequence instanceof RNASequence)
-                returnSeq = new RNASequence(sequenceString, DNACompoundSet.getDNACompoundSet());
+                returnSeq = new RNASequence(sequenceString, RNACompoundSet.getRNACompoundSet());
             else if (sequence instanceof ProteinSequence)
                 returnSeq = new ProteinSequence(sequenceString, AminoAcidCompoundSet.getAminoAcidCompoundSet());
             else if (sequence == null){
--- a/biojava-core/src/main/java/org/biojava/nbio/core/search/io/Hsp.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/search/io/Hsp.java
@@ -71,6 +71,8 @@ public abstract class Hsp <S extends Sequence<C>, C extends Compound> {
     private Double percentageIdentity = null;
     private Integer mismatchCount = null;
     private SimpleSequencePair<S, C> returnAln;
+    private S sequence;
+    private C compound;
 
     @Override
     public int hashCode() {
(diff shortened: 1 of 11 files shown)


biojava-core/src/main/java/org/biojava/nbio/core/search/io/Hsp.java
```diff
@@ -125,19 +129,24 @@ public boolean equals(Object obj) {
         return returnAln;
```
sbliven: The above assignments to alignedQuery/Hit need `<>` added so the types get inferred. This works if getSequence returns a `S` as it should.

biojava-core/src/main/java/org/biojava/nbio/core/search/io/Hsp.java
```diff
@@ -125,19 +129,24 @@ public boolean equals(Object obj) {
...
+    private Sequence<C> getSequence(String gappedSequenceString){
         if (gappedSequenceString == null) return null;
         
-        Sequence returnSeq = null;
+        //Sequence<C> returnSeq = null;
+        S returnSeq = null;
         String sequenceString = gappedSequenceString.replace("-", "");
         
         try {
+            // these casting are unavoidable risky operations due to the guess made on sequence string.
```
sbliven: The operation of converting a character into a particular compound set is risky. However, this is a different sort of risk; the risk that `S` won't match the actual sequence we instantiate. You can think of it as already "knowing" what S is before the method is called, so instantiating like this is taking an unavoidable risk and adding a second run-time risk.

Anyways, I wouldn't even care about this particular function if it were the only place with unchecked casting. However, it doesn't even serve its purpose of making the rest of the package type-safe.

biojava-core/src/main/java/org/biojava/nbio/core/search/io/Result.java
```diff
@@ -18,163 +18,163 @@
...
-            }
-
-            @Override
-            public void remove() {
-                throw new UnsupportedOperationException("The remove operation is not supported by this iterator");
-            }
-        };
-    }
-}
+package org.biojava.nbio.core.search.io;
```
sbliven: Yay windows carriage returns for messing up diffs!

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastHitBuilder.java
```diff
@@ -33,49 +34,49 @@
...
-public class BlastHitBuilder {
+public class BlastHitBuilder<S extends Sequence<C>, C extends Compound> {
     private int hitNum;
     private String hitId;
     private String hitDef;
     private String hitAccession;
     private int hitLen;
-    private Sequence hitSequence;
-    private List<Hsp> hsps;
+    private Sequence<?> hitSequence;
```
sbliven: If you're going to make BlastHitBuilder parametric then you better use those parameters here. Should be an `S`.

biojava-core/src/main/java/org/biojava/nbio/core/search/io/ResultFactory.java
```diff
@@ -18,59 +18,60 @@
...
+/**
+ * Designed by Paolo Pavan.
+ * You may want to find my contacts on Github and LinkedIn for code info 
+ * or discuss major changes.
+ * https://github.com/paolopavan
+ * 
+ * @author Paolo Pavan
+ */
+
+public interface ResultFactory <S extends Sequence<C>, C extends Compound>{
```
sbliven: It is inconsistent to have `ResultFactory` parameterized but not `Result`

biojava-core/src/main/java/org/biojava/nbio/core/search/io/blast/BlastHitBuilder.java
```diff
@@ -33,49 +34,49 @@
  * 
  * @author Paolo Pavan
  */
-public class BlastHitBuilder {
+public class BlastHitBuilder<S extends Sequence<C>, C extends Compound> {
```
sbliven: Again, inconsistent to parameterize HitBuilder but not Hit

sbliven:
Closing due to lack of consensus on how to best solve #354.

PR #619, opened 2016-12-17
Add Zoom and Spin Buttons in the Structure-Gui

SofiaSilveira:
For a course at university (ESOF) we were supposed to try to implement a new feature to this project.
Clicking with the right mouse button on the Structure-Gui panel, a menu is shown with properties and options: zooming, spinning, capturing as an image, recording a GIF, etc. However, we think that the using of the right click is not intuitive at all since there are options that, as previously said, can be set through controls that are already visible when the user interface is opened. This can induce users to think that there are no more available options. To avoid this problem we think that it would be easier if some of the most important options were available in the main window of the GUI.

Considering that zooming is one of the most important features to be able to visualize a protein in a better way, implementing a control to set the zoom is one of the features that we decided to add to this GUI. With this button the user will have the opportunity to zoom in and out as many times as he wants inside a predefined range (0% until 500%). Additionally, spinning the protein also allows a better visualization of the structure, so we decided to add the possibility to start and stop the spinning of the protein.

### DIFF
--- a/biojava-structure-gui/src/main/java/org/biojava/nbio/structure/align/gui/jmol/StructureAlignmentJmol.java
+++ b/biojava-structure-gui/src/main/java/org/biojava/nbio/structure/align/gui/jmol/StructureAlignmentJmol.java
@@ -38,31 +38,34 @@ import org.biojava.nbio.structure.gui.util.color.ColorUtils;
 import org.biojava.nbio.structure.jama.Matrix;
 
 import javax.swing.*;
+import javax.swing.event.ChangeEvent;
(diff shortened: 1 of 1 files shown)


andreasprlic:
Hi Sofia,

thanks for the pull request. It looks good. In the future would be great to get pull requests for some of our open tickets! For example here a few possible projects for students:

https://github.com/biojava/biojava/issues?q=is%3Aissue+is%3Aopen+label%3A%22Ideas+for+Student+Projects%22

Thanks,

Andreas

PR #630, opened 2017-01-13
Fix #552 maximum RMSD parameter to SmithWaterman3Daligner superpositions

### LINKED ISSUE #552: Maximum RMSD Parameter for SmithWaterman3D Aligner
I think that a useful feature for the structural alignment algorithm `SmithWaterman3Daligner` would be to take a maximum **RMSD** as parameter and drop alignment columns iteratively (the column with highest distance each step) until the the **RMSD** of the alignment is lower than the specified.

This is similar to what **PyMol** does with the [align command](http://pymol.org/dokuwiki/doku.php?id=command:align).

I have been looking at the `SmithWaterman3Daligner` class and found this comment:

> version history:
>  1.1 - Added more parameters to the command line, including -maxOptRMSD
>  1.0 - Initial version

The parameter **-maxOptRMSD** sounds like what I am describing here, but I could not find any such parameter in the `SmithWaterman3DParameters` class, nor in the rest of the algorithm code. 

Does anyone know if this feature was implemented?

lafita:
Implement the feature requested in #552. The sequence alignment from SmithWaterman is reduced until one of the two, minimum length or maximum RMSD, is fulfilled, in order to produce higher quality superposition outcomes.


lafita:
I created an integration test for the `SmithWaterman3Daligner`, so this can be finally merged!

PR #634, opened 2017-02-15
Fix Quaternary symmetry symmetry detection bug

sbliven:
In special cases it was possible to have pairs of axes that
pass our quality thresholds but form invalid symmetry groups.

Now alignments that superimpose a subunit onto itself are forbidden,
which should restrict it to the chiral groups.

Also add numerous notes and documentation.


josemduarte:
Nice thanks! On time for the next bugfix release. I will try to do the release before next week. I'll also merge this one onto master.

PR #636, opened 2017-02-23
Bugfix: chromosome mapping tool

valasatava:
- Fix in the chromosome mapping tool to return -1 if the genomic position falls outside coding region on the forward strand
- Test is added to check if the boundaries are handled correctly

### DIFF
--- a/biojava-genome/src/test/java/org/biojava/nbio/genome/TestGenomeMapping.java
+++ b/biojava-genome/src/test/java/org/biojava/nbio/genome/TestGenomeMapping.java
@@ -10,6 +10,7 @@ import org.junit.Test;
 
 import java.io.InputStream;
 import java.net.URL;
+import java.util.Arrays;
 import java.util.List;
 import java.util.zip.GZIPInputStream;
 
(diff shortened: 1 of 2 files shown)


biojava-genome/src/test/java/org/biojava/nbio/genome/TestGenomeMapping.java
```diff
@@ -148,4 +145,34 @@ private void validateExon(int exonNr, int start, int stop, List<Range<Integer>>
...
+    			continue;
+    		if ( !gcp.getGenebankId().equals(genebankId) )
+    			continue;
+    		
+    		int cdsSE = ChromosomeMappingTools.getCDSPosForChromosomeCoordinate(posExonStart-1, gcp);
+    		assertEquals(cdsSE, -1);
+    		
+    		int cdsEE = ChromosomeMappingTools.getCDSPosForChromosomeCoordinate(posExonEnd+1, gcp);
+    		assertEquals(cdsEE, -1);
+
```
andreasprlic: Can you also add a check that the CDS position for first base in the coding region is correct?

PR #640, opened 2017-03-02
Bugfix in hmmer web service

josemduarte:
The header was using "Accept:" when it should be "Accept".


andreasprlic:
shouldn't this go also in the main branch?

josemduarte:
I'll merge it to master too, thanks!

PR #642, opened 2017-03-21
Extracting protein sequence for genetic positions

valasatava:
The class that helps to extract protein sequence for genetic positions


biojava-genome/src/main/java/org/biojava/nbio/genome/util/ChromosomeToProteinMapper.java
```diff
@@ -0,0 +1,193 @@
...
+public class ChromosomeToProteinMapper {
+	
+	private TwoBitParser parser;
+	private String GENOME_URI;
+	
+	/** Sets a path to the genome.2bit file.
+	 * 
+	 *  @param path The path to the file containing the genome in .2bit format.
+	 */
+	public void setGenomeURI(String path) {
```
andreasprlic: can you make the input a URI (or URL) rather than a String?

andreasprlic: actually, since the TwoBit Parser requires a File object as an input, can you make this accept a File object, that points to a .2bit file?

biojava-genome/src/main/java/org/biojava/nbio/genome/util/ChromosomeToProteinMapper.java
```diff
@@ -0,0 +1,193 @@
...
+	
+	/** Sets a chromosome for TwoBitParser.
+	 * 
+	 * @param chr The chromosome name (e.g. chr21)
+	 */
+	public void setChromosome(String chr) throws Exception {
+		parser.close();
+		String[] names = parser.getSequenceNames();
+		for(int i=0;i<names.length;i++) {
+			if ( names[i].equals(chr) ) {
```
andreasprlic: what about equalsIngoreCase ? to make sure chr21 is the same as Chr21 (just in case)

biojava-genome/src/main/java/org/biojava/nbio/genome/util/ChromosomeToProteinMapper.java
```diff
@@ -0,0 +1,193 @@
...
+		return transcription;
+	}
+	
+    /** Converts the DNA sequence to protein sequence.
+    *
+    * @param sequence the DNA sequence
+    * 
+    * @return the protein sequence
+    */
+	public String convertDNAtoProteinSequence(String sequence) throws CompoundNotFoundException {
```
andreasprlic: name the argument dnaSequence to make it clear in javadoc that the input should be a DNA sequence.

biojava-genome/src/main/java/org/biojava/nbio/genome/util/ChromosomeToProteinMapper.java
```diff
@@ -0,0 +1,193 @@
...
+    /** Extracts the exons boundaries in CDS coordinates for genes living on the reverse DNA strand.
+     *
+     * @param exonStarts The list holding the genetic coordinates pointing to the start positions of the exons (including UTR regions)  
+     * @param exonEnds The list holding the genetic coordinates pointing to the end positions of the exons (including UTR regions)
+     * @param cdsStart The start position of a coding region
+     * @param cdsEnd The end position of a coding region
+     * 
+     * @return the list of genetic positions corresponding to the exons boundaries in CDS coordinates
+    */
+    public static List<Range<Integer>> getCDSRegionsReverse(List<Integer> exonStarts, List<Integer> exonEnds,
```
andreasprlic: can you move these methods to the ChromosomeMapping tools? There are already a set of utility methods like that there.

biojava-protein-disorder/pom.xml
```diff
@@ -3,7 +3,7 @@
 	<parent>
 		<artifactId>biojava</artifactId>
 		<groupId>org.biojava</groupId>
-		<version>5.0.0-SNAPSHOT</version>
+		<version>5.0.0-YV-SNAPSHOT</version>
```
josemduarte: No need to change this

andreasprlic:
This is turning into a mix between minor bug fixes and new features. Can we add a unit test for the bug fix and also for making sure that protein sequences get extracted correctly?
Thanks

biojava-aa-prop/src/main/java/org/biojava/nbio/aaproperties/PeptideProperties.java
```diff
@@ -564,7 +564,7 @@ public static final double getEnrichment(String sequence, String aminoAcidCode){
 	 * 		a protein sequence consisting of non-ambiguous characters only
 	 * @return the array of charges of amino acids in the protein (1 if amino acid is positively charged, -1 if negatively charged, 0 if not charged)
 	 */
-	public static final int[] getChargesOfAminoAcidsInProtein(String sequence) {
+	public static final int[] getChargesOfAminoAcids(String sequence) {
```
andreasprlic: if you want to enforce this to be protein-sequences can you make the arguments a BioJava ProteinSequence object?

valasatava: It will not be consistent with other methods of the class that work with sequence input as String

andreasprlic: Good point. We should change the interface for this class. However that's beyond the scope of this PR, so don't do this now.

PR #645, opened 2017-03-28
Adding equals and hashcode methods to Sequences

valasatava:
Sequences in biojava-core are currently missing an equals method. This pull request fixes this.

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/alignment/SimpleAlignedSequence.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/alignment/SimpleAlignedSequence.java
@@ -270,7 +270,7 @@ public class SimpleAlignedSequence<S extends Sequence<C>, C extends Compound> im
 		return compounds;
 	}
 
-
+	@Override
 	public boolean equals(Object o){
 
 		if(! Equals.classEqual(this, o)) {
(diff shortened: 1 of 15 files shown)


biojava-core/src/main/java/org/biojava/nbio/core/sequence/template/Sequence.java
```diff
@@ -79,6 +79,10 @@
 	 */
 	public List<C> getAsList();
 
+	public boolean equals(Sequence<C> other);
+
+	public int hashCode();
+
```
josemduarte: No need to declare these 2 here. They are inherited from `Object`

biojava-core/src/main/java/org/biojava/nbio/core/alignment/SimpleAlignedSequence.java
```diff
@@ -270,6 +270,19 @@ public AccessionID getAccession() {
 	}
 
 	@Override
+	public boolean equals(Sequence<C>other){
```
andreasprlic: actually we should have written this as

equals(Object) 

instead of equals(Sequence<C>)

biojava-core/src/main/java/org/biojava/nbio/core/sequence/loader/StringProxySequenceReader.java
```diff
@@ -160,4 +160,28 @@ public int countCompounds(C... compounds) {
 	public SequenceView<C> getInverse() {
 		return SequenceMixin.inverse(this);
 	}
+
+	@Override
+	public boolean equals(Sequence<C> other){
```
andreasprlic: same here

PR #646, opened 2017-03-28
Upgrading to mmtf 1.0.5

josemduarte:
Should bring faster parsing for mmtf, thanks @altaite


andreasprlic:
Just to comment, this is a significant performance improvement: PDB ID 4CUP via mmtf used to parse in ~ 900ms. Now we are down to <200ms.  Almost twice as fast as parsing from PDB file format!

andreasprlic:
actually, I realized the performance test included IO. I'll do another comparison that will exclude any contributions from IO in parsing.

altaite:
It will overestimate PDB, but just slightly, so its fine. IO is usually not
the bottleneck. Opening lot of small MMTF files can be, it takes 50 % of
total time. Opening, not reading, I am not sure why.

andreasprlic:
just a comment on performance. For a mid-size PDB entry like 3HBX on my laptop mmtf parses in 50 ms vs pdb in ~100 ms.  this is from ram (cutting out any influence from slow drives)

andreasprlic:
... and that was not using the latest code (thanks due to IDE caching). Forcing everything to the latest 1.0.5 of mmtf, which does not use Jackson any longer, the latest numbers are:

| PDB ID | method | average load time (100 repetitions) | size of structure |
|--------| --------|------------------------------------| ------------------|
| 4CUP | mmtf | 5 ms.   | small |
| 4CUP | pdb   | 20 ms. | small |
| 3HBX | mmtf | 14 ms. | mid size |
| 3HBX | pdb | 48 ms. | mid size |

PR #647, opened 2017-03-29
more precise comparison of parsing speed between mmtf and pdb formats 

andreasprlic:


andreasprlic:
Based on mmtf java version 1.0.5, which does not use Jackson any longer, the latest performance comparison numbers are:

| PDB ID | method | average load time (100 repetitions) | size of structure |
|--------| --------|------------------------------------| ------------------|
| 4CUP | mmtf | 5 ms.   | small |
| 4CUP | pdb   | 20 ms. | small |
| 3HBX | mmtf | 14 ms. | mid size |
| 3HBX | pdb | 48 ms. | mid size |

PR #654, opened 2017-04-05
Get CDS ranges patch

valasatava:
Bug fix to get CDS ranges without changing the original data


andreasprlic:
I see that you merged in the JRONN related changes as well. Next time better to use feature branches rather than master :-)

PR #660, opened 2017-04-27
Add methods for reading RNA sequences.

heuermh:
Added methods for reading RNA sequences by copy paste search replace.

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/FastaReaderHelper.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/FastaReaderHelper.java
@@ -23,10 +23,12 @@ package org.biojava.nbio.core.sequence.io;
 
 import org.biojava.nbio.core.sequence.DNASequence;
 import org.biojava.nbio.core.sequence.ProteinSequence;
+import org.biojava.nbio.core.sequence.RNASequence;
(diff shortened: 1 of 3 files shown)


josemduarte:
There are some issues with the tests, most likely unrelated to this. I'll try restarting them again.

lafita:
I think the errors are with CATH tests (a NPE). I will run the tests locally to see if they pass.

Also, could we add some tests for the new methods? Just copying the DNA tests and replacing them with RNA would be enough.

lafita:
Tests pass locally and I see that also in travis. Looks ok to me although would be nice to include the tests.

heuermh:
Sure, I can add some tests.  I was hunting for some Genbank RNA files in the wild but I couldn't find any, they aren't usually represented that way.

lafita:
Do you mean an RNA database? I know about RNAcentral for non-coding RNA sequences http://rnacentral.org

heuermh:
Specifically I meant Genbank sequences using the RNA alphabet.  You can find FASTA formatted RNA sequences in the RNA alphabet in some of those databases, but Genbank RNA sequences (like AY892647.1 you linked) are represented in the DNA alphabet.

lafita:
True, I wonder why though! Seems that a simple U for T replacement is needed.

josemduarte:
So it seems that we can't have an RNA file for the test, right? I'll merge for the moment, but it'd be good to have a test at some point.

Issue #663, opened 2017-05-30
GenbankReader#process(int max) closes the underlying InputStream


psoares:
I tried to use the GenbankReader to read a Genbank file. I wanted to do this through repeated calls to the version of the process method that takes a number of records as the only parameter.
After the first call, the underlying inputstream is closed and the second call fails.

Here is an example :

```java
import org.biojava.nbio.core.sequence.DNASequence;
import org.biojava.nbio.core.sequence.compound.DNACompoundSet;
import org.biojava.nbio.core.sequence.compound.NucleotideCompound;
import org.biojava.nbio.core.sequence.io.FileProxyDNASequenceCreator;
import org.biojava.nbio.core.sequence.io.GenbankReader;
import org.biojava.nbio.core.sequence.io.GenbankSequenceParser;
import org.biojava.nbio.core.sequence.io.GenericGenbankHeaderParser;
import org.biojava.nbio.core.sequence.template.AbstractSequence;

import java.io.File;
import java.io.FileInputStream;
import java.io.InputStream;
import java.util.LinkedHashMap;

class TestGBReading {

    public static void main(String[] args) {
        File file = new File("/tmp/gbpat100.seq");

        LinkedHashMap<String, DNASequence> sequences = null;
        GenbankReader<DNASequence, NucleotideCompound> gbReader = null;
        try {
            gbReader =
                    new GenbankReader<DNASequence, NucleotideCompound>(
                            file,
                            new GenericGenbankHeaderParser<DNASequence, NucleotideCompound>(),
                            new FileProxyDNASequenceCreator(
                                    file,
                                    DNACompoundSet.getDNACompoundSet(),
                                    new GenbankSequenceParser<AbstractSequence<NucleotideCompound>,
                                            NucleotideCompound>()
                            )
                    );
            sequences = gbReader.process(10);
            System.out.println("Reading first 10");
            for (DNASequence seq : sequences.values()) {
                System.out.println(" - " + seq.getAccession());
            }

            sequences = gbReader.process(10);
            System.out.println("Reading next 10");
            for (DNASequence seq : sequences.values()) {
                System.out.println(" - " + seq.getAccession());
            }

        } catch (Exception e) {
            e.printStackTrace();
        }

    }
}```

josemduarte:
This looks like a bug, I'll submit a pull request.

lafita:
Is this one fixed by PR #666? Should we close the issue?

psoares:
I think so...

PR #665, opened 2017-05-31
Fixing issue #663

josemduarte:


josemduarte:
Better solution in #666

PR #667, opened 2017-06-02
Add list of comments into AbstractSequence and GenericGenbankHeaderParser

jimty0511:
Add List<comments> field with getter and setter in AbstractSequence.java;
Add setComments section into parseHeader() in GenericGenbankHeaderParser.java;
Update GenbankReaderTest with the comments checking

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericGenbankHeaderParser.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericGenbankHeaderParser.java
@@ -30,134 +30,103 @@ import org.biojava.nbio.core.sequence.template.AbstractSequence;
 import org.biojava.nbio.core.sequence.template.Compound;
 
 import java.util.ArrayList;
-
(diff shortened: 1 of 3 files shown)


josemduarte:
Thanks for the patch! 

The diff didn't come up very cleanly. It's difficult to see what actually changed. Did you do any indentation reformatting? Or are you perhaps in a Windows system that changes the end of lines?

There's also something weird in the list of commits showing up a bunch of old commits. It would also be nice if you could rebase on latest master and pull request based on that.

jimty0511:
Hi,

Thanks for your email.

Here are the things I have changed:

1. I added a List<String> comments in the AbstractSequence.java with getter and setter.

2. For the GenericGenbankHeaderParser.java, I generated all the getters for all the fields. And added sequence.setComments(comments) in the parseHeader() method.


So after the GenbankReader.process() method, every sequence contains a list of String of comments.

And for the wired thing is that in my branch, the FileProxyRNASequenceCreator class was missing, and I added it back after I merged with the biojava/biojava branch.

Jim

PR #669, opened 2017-06-07
Add Reference tag parsing process for Genbank file.

jimty0511:
1. Added a new folder into main/java/org/biojava/nbio/core/sequence package
for Reference tag. It contains ReferenceInterface, AbstractReference and
GenbankReference (Only for Genbank format files).
2. Added List of AbstractReference with getter and setter into AbstractSequence.java.
3. Updated GenbankSequenceParser to parse the Reference tag.
4. Implemented addReference method in GenericGenbankHeaderParser because
some Genbank format file may contain multiple Reference tag.
5. Updated GenbankReaderTest with to check the Reference tag parsing process.

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/sequence/reference/GenbankReference.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/sequence/reference/GenbankReference.java
@@ -1,0 +1,66 @@
+package org.biojava.nbio.core.sequence.reference;
+
+/**
+ * For Genbank format file only.
+ *
+ * @Author Jim Tang
+ */
+public class GenbankReference extends AbstractReference {
+
+    /**
+     * The authors are a list of Inventors that retrieved from the Reference section.
+     */
+    private String authors;
+
+    /**
+     * The title that retrieved from the Reference section.
+     */
+    private String title;
+
+    /**
+     * The journal usually contains the Publication Number, Publication Date and Assignee
+     */
+    private String journal;
+
+    /**
+     * @return
+     */
+    public String getAuthors() {
+        return authors;
+    }
+
+    /**
+     * @param authors
+     */
+    public void setAuthors(String authors) {
+        this.authors = authors;
+    }
+
+    /**
+     * @return
+     */
+    public String getTitle() {
+        return title;
+    }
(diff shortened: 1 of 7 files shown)


biojava-core/src/main/java/org/biojava/nbio/core/sequence/reference/GenbankReference.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * The journal usually contains the Publication Number, Publication Date and Assignee
+     */
+    private String journal;
+
+    /**
+     * @return
+     */
+    public String getAuthors() {
```
josemduarte: Here we should have \@Override instead of javadocs. The docs are in the interface.

biojava-core/src/main/java/org/biojava/nbio/core/sequence/template/AbstractSequence.java
```diff
@@ -323,6 +325,22 @@ public void setSequenceScore(Double sequenceScore) {
 	}
 
 	/**
+	 *
+	 * @return
+	 */
+	public List<AbstractReference> getReferences() {
```
josemduarte: Can you add javadocs here?

jimty0511: Hi, just committed the new changes. Do I need to close this request and create a new one? Thanks.

josemduarte: Just keep committing to the same branch. The Pull Request will automatically update with those commits.

jimty0511: Is there any other change needed?

heuermh:
`@since 5.0.0` doc tags would be nice to have, and since commit d7682bf2936dd22dddecf181d191803c4fc876b2 failed to build, the commits in this pull request should be squashed.

jimty0511:
I think the reason why the d7682bf failed is I did a committed before the build is finished. And the latest build db24180 is passed.

PR #688, opened 2017-07-13
Fixing tests after PDB v5 remediation

josemduarte:


lafita:
Did you mean mmCif v5?

PR #691, opened 2017-07-21
Base Pair Step Parameters Version 1

lukeczapla:
This is my PR for the issue #689  

The implementing class is BasePairParameters.java inside the secstruc package.  This seems to be sort of a secondary structure issue for nucleic acids.  The nucleic acid people agreed on this convention and it was originally proposed in 1995 by el Hassan and Calladine.  The citation to the standardized one that I implement is: https://www.ncbi.nlm.nih.gov/pubmed/11601858

There is a test class in the tests folder, too.  It just downloads 1KX5 from RCSB (the nucleosome core particle bound to 147 bp DNA) and prints out the base pair and base-pair step parameters using the new class.  This was validated against another program that does the similar calculation, 3DNA.  The BioJava version would be more accessible for people doing research in Java, since it's a package that can go into your more complicated analysis, and not a standalone program.   There are no new dependencies.

[Also, I made the EcodInstallation no-arg constructor public just for consistency with all the other classes like Cath and SCOP.  I have been told it's better to use the EcodFactory and so I switched to that, but it makes sense that if the other EcodInstallation constructors are public, that the no-arg one might as well be public too.]


biojava-structure/src/main/java/org/biojava/nbio/structure/secstruc/BasePairParameters.java
```diff
@@ -0,0 +1,403 @@
...
+
+    private Structure structure;
+    private String pairSequence = "", pdbId = "";
+
+    private double[] pairParameters;
+    private double[][] pairingParameters;
+    private double[][] stepParameters;
+
+
+    // Either pass the name of a PDB file (ending in .pdb) or pass the pdbId
```
lafita: Use javadoc

biojava-structure/src/test/java/org/biojava/nbio/structure/secstruc/TestBasePairParameters.java
```diff
@@ -0,0 +1,31 @@
+import org.biojava.nbio.structure.secstruc.BasePairParameters;
+import org.junit.Test;
+
+/**
+ * Created by luke on 7/21/17.
+ */
+public class TestBasePairParameters {
```
lukeczapla: That had already been done with the commit "Made test use no println, and class use Logger"

I will rewrite it to use the Structure object alone (it would be the user's responsibility whether to use StructureIO or to use getStructure(filename) on a local file) and add the JavaDocs and other suggestions.  Thank you!  

--

lafita: Thanks! I had not seen that commit when I reviewed, sorry.

biojava-structure/src/main/java/org/biojava/nbio/structure/secstruc/BasePairParameters.java
```diff
@@ -0,0 +1,403 @@
...
+/**
+ * Contributed to BioJava under its LGPL
+ * This module calculates the el Hassan-Calladine Base Pairing and Base-pair Step Parameters
+ * Citation: https://www.ncbi.nlm.nih.gov/pubmed/11601858
+ *
+ * Created by luke on 7/20/17.
+ */
+public class BasePairParameters {
+
+    private static String[] standardBases = new String[] {
```
lukeczapla: Standard bases are obtained from http://ndbserver.rutgers.edu/ndbmodule/archives/reports/tsukuba/Table1.html  which was also a table in the paper I cite in the header (JMB 2001 Olson et. al.).  I thought it would be better to hard code it so no issues come up with finding resources.  Previously I had them in a local folder but if someone wants to analyze helical regions it makes sense that the agreed upon standard is already in the code.

biojava-structure/src/test/java/org/biojava/nbio/structure/secstruc/TestBasePairParameters.java
lafita: For the input of the method, I was wondering if it would be better to pass directly a `Nucleotide` Chain (or a `List` of `Nucleotide`), since the method only works with the nucleotide chains of the `Structure` object. It would also save you the `Nucleotide` chain extraction part of the code. This is only a suggestion, totally your choice!

biojava-structure/src/main/java/org/biojava/nbio/structure/secstruc/BasePairParameters.java
```diff
@@ -0,0 +1,403 @@
+package org.biojava.nbio.structure.secstruc;
```
lukeczapla: Ok, I am going to call the package basepairs, since it's obtaining both the pairing parameters (buckle, propeller, opening, shear, stretch, and stagger), and the base-pair step parameters (tilt, roll, twist, shift, slide, and rise).  Today I will work on JavaDoc for everything that the user might want to customize or override.

biojava-structure/src/main/java/org/biojava/nbio/structure/secstruc/BasePairParameters.java
```diff
@@ -0,0 +1,403 @@
...
+    private double[][] stepParameters;
+
+
+    // Either pass the name of a PDB file (ending in .pdb) or pass the pdbId
+    public BasePairParameters(String name) {
+        PDBFileReader pdbFileReader = new PDBFileReader();
+        pdbFileReader.setPath(".");
+        try {
+            if (name.contains(".pdb")) structure = pdbFileReader.getStructure(name);
+            else structure = StructureIO.getStructure(name);
```
lukeczapla: Agreed.  the new version is using Structure directly, but this implementation had supported both using StructureIO and loading a local file, that second part is why the PDBFileReader was there.

biojava-structure/src/test/java/org/biojava/nbio/structure/secstruc/TestBasePairParameters.java
lukeczapla: It's an interesting idea.  Right now it's mainly the removeDups flag that's relevant in that method.  Some X-ray structures contain nearly-identical units, and so removeDups intends to only count the first instance of a helix of interest.  

The usage of this code at the moment is to run through RCSB and extract parameters of unique sequence/protein combinations for about 4000 structures, for only X-ray structures. 
 There are some notable NMR structures that are very accurate but they haven't been considered yet, because many among them are not as reliable [in the opinion of some people... I think they're alright].  The speed is amazing with all your caching, so I can see the effect of changing criteria in a manner of a minute or two.

So I reworked the constructors so different criteria can be provided (whether its DNA or RNA and whether to removeDups, and just the generic version for DNA that takes a Structure), and at some point RNA/DNA hybrid helices might be considered. 

So I added commits to make the changes, but let me know if you think anything else could be optimized and commented.

lukeczapla:
The latest implement has dealt with the issues that were brought up, with these commits:

_Moved base pair code into new package basepairs
Test fix_

I will go on and make another similar class that just has a different implement of findPairs().  For people studying RNA tertiary structure, it makes sense to write some code that aligns part of the sequence of an RNA strand with some other part of the same RNA strand.  Often, RNA has unpaired looped regions and then base pairs with itself.  I am not particularly focused on RNA tertiary structure, but I will like to try writing that and seeing if it works to analyze these complex structures.

lukeczapla:
I added the new class (TertiaryBasePairParameters, extending the existing class) and validated it on a hairpin RNA (PDB id 3php).  It finds base pairs within a single strand.  I made it easier to get the list of all the pairs you have too, with a getPairingNames() method, and it would have "AT" if it's an A-T base pair.  So for record keeping there is a lot more useful information, the pairSequence is just the letter of the base 1 in the pair, but the spaces separate gaps so that analyses don't assume that a "step" is necessarily from one base pair to the one directly above it.  It's usually easy to tell anyways, but it is a sure way to be sure.

I don't know if you accept contributions for tutorials and such, but I could make a demo code and document its usage (basically it could be an "app" that returns the helical parameters just like my original Test class had written out of bunch of stuff).  Anything that I find that is needed in the analysis will be a good candidate for inclusion into the class.

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,576 @@
...
+        return this;
+    }
+
+
+    /**
+     * This reports all the pair parameters, in the order of:
+     * buckle, propeller, opening (in degrees), shear, stagger, stretch (in Å).
+     * @return A double[][] with length equal to number of base pairs for rows, and 6 columns
+     */
+    public double[][] getPairingParameters() {
```
lafita: In general, I would not recommend returning nested arrays directly, but rather create getter methods for each of the parameter arrays. This would be a much friendlier API for the user and it can avoid trivial bugs and nested loops to change the data. For example, we could have: `getBuckle(), getPropeller(), getOpening(),` etc.

You might keep the internal data structures and this getter method, but add additional getter methods for each.

biojava-structure/src/test/java/org/biojava/nbio/structure/basepairs/TestBasePairParameters.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+public class TestBasePairParameters {
+
+    @Test
+    public void testBasePair() {
+
+        Structure structure;
+        try {
+            structure = StructureIO.getStructure("1KX5");
+        } catch (IOException|StructureException e) {
```
lafita: Instead of catching the exceptions, add a `throws` declarations to the method, so that the `Test` status is `ERROR` instead of `FAIL` when the structure download gives an exception.

biojava-structure/src/test/java/org/biojava/nbio/structure/basepairs/TestBasePairParameters.java
```diff
@@ -0,0 +1,79 @@
...
+        assertEquals(steps[1][1], 0.785, 0.1);
+        assertEquals(steps[1][2], 32.522, 0.5);
+        // next three in Å, shift, slide, rise
+        assertEquals(steps[1][3], -0.873, 0.01);
+        assertEquals(steps[1][4], -0.607, 0.01);
+        assertEquals(steps[1][5], 3.070, 0.01);
+
+        try {
+            structure = StructureIO.getStructure("3PHP");
+        } catch (IOException|StructureException e) {
```
lafita: Same as above

biojava-structure/src/test/java/org/biojava/nbio/structure/basepairs/TestBasePairParameters.java
```diff
@@ -0,0 +1,79 @@
...
+            e.printStackTrace();
+            structure = null;
+            assertEquals(1, 2);
+        }
+        bp = new TertiaryBasePairParameters(structure, true, false).analyze();
+        assertEquals(9, bp.getPairingParameters().length);
```
lafita: Why don't you compare the parameters here or below? Do you assume they are correct from the test above?

lafita:
> I don't know if you accept contributions for tutorials and such, but I could make a demo code and document its usage (basically it could be an "app" that returns the helical parameters just like my original Test class had written out of bunch of stuff). Anything that I find that is needed in the analysis will be a good candidate for inclusion into the class.

@lukeczapla we would encourage any contribution to the BioJava tutorial (https://github.com/biojava/biojava-tutorial). I think it would be great if you can add a new chapter inside the `Book 3: The Structure Book`, maybe call it `Chapter XX - Nucleotide Base Pair Parameters`, or something like that. Let me know if you have trouble setting a new chapter, I can set it for you and then you can fill in the text.

In addition, what we usually do is to create a Demo class and put it into the demo package (src/main/java/demo). You can take a look at other demos for inspiration.

lukeczapla:
I just added all the full tests against 3DNA for the three classes, and the methodology is not a clone of that external program so there is some "epsilon" between values coming out of our method vs. that program 3DNA.  It's about as close as it can get, though.  The author of the JMB 2001 paper standardizing this system (Wilma Olson) has been shown the results and she was happy to see how good it is.

The getters you requested were added, with documentation about which are coming out in degrees and which are coming out in å.   Step parameters are for the step to the current base pair from the one before it, and pairing parameters are for the individual base pairs themselves.  There's a little check for an out-of-bounds value of which bp you want, for getBuckle(n) ..., getRise(n), etc., in which case it returns null.  But it could throw an Exception instead, if you think that's more appropriate.

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -235,6 +235,127 @@ public String getPairSequence() {
     }
 
     /**
+     * Return the buckle in degrees for the given base pair
+     * @param bp the number of the base pair (starting with 0)
+     * @return the value as a double (in degrees)
+     */
+    public Double getBuckle(int bp) {
+        if (bp < 0 || bp >= getPairingParameters().length) return null;
```
lafita: Better throw an exception (`IllegalArgumentException` or `IndexOutOfBoundsException`, as you prefer) with a message describing what happened. The same for all other getters.

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/MismatchedBasePairParameters.java
```diff
@@ -0,0 +1,85 @@
...
+import javax.vecmath.Matrix4d;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Contributed to BioJava under its LGPL
+ * This class allows for finding inter-strand base pairs that are not necessarily canonical Watson-Crick pairs.
+ * The implementation of findPair is different than that of the base class.
+ * Created by luke on 7/23/17.
+ */
```
josemduarte: Minor comment: could you use @author annotations (here an in other new classes that you introduced).

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/TertiaryBasePairParameters.java
```diff
@@ -0,0 +1,77 @@
...
+                    Atom a2 = g2.getAtom("C1'");
+                    if (a1 == null || a2 == null) continue;
+                    // C1'-C1' distance is one useful criteria
+                    if (Math.abs(a1.getCoordsAsPoint3d().distance(a2.getCoordsAsPoint3d())-10.0) > 5.0) continue;
+                    Group[] ga = new Group[] {g1, g2};
+                    Matrix4d data = basePairReferenceFrame(ga);
+                    // if the stagger is greater than 2 Å, it's not really paired.
+                    if (Math.abs(pairParameters[5]) > 2.0) continue;
+                    // if the propeller is ridiculous it's also not that good of a pair.
+                    if (Math.abs(pairParameters[1]) > 60.0) {
```
josemduarte: There's a few numerical constants in here that should be extracted into constants with javadocs in `BasePairParameters`. They are used both in here and above in the other `findPairs` implementation.

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/MismatchedBasePairParameters.java
```diff
@@ -0,0 +1,85 @@
...
+
+    }
+
+    /**
+     * This is an implementation for finding non-canonical base pairs when there may be missing or overhanging bases.
+     * @param chains The list of chains already found to be nucleic acids
+     * @return The list of the atom groups (residues) that are pairs, a Group[2] array
+     */
+    @Override
+    public List<Group[]> findPairs(List<Chain> chains) {
```
lukeczapla: Ok, I have done that!

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/TertiaryBasePairParameters.java
lukeczapla: I think that makes sense.  It is a judgment call for determining if two bases are "paired" or not.  Their stagger is a major factor.  So if these constants are put somewhere it would be easier to customize it for someone who wants to use different criteria for finding base pairs.
For the TertiaryBasePairParameters class, the rules are a little bit more relaxed because it's meant to look for things like G-U wobble base pairs and Reverse Hoogsteen base pairs.   In the base class, it's just lining up the sequences that's used and it's the simplest criteria.  So different classes have these as static final double values

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/MismatchedBasePairParameters.java
lukeczapla: Also complete!

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
lukeczapla: Yeah I think IllegalArgumentException sounds good.  So I did that

lukeczapla:
"Proposed revisions" attempts to address everything.  From the user's perspective, only an advanced user doing specialized research might want to override the cutoff settings that determine whether two bases are truly paired, so they're protected but non-final and have getters and setters.  Extending the classes is ideal for someone to take this even further, e.g., characterizing quadruplex structures.   getLength() works and the other suggestions were all integrated into the code.

lukeczapla:
ADDED ALSO - Serialization support.  The tests were just like this:

`structure = StructureIO.getStructure("1P71");
BasePairParameters bp = new MismatchedBasePairParameters(structure, false, false, false).analyze();
'


and then

`BasePairParameters bp = null;

File f = new File("serialize.test");
        if (f != null && f.exists()) try {
            InputStream file = new FileInputStream(f);
            InputStream buffer = new BufferedInputStream(file);
            ObjectInput input = new ObjectInputStream(buffer);
            //deserialize the Map
            bp = (BasePairParameters) input.readObject();
            file.close();

        } catch (Exception ex) {
            ex.printStackTrace();
            return;
        }
System.out.println(bp);   // <--- this outputs all the parameters that were in the original object
`


and then 

`

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/MismatchedBasePairParameters.java
```diff
@@ -0,0 +1,123 @@
...
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Contributed to BioJava under its LGPL
+ * This class allows for finding inter-strand base pairs that are not necessarily canonical Watson-Crick pairs.
+ * The implementation of findPair is different than that of the base class.
+ * @author Luke Czapla
+ * @since 5.0.0-snapshot
```
josemduarte: This should be `@since 5.0.0`. We don't use SNAPSHOTS in tags since those are transient builds changing every commit.

lukeczapla: Ok, all fixed in the commit "removed SNAPSHOT from ..."

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
+package org.biojava.nbio.structure.basepairs;
```
heuermh: Please add the license header

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+
+import org.biojava.nbio.structure.*;
```
heuermh: Star imports should be avoided

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+import org.biojava.nbio.structure.geometry.SuperPosition;
+import org.biojava.nbio.structure.geometry.SuperPositionQCP;
+import org.biojava.nbio.structure.io.PDBFileReader;
+
+import javax.vecmath.Matrix4d;
+import javax.vecmath.Point3d;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.*;
```
heuermh: Star imports should be avoided

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static java.lang.Math.*;
```
heuermh: Star imports should be avoided

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+
+/**
+ * Contributed to BioJava under its LGPL
```
heuermh: Please remove this line

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+ * This module calculates the el Hassan-Calladine Base Pairing and Base-pair Step Parameters
```
heuermh: Javadoc should be complete sentences, end with period

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+ */
+public class BasePairParameters implements Serializable {
+
+    private static final long serialVersionUID = 6214502385L;
+    private static Logger log = LoggerFactory.getLogger(BasePairParameters.class);
+
+    // See URL http://ndbserver.rutgers.edu/ndbmodule/archives/reports/tsukuba/Table1.html
+    // and the paper cited at the top of this class (also as Table 1).
+    // These are hard-coded to avoid problems with resource paths.
+    public static String[] standardBases = new String[] {
```
heuermh: All of these constants should be `private static final` and accessed via methods.  The field names should be UPPERCASE_WITH_UNDERSCORES.

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+                    "ATOM      6  C4    U A   1       0.989   2.884   0.000\n" +
+                    "ATOM      8  C5    U A   1       1.089   4.311   0.000\n" +
+                    "ATOM      9  C6    U A   1      -0.024   5.053   0.000\n"
+    };
+
+    // this is also hard-coded data about standard WC base pairs for both DNA and RNA
+    protected static String[] baseListDNA = {"A", "G", "T", "C"};
+    protected static String[] baseListRNA = {"A", "G", "U", "C"};
+    protected static Map<String, Integer> map;
+   // private static List<String> RNAspecific = Arrays.asList("U", "URA"),
```
heuermh: Please remove commented out code

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+   //        DNAspecific = Arrays.asList("DC", "C", "CYT");
+    protected static Map<Integer, List<String>> ringMap;
+    static {
+        map = new HashMap<>();
+        map.put("DA", 0); map.put("ADE", 0); map.put("A", 0);
+        map.put("DG", 1); map.put("GUA", 1); map.put("G", 1);
+        map.put("DT", 2); map.put("THY", 2); map.put("T", 2); map.put("U", 2); map.put("URA", 2);
+        map.put("DC", 3); map.put("CYT", 3); map.put("C", 3);
+        // chemically modified bases, leaving out to ignore (to treat as gaps) right now.
```
heuermh: Please remove commented out code

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+     */
+    public BasePairParameters(Structure structure, boolean useRNA, boolean removeDups, boolean canonical) {
+        this.structure = structure;
+        this.useRNA = useRNA;
+        this.canonical = canonical;
+        this.nonredundant = removeDups;
+
+    }
+
+    public BasePairParameters(Structure structure, boolean useRNA, boolean removeDups) {
```
heuermh: Add javadoc

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+        this(structure, useRNA, removeDups, false);
+    }
+
+    public BasePairParameters(Structure structure, boolean useRNA) {
```
heuermh: Add javadoc

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+    }
+
+
+    /**
+     * Return the buckle in degrees for the given base pair
+     * @param bp the number of the base pair (starting with 0)
+     * @return the value as a double (in degrees)
+     */
+    public Double getBuckle(int bp) {
+        if (bp < 0 || bp >= getPairingParameters().length) throw new IllegalArgumentException();
```
heuermh: Please add a message for the exception

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+        return pairingParameters[bp][0];
+    }
+
+    /**
+     * Return the propeller ("propeller-twist") in degrees for the given base pair
+     * @param bp the number of the base pair (starting with 0)
+     * @return the value as a double (in degrees)
+     */
+    public Double getPropeller(int bp) {
+        if (bp < 0 || bp >= getPairingParameters().length) throw new IllegalArgumentException();
```
heuermh: Please add a message for the exception

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+        return pairingParameters[bp][1];
+    }
+
+    /**
+     * Return the opening in degrees for the given base pair
+     * @param bp the number of the base pair (starting with 0)
+     * @return the value as a double (in degrees)
+     */
+    public Double getOpening(int bp) {
+        if (bp < 0 || bp >= getPairingParameters().length) throw new IllegalArgumentException();
```
heuermh: This check should be refactored into a method instead of repeated several times.  Or if guava is on the classpath for this module, use `checkArgument`

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+     * @return The list of corresponding Watson-Crick groups as pairs, as a Pair of nucleic acid Groups
+     */
+    public List<Pair<Group>> findPairs(List<Chain> chains) {
+        List<Pair<Group>> result = new ArrayList<>();
+        for (int i = 0; i < chains.size(); i++) {
+            Chain c = chains.get(i);
+            for (int j = i+1; j < chains.size(); j++) {
+                String complement = complement(chains.get(j).getAtomSequence(), useRNA);
+                String match = longestCommonSubstring(c.getAtomSequence(), complement);
+                //log.info(c.getAtomSequence() + " " + chains.get(j).getAtomSequence() + " " + match);
```
heuermh: Rather than comment out this log statement, you could leave it in and use a lower log level, e.g.
```
if (log.isDebugEnabled()) {
  log.debug(...)
}
```

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+    }
+
+
+    /**
+     * This method calculates pairing and step parameters from 4x4 transformation matrices
+     * that come out as Matrix4d;
+     * @param input the 4x4 matrix representing the transformation from strand II -> strand I or pair i to pair i+1
+     * @return Six parameters as double[6]
+     */
+    public static double[] calculatetp(Matrix4d input) {
```
heuermh: `calculatetp` &rarr; `calculateTp`

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+        M[3] = (cm*cg*cp-sm*sp)*A[0][3]+(sm*cg*cp+cm*sp)*A[1][3]-sg*cp*A[2][3];
+        M[4] = (-cm*cg*sp-sm*cp)*A[0][3]+(-sm*cg*sp+cm*cp)*A[1][3]+sg*sp*A[2][3];
+        M[5] = (cm*sg)*A[0][3]+(sm*sg)*A[1][3]+cg*A[2][3];
+
+        return M;
+
+    }
+
+
+    public static char complementBase(char base, boolean RNA) {
```
heuermh: Isn't this already present elsewhere in biojava?

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+        if (base == 'A' && RNA) return 'U';
+        if (base == 'A') return 'T';
+        if (base == 'T' && !RNA) return 'A';
+        if (base == 'U' && RNA) return 'A';
+        if (base == 'C') return 'G';
+        if (base == 'G') return 'C';
+        return ' ';
+    }
+
+    public static String complement(String sequence, boolean RNA) {
```
heuermh: Isn't this already present elsewhere in biojava?

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+        String result = "";
+        for (int i = sequence.length() - 1; i >= 0; i--) {
+            result += complementBase(sequence.charAt(i), RNA);
+        }
+        return result;
+    }
+
+    public static double[] cross(double[] a, double[] b) {
```
heuermh: Add javadoc

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+        assert a.length >= 3 && b.length >= 3;
+        double[] result = new double[4];
+        result[0] = a[1]*b[2]-a[2]*b[1];
+        result[1] = a[2]*b[0]-a[0]*b[2];
+        result[2] = a[0]*b[1]-a[1]*b[0];
+        return result;
+    }
+
+    public static double[] removeComponent(double[] a, double[] b) {
```
heuermh: Add javadoc

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+            dot += a[i]*b[i];
+        }
+        for (int i = 0; i < 3; i++) {
+            result[i] = a[i]-dot*b[i];
+        }
+        return result;
+
+    }
+
+    public static String longestCommonSubstring(String s1, String s2) {
```
heuermh: Add javadoc

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
```diff
@@ -0,0 +1,718 @@
...
+                if (x > max) {
+                    max = x;
+                    start = i;
+                }
+            }
+        }
+        return s1.substring(start, (start + max));
+    }
+
+    protected static boolean match(char a, char b, boolean RNA) {
```
heuermh: Isn't this already present elsewhere in biojava?

lukeczapla: If you don't know, I don't know!   I'll JavaDoc everything and fix the styling issues.

heuermh: That was a supposed to be an I'm-not-going-to-do-your-homework-for-you question :)

http://biojava.org/docs/api4.2.1/org/biojava/nbio/core/sequence/template/Compound.html#equalsIgnoreCase-org.biojava.nbio.core.sequence.template.Compound-

http://biojava.org/docs/api4.2.1/org/biojava/nbio/core/sequence/views/ComplementSequenceView.html

etc.

lukeczapla: I'll make all those utility classes private/protected so that it's clear that it is only internal usage

lukeczapla: if you can point me to a tutorial on how that works, I'll integrate it into the code.  Right now, I am just using atom records from chains.  So if that method will work on all PDB files just as well as the current method does (even ones generated from simulation codes with no headers), I'm open to trying it!

lukeczapla:
Added: JavaDoc for all internal function calls.  Changed log.info to a debug warning.  Renamed calculatetp to calculateTp.  Added BioJava license header to the top of each of the 4 files.  Removed commented-out code.  Changed "starred" imports to be specific imports.

lafita:
@lukeczapla let us know when you finish to address the comments on the code and we will review it again. I see that there are still some of @heuermh comments pending, specially in the `BasePairParameter` class.

lukeczapla:
Ok, I think I got everything.  As mentioned, I don't think the NucleotideCompound class is going to help much.  I can show you some interesting files I analyze, and they come out of programs that JUST write out the base-pairs (although sometimes they'll use this 3DNA tool to rebuild the sugar and phosphates, and I have a new JavaScript version of that).  Please let me know exactly what I missed otherwise.   See the new commits.

Thanks.

lukeczapla:
I would be willing to try to make a "rebuild" functionality (taking the base pair planes and building the sugars and phosphates in realistic positions) for BioJava, too.  It's in a way a mini-simulation and I know there's no simulation functionality but I have some very basic tools to make this work.

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
lukeczapla: I added the license header

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/BasePairParameters.java
lukeczapla: I reworked it to import particular classes

lafita: You should delete this line then.

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/MismatchedBasePairParameters.java
```diff
@@ -0,0 +1,154 @@
...
+                    }
+                    if (foundPair) pairSequence += (c.getAtomSequence().charAt(i));
+                    lastFoundPair = foundPair;
+                }
+            }
+        }
+        return result;
+    }
+
+    public static double getMaxStagger() {
```
lafita: Add JavaDoc to the getters and setters

lukeczapla:
sorry guys, I have to look into why IntelliJ refactored in the * import and change my settings.  I think it should now be OK with the two latest commits.  There's probably a setting in there to tell IntelliJ not to do that, definitely for java.util it was too much but I guess because so much of org.biojava.nbio.structure was used it just decided to turn it back into a * import!

josemduarte:
Could you merge the latest master into your branch? The tests should work then, I've committed a couple of small fixes needed after the v5 mmcif remediation

lukeczapla:
We'll see how it does on the tests.  Once this gets merged I'll basically just refork the whole package because my end gets a little messy with getting your changes to the pom.xml and such.  I think I got it though.

lafita:
@lukeczapla if you need extra dependencies in the pom.xml and your code is more experimental, I would recommend creating a new Maven project with the BioJava dependencies you need in it. This is better practice than forking the whole BioJava repository and makes the integration of the code much easier in the future (since it has been independently developed and there are no conflicts). We have done that many times in the past, and it follows our philosophy of code modularity and simplicity.

lukeczapla:
Hi @lafita I have a separate project and just put these files together for BioJava in case other people want to use them within your framework.  I know several people who have interest in this, and that's why I went through the whole review process and spent all this time commenting everything as requested.  There are no dependencies in the code I put in the PR except ones that you guys use!

lafita:
Yes this PR is fine, I was referring to your sentence before:

>  Once this gets merged I'll basically just refork the whole package because my end gets a little messy with getting your changes to the pom.xml and such

lukeczapla:
If I develop anything new suitable for BioJava, that would be my plan.  I was just doing that whole thing to contribute to your project (for free).  I'm not sure what's going on with the other reviewers but if the standard for commenting/documentation is of the level of a random BioJava file (e.g., CallableStructureAlignment.java) it seems I've done my job above and beyond!

lafita:
If we do not hear anything from @josemduarte or @heuermh by Wednesday we can merge the PR.

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/MismatchedBasePairParameters.java
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * This class allows for finding inter-strand base pairs that are not necessarily canonical Watson-Crick pairs.
+ * The implementation of findPair is different than that of the base class.  This class does not consider intra-strand
+ * base pairing and for that, the TertiaryBasePairParameters class should be used.
+ * @author Luke Czapla
+ * @since 5.0.0
+ *
+ */
+public class MismatchedBasePairParameters extends BasePairParameters implements Serializable {
```
lafita: Do not need to implement `Serializable`, since `BasePairParameters` already is serializable.

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/TertiaryBasePairParameters.java
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * This class also finds the base pairing and base-pair step parameters but has a broader definition
+ * of a base pair so that non-canonical-WC base pairs will be detected and reported.  This is useful
+ * for RNA that has folded into different regions, and for higher-order DNA structures.  Intra-strand
+ * pairings are considered in this class (but in not the base class or MismatchedBasePairParameters class)
+ * @author Luke Czapla
+ * @since 5.0.0
+ *
+ */
+public class TertiaryBasePairParameters extends BasePairParameters implements Serializable {
```
lafita: Do not need to implement `Serializable`, same as above.

lukeczapla: It's interesting, originally I didn't have those two as Serializable, and my tests worked fine.  I forget when I added that but I just committed the version that takes that out.

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/MismatchedBasePairParameters.java
```diff
@@ -0,0 +1,186 @@
...
+
+    // These are the criteria used to select proper base pairs.
+    protected static double MaxStagger = 2.0, MaxShear = 5.0, MaxStretch = 5.0,
```
lukeczapla: They have getters and setters actually, so they are not actually constants.  If someone really knows what they're doing, they can change these criteria.. I was just talking to Wilma Olson (a professor at Rutgers) about this in her office because I was down there for a teachers conference.  There are many different ways people choose the criteria for a base pair, she seemed to think this was good and showed me exactly what kinds of "base pairs" get taken out by the particular Stagger criteria I used.  It was stuff that shouldn't really belong as characterized as a base pair.

biojava-structure/src/main/java/org/biojava/nbio/structure/basepairs/TertiaryBasePairParameters.java
```diff
@@ -0,0 +1,136 @@
...
+
+    // These are the criteria used to select proper base pairs.
+    protected static double MaxStagger = 2.0, MaxPropeller = 60.0;
```
lukeczapla: As mentioned above, they have getters and setters and those who really know what they're doing can change these criteria before calling analyze().  I had a long discussion about the methodology and it seems to be good with the default parameters but if people want to look for unusual "pairings" everything can be changed.

lukeczapla: Would you like me to call them maxStagger (etc.) instead of MaxStagger?  That I think follows convention...

heuermh: If they have getters and setters then by definition they aren't constants, and so shouldn't be static or final or visible at protected scope.  For this situation, I usually prefer
```java
public static final double DEFAULT_MAXIMUM_STAGGER = 2.0d;
private double maximumStagger = DEFAULT_MAXIMUM_STAGGER;
public final double getMaximumStagger() { return maximumStagger; }
public final void setMaximumStagger(final double maximumStagger) { this.maximumStagger = maximumStagger; }
```
A downside of this pattern is the class is not immutable.  Immutable classes are preferrable for thread safety and performance issues.

lukeczapla: I agree with you there.  So I made these values non-static (for thread safety with multiple independent instances) and created the final DEFAULT_MAX_STAGGER etc.  I'm not sure the use of the final keyword in the setter, it's possible that someone would create the object, call analyze() once and look at that data, then change the maxStagger or one of those values, and then call analyze() again and look at the new data.  That would be like if someone wanted to see those weird "base pairs" that Wilma Olson showed me, they were hardly really paired because Stagger is a measure of how planar the pair is.  But there is a lot of distorted crap that specialists might like to see.

PR #676, opened 2017-06-29
Serialize all the Structure objects

lafita:
This is a step towards #673.

I made the interfaces of the classes extend `Serializable`, so that any implementation is serializable, and I created a test that checks the serialization of a BioJava structure.

Issue #673, opened 2017-06-29
Make biojava objects serializable


lafita:
Before the 5.0 release it would be nice to make BioJava objects serializable, which allows their use by applications like Spark.

heuermh:
That isn't necessarily sufficient for use in Apache Spark.  See https://github.com/heuermh/biojava-adam/blob/master/src/main/java/org/biojava/nbio/adam/BiojavaKryoRegistrator.java#L43

lafita:
It might not be sufficient but it is necessary, so I think there is no harm in serializing all possible BioJava objects.

@heuermh do you know what other conditions are needed to have full access to Apache Spark?

heuermh:
Perhaps not necessary either :)

With that registrator everything (that I've run into so far) can be serialized using [Kryo](https://github.com/EsotericSoftware/kryo), which is much more efficient than Java serialization both in terms of space and performance.  I plan to donate that repo to the biojava organization when its dependencies are released.

lafita:
We need a test similar to 498a36fb20b6c07178d56ad575bf6c594a9db268 (`structure`) for the `core` data models, and fix any serialization issues that occur.

lukeczapla:
Hi guys, I addressed this in my code, but the Structure object is not serializable and the particular error I get is `java.io.NotSerializableException: org.biojava.nbio.structure.secstruc.SecStrucState`

So I marked my Structure object as transient (since it's the step parameter data that is mostly important for serialization in my package, meaning the Structure has been converted into its step parameter and pair parameter data in the procedure) and it is working to serialize and deserialize results.

lafita:
Good catch @lukeczapla. We need to make the `SecStrucState` class serializable, since everything in a `Structure` object has to be serializable. I forgot to assign the secondary structure in the test I created in #676. I will fix that as soon as possible.

josemduarte:
#676 already took care of some of this. Is there something more to be done @lafita , @heuermh ?

I'd propose that we push this over to the 5.1 milestone. Even if we need to add some more serializable interfaces it would be an addition/minor API change.

lafita:
We serialized all classes we needed for Spark, but we encountered a bug #676 
I would say we can close this issue and open new ones if we need more classes to be serialized.

PR #693, opened 2017-07-28
Make secondary structure objects serializable

lafita:
This PR implements Serializable in the SecStruc objects that are stored in a Structure and adds the secondary structure to the serialized structure in the test.

The problem was reported by @lukeczapla in #673 and was missing from #676


josemduarte:
The tests are fixed now, could you merge the latest master so that they are triggered again?

lukeczapla:
I will try to unmark the Structure object as transient and serialize my class, to see if it works.  that'd be really nice if the whole class could be serialized with all the structural data!

Issue #695, opened 2017-08-02
OBO parser: namespace


tulay:
Hi,
I don't see where namespace is stored in the ontology. For Gene ontology there are several namespaces as opposed to the single sequence namespace in [biosapiens.obo](http://sourceforge.net/p/song/svn/HEAD/tree/trunk/subsets/biosapiens.obo?format=raw) in the demo.
Thanks!

lafita:
I think the `namespace` is not parsed, but I would imagine that `Ontology.getName()` is the place for it, currently only set by the user. 

Is the `namespace` important for the interpretation of the ontology? What is the use case for parsing it as well?

Can you provide a file where there are several namespaces? 

Thanks @tulay!

tulay:
An example ontology for this issue: [go.obo](http://www.geneontology.org/ontology/go.obo)

PR #698, opened 2017-08-08
Biojava #695: Stores namespace as annotation

tulay:
Fix to #695

### DIFF
--- a/biojava-ontology/src/test/java/org/biojava/nbio/ontology/TestParseOBO.java
+++ b/biojava-ontology/src/test/java/org/biojava/nbio/ontology/TestParseOBO.java
@@ -24,22 +24,16 @@ package org.biojava.nbio.ontology;
 
 import junit.framework.TestCase;
 import org.biojava.nbio.ontology.io.OboParser;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


biojava-ontology/src/test/java/org/biojava/nbio/ontology/TestParseOBO.java
```diff
@@ -0,0 +1,83 @@
...
+
+			ontology = parser.parseOBO(oboFile, "so-xp/subsets/biosapiens",
+                    "snippet from biosapiens protein feature ontology");
+			Set<Term> keys = ontology.getTerms();
+
+			assertTrue(keys.size() > 1);
+			assertTrue(ontology.getTerm("SO:0000691").getAnnotation().containsProperty(NAMESPACE));
+            assertEquals("sequence",ontology.getTerm("SO:0000691").getAnnotation().getProperty(NAMESPACE));
+
+		} catch (Exception e) {
```
josemduarte: Better to throw the exception instead of catching. That way you'd get an error instead of a failure, which is what you want when an exception happens.

tulay: Agreed.

PR #701, opened 2017-08-15
Fixing bug in Chromosome Mapping Tools 

valasatava:
bug in calculating CDS length

### DIFF
--- a/biojava-genome/src/main/java/org/biojava/nbio/genome/util/ChromosomeMappingTools.java
+++ b/biojava-genome/src/main/java/org/biojava/nbio/genome/util/ChromosomeMappingTools.java
@@ -504,7 +504,6 @@ public class ChromosomeMappingTools {
      */
     public static int getCDSLengthReverse(List<Integer> exonStarts, List<Integer> exonEnds, int cdsStart, int cdsEnd) {
 
-        boolean inCoding = false;
         int codingLength = 0;
 
         if (cdsEnd < cdsStart) {
@@ -512,8 +511,6 @@ public class ChromosomeMappingTools {
             cdsEnd = cdsStart;
             cdsStart = tmp;
         }
-
-        int lengthExons = 0;
 
         // map reverse
         for (int i = exonStarts.size() - 1; i >= 0; i--) {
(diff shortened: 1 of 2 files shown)


biojava-genome/src/main/java/org/biojava/nbio/genome/util/ChromosomeMappingTools.java
```diff
@@ -28,7 +28,11 @@
 
     public static final String CHROMOSOME = "CHROMOSOME";
     public static final String CDS = "CDS";
-    
+
+    public static int base = 1;
```
josemduarte: Shouldn't this be a member of this class rather than a static?

biojava-genome/src/test/java/org/biojava/nbio/genome/TestChromosomeMappingTools.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static junitx.framework.ComparableAssert.assertEquals;
+
+/**
+ * Created by Yana Valasatava on 8/14/17.
```
josemduarte: Could you use the `@author` tag here and also write a minimal explanation of what the test does?

biojava-genome/src/test/java/org/biojava/nbio/genome/TestChromosomeMappingTools.java
```diff
@@ -0,0 +1,61 @@
+package org.biojava.nbio.genome;
+
+import org.biojava.nbio.genome.util.ChromosomeMappingTools;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static junitx.framework.ComparableAssert.assertEquals;
```
josemduarte: This should be imported from package org.junit.assert. I usually simply do this:
```java
import static org.junit.Assert.*;
```
which imports all the assert static methods.

valasatava:
Is there anything else I can improve?

PR #709, opened 2017-10-07
several fixes for biological assemblies calculations

valasatava:


biojava-structure/src/main/java/org/biojava/nbio/structure/symmetry/core/QuatSymmetryResults.java
```diff
@@ -111,14 +111,17 @@ public QuatSymmetryResults(List<SubunitCluster> clusters,
...
-	
+
 	/**
 	 * Returns the List of Subunits used to calculate symmetry.
 	 *
 	 * @return an unmodifiable view of the List
 	 */
 	public List<Subunit> getSubunits() {
-		return Collections.unmodifiableList(subunits);
+		if (subunits != null)
```
lafita: Subunits should never be null. The bug is that some code is missing from the helical symmetry constructor. The right fix is to add the following code into line 101:

<pre>
 subunits = new ArrayList<Subunit>();
 for (SubunitCluster c : clusters) {
 	subunits.addAll(c.getSubunits());
 }
<pre>

josemduarte: Thanks for the tip, now done in 96c6816366bbc7261e304aafee79c8a3ba04e

Issue #708, opened 2017-10-05
QuatSymmetryResults.isPseudoSymmetry() is always false


josemduarte:
It looks like the flag is never set to true in the cases where there is pseudosymmetry. This test actuall has a TODO note for that: https://github.com/biojava/biojava/blob/5f138f314712c2a5fd3950f1b034dae7661231a8/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/symmetry/TestQuatSymmetryDetectorExamples.java#L95

josemduarte:
A question here is when should a symmetry result be pseudo-symmetric and when should it be pseudo-stoichiometric. I would say both have the same meaning. 

For pseudostoichiometry, this is what we are doing:

https://github.com/biojava/biojava/blob/359dce819e7678879ae41301b11fa268f92e81d1/biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitClusterUtils.java#L57-L74

i.e. whenever the method for clustering was `STRUCTURE` then we call it pseudostoichiometric. Why such a definition? what if we clustered by the `SEQUENCE` method, shouldn't that be pseudostoichiometric too?

In my understanding pseudosymmetry/pseudostoichiometry should be true whenever some clustering happened (either by `SEQUENCE` or by `STRUCTURE`)

josemduarte:
After some research, it looks like this combination is possible:
```
isPseudoStoichiometry=true
isPseudoSymmetric=false
```
which happens whenever the stoichiometry is based on clustered subunits, but there's no symmetry detected on clustered subunits.

This combination is not possible:
```
isPseudoStoichiometry=false
isPseudoSymmetric=true
```

josemduarte:
I'd argue that the combination
```
isPseudoStoichiometry=true
isPseudoSymmetric=false
```
doesn't add much value and that we could drop `isPseudoSymmetric` to simplify the code. I'll submit a PR for that.

lafita:
I agree with that, I think that if there is pseudo-stoichiometry there has to be pseudo-symmetry. 

I think the original idea for splitting them was that pseudo-stoichiometry depends on the clustering algorithm and pseudo-symmetry depends on the symmetry detection algorithm, so they could differ. However, I would say that biologically it makes little sense, so let's keep pseudo-stoichiometry only.

[...]

PR #710, opened 2017-10-09
Fixes for quaternary symmetry code

josemduarte:
As discussed in #708, this removes isPseudoSymmetric and keeps only isPseudoStoichiometric. Also slightly changes the implementation of isPseudoStoichiometric.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitClusterUtils.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitClusterUtils.java
@@ -56,7 +56,7 @@ public class SubunitClusterUtils {
 
 	/**
 	 * A pseudostoichiometric {@link SubunitCluster} was obtained using the
-	 * {@link SubunitClustererMethod#STRUCTURE} similarity.
+	 * {@link SubunitClustererMethod#STRUCTURE} or {@link SubunitClustererMethod#SEQUENCE} similarity.
 	 * 
 	 * @param clusters
 	 * @return true if any of the clusters is pseudostoichiometric, false
@@ -65,7 +65,8 @@ public class SubunitClusterUtils {
 	public static boolean isPseudoStoichiometric(List<SubunitCluster> clusters) {
 
 		for (SubunitCluster c : clusters) {
-			if (c.getClustererMethod() == SubunitClustererMethod.STRUCTURE)
+			if (c.getClustererMethod() == SubunitClustererMethod.STRUCTURE ||
+				c.getClustererMethod() == SubunitClustererMethod.SEQUENCE)
 				return true;
 		}
 		return false;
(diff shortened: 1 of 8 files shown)


biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitClusterUtils.java
```diff
@@ -65,7 +65,8 @@ public static String getStoichiometryString(List<SubunitCluster> clusters) {
 	public static boolean isPseudoStoichiometric(List<SubunitCluster> clusters) {
 
 		for (SubunitCluster c : clusters) {
-			if (c.getClustererMethod() == SubunitClustererMethod.STRUCTURE)
+			if (c.getClustererMethod() == SubunitClustererMethod.STRUCTURE ||
+				c.getClustererMethod() == SubunitClustererMethod.SEQUENCE)
```
lafita: This will cause assemblies with different entity chains (not 100% identical) to be called pseudo-stoichiometric. I am not saying this is a bad thing, just in case you were not aware.

josemduarte: To me any non-identical sequence should be consider pseudo. Perhaps I'd lower the identity theshold to 95% (instead of 100%). Would it be possible to get the threshold here?

lafita: The Subunit clustering algorithm is designed with three levels:

- Identity: chains with same SEQRES (same entity) are clustered together
- Sequence: sequence alignment is performed on the SEQRES and if the sequence identity is higher than a threshold chains are clustered together
- Structure: structure alignment with a RMSD and coverage threshold

I think there is not a unique definition of pseudo-stoichiometry, so maybe we should give a clear definition for this flag. In the context of symmetry analysis, I would say that structural alignment has to be used to generate the pseudo-stoichiometric alignment (geometric property), so this was the reason to set TRUE only for structural clustering.

lafita: Since the user can change the sequence identity threshold, I agree that a sequence similarity clustering with 30% threshold should be considered pseudo-stoichiometry, for example.

lafita: I might have complicated things too much. Maybe the easiest solution is to just re-implement the first level (Identity) to be a sequence alignment at fixed 95% sequence identity.

josemduarte: >Maybe the easiest solution is to just re-implement the first level (Identity) to be a sequence alignment at fixed 95% sequence identity.

I've added a commit that does exactly that. Could you recheck if it looks ok?

PR #716, opened 2017-12-07
Fixing issue with Calc.transform(Chain,Matrix4d)

### LINKED ISSUE #715: Calc.transform(Chain) does not transform alt groups
Looks like an issue in master branch and in 4.2.x branch.

josemduarte:
Should fix #715

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/Calc.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/Calc.java
@@ -538,12 +538,13 @@ public class Calc {
 	 * @param m
 	 */
 	public static final void transform(Group group, Matrix4d m) {
-		AtomIterator iter = new AtomIterator(group);
-
-		while (iter.hasNext()) {
-			Atom atom = iter.next();
+		for (Atom atom : group.getAtoms()) {
 			transform(atom, m);
-
+		}
+		for (Group altG : group.getAltLocs()) {
+			for (Atom atom : altG.getAtoms()) {
+				transform(atom, m);
+			}				
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


biojava-structure/src/main/java/org/biojava/nbio/structure/Calc.java
```diff
@@ -601,12 +601,14 @@ public static final void translate(Atom atom, Vector3d v) {
...
-		AtomIterator iter = new AtomIterator(group);
 
-		while (iter.hasNext()) {
-			Atom atom = iter.next();
+		for (Atom atom : group.getAtoms()) {
 			translate(atom, v);
-
+		}
+		for (Group altG : group.getAltLocs()) {
+			for (Atom atom : altG.getAtoms()) {
```
lafita: Why not using `translate(g, v)`? It would be a recursive call to this function, nicer solution.

biojava-structure/src/main/java/org/biojava/nbio/structure/Calc.java
```diff
@@ -538,12 +538,13 @@ public static final void transform(Atom atom, Matrix4d m) {
...
-		AtomIterator iter = new AtomIterator(group);
-
-		while (iter.hasNext()) {
-			Atom atom = iter.next();
+		for (Atom atom : group.getAtoms()) {
 			transform(atom, m);
-
+		}
+		for (Group altG : group.getAltLocs()) {
+			for (Atom atom : altG.getAtoms()) {
```
lafita: Recursive call to `transform(g, m)` could also be possible here.

josemduarte: Good point, now changed

PR #724, opened 2018-01-19
Small refactoring for more flexible symmetry detection.

biocryst:
More options for the clustering parameters and superposition algorithms.
I tried to preserve the previous default behaviour whenever possible.

@lafita Aleix, I believe most of the code I touched was written by you, could you please take a look? Thank you!

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/align/ce/CeParameters.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/align/ce/CeParameters.java
@@ -71,11 +71,11 @@ public class CeParameters implements ConfigStrucAligParams  {
 	protected boolean showAFPRanges;
 	protected int  sideChainScoringType;
 
-	// whether to perform final optimisation of the alignment (slow),
-	// which "attempts  to  increase the  alignment  length
-	// found  previously  while  keeping  the r.m.s.d.  at
-	// about  the  same  level" (Shindyalov and Bourne, 1998)
-	// may not be needed for some applications
+	/**
+	 * Whether the CE algorithm should extend the best found trace with dynamic programming,
+	 * while keeping RMSD at about the same level. (Shindyalov and Bourne, 1998)
+	 * This is useful for edge cases with remote homology, but can be slow for large structures.
+	 */
 	private boolean optimizeAlignment;
 
 	protected static final double DEFAULT_GAP_OPEN = 5.0;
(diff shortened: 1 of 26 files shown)


biojava-structure/src/main/java/org/biojava/nbio/structure/align/ce/CeParameters.java
```diff
@@ -418,6 +425,12 @@ public void setSubstitutionMatrix(
 	}
 
 
+	public boolean isOptimizeAlignment() {
+		return optimizeAlignment;
```
lafita: Can you copy the documentation for the parameter here? Users can read it here.

biojava-structure/src/main/java/org/biojava/nbio/structure/align/ce/CeParameters.java
```diff
@@ -418,6 +425,12 @@ public void setSubstitutionMatrix(
...
+	}
 
+	public void setOptimizeAlignment(boolean optimizeAlignment) {
+		this.optimizeAlignment = optimizeAlignment;
```
lafita: Can you copy the documentation for the parameter here? Users can read it here. (same here)

biojava-structure/src/main/java/org/biojava/nbio/structure/align/fatcat/calc/FatCatParameters.java
```diff
@@ -455,4 +456,11 @@ private Object  getValue(String name){
 
 	}
 
+	public boolean isOptimizeAlignment() {
+		return optimizeAlignment;
+	}
```
lafita: Can you copy the documentation for the parameter here? Users can read it here. (same here)

biojava-structure/src/main/java/org/biojava/nbio/structure/align/fatcat/calc/FatCatParameters.java
```diff
@@ -455,4 +456,11 @@ private Object  getValue(String name){
...
+
+	public void setOptimizeAlignment(boolean optimizeAlignment) {
+		this.optimizeAlignment = optimizeAlignment;
```
lafita: Can you copy the documentation for the parameter here? Users can read it here. (same here)

biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitClustererParameters.java
```diff
@@ -39,13 +41,62 @@
...
+	private static final double hcSequenceIdentityLocal = 0.95;
+	private static final double hcSequenceCoverageLocal = 0.75;
+	private static final double hcSequenceIdentityGlobal = 0.85;
+
+	public SubunitClustererParameters() {
+		this(false); // initialize with local metrics by default
+	}
+
+	public SubunitClustererParameters(boolean iUseGlobalMetrics) {
+		useGlobalMetrics = iUseGlobalMetrics;
```
josemduarte: Docs for this would be important. Also the 2 variables can be called the same way and then use:
```java
this.useGlobalMetrix=useGlobalMetrics;
```

biocryst:
@lafita, @josemduarte Thanks for the review! I updated the documentation, hope it's better now.

Regarding the parameter for the sequence identity calculation, I believe it is better to make this one mandatory. By looking quickly through literature "the default" definition seems to vary depending on whether one works with sequences or structures, uses local or global alignments, etc.. Moreover, from the code history here, when this method was first introduced it normalized by the alignment length, then it was changed to the number of aligned pairs, but somewhat confusing doc-strings remained. I could only figure out what it actually counts through debugging...

lafita:
Ok, I understand. If it is properly documented should not be a problem!

Issue #719, opened 2018-01-10
calcLocalSymmetries works on a cluster level, not subunit level


biocryst:
In the current implementation a local symmetry can only be detected, if

1) the assembly contains multiple clusters;
2) all subunits in a cluster (or clusters) are related by a symmetry axis.

(see https://github.com/biojava/biojava/blob/32255b15d7af3f8eb550e05d3ef161e0ab3dfff8/biojava-structure/src/main/java/org/biojava/nbio/structure/symmetry/core/QuatSymmetryDetector.java#L185 and below)

These are rarely satisfied in practice. See PDB 3L28 as an example of what is missed. It is a homo 6-mer (so one cluster in total) with two local C2 axes (so only 2 out of 6 subunits are related for each symmetry). http://www.rcsb.org/pdb/explore/jmol.do?structureId=3L28&bionumber=1

Using power-set of subunits would be a quick fix, although terribly inefficient.

lafita:
The power-set of subunits, instead of subunit clusters, was used before, but it meant that the algorithm scaled exponentially on the number of subunits instead of on the number of subunit clusters. 

This comes from the assumption that homomeric subunits cannot be asymmetric, because it would require unequal engaged interfaces in each monomer. However, there are such cases in the PDB, as you mention, and they could be possible annotation errors. It just occurred to me that this could be a good check for wrongly annotated assemblies.
If we want to detect them again I would suggest a more clever heuristic to find them, rather than the combinatorial way.

biocryst:
Closed due to #726

PR #726, opened 2018-02-07
Local symmetry changes

biocryst:
As a follow-up to #719. Local symmetry is now calculated in two steps:
1. Check symmetry of individual clusters and groups of clusters (no  powerset used, just group all by found symmetry and stoichiometry).

2. Check symmetry of subunit groups based on protein contact graph, similar to the way it was done before clustering was introduced for this functionality. As a heuristic optimisation, the graph is recursively decomposed via clique minimal separators. This allows extracting all connected (possibly overlapping) sets of subunits in reasonable time for all but a few assemblies (currently 3J3Y, 3JAO and 3J3Q go over the 2 minutes threshold).

Since there may be hundreds of local symmetries, they are grouped hierarchically (e.g. if protein has stoichiometry of A2B2C, and there are 3 local C2 symmetries found which correspond to A2, B2 and A2B2, then only A2B2 will be output. Likewise, D2 will supersede two C2 symmetries it is based on, etc.

Notes:
No changes to the method interfaces were made.
Update of jrgapht version is required for CliqueMinimalSeparatorDecomposition.


biojava-structure/src/main/java/org/biojava/nbio/structure/symmetry/core/SubunitContactGraph.java
```diff
@@ -22,6 +22,7 @@
 
 import javax.vecmath.Point3d;
 
+import org.jgrapht.Graph;
 import org.jgrapht.UndirectedGraph;
```
josemduarte: Could you remove this import? It's unused and has been deprecated in jgrapht

biojava-structure/src/main/java/org/biojava/nbio/structure/symmetry/core/SubunitContactGraph.java
```diff
@@ -45,17 +46,17 @@
...
 
 	private static final double DISTANCE_CUTOFF = 8;
-	private static final int MIN_CONTACTS = 10;
+	private static final int MIN_CONTACTS = 5;
 
 	private SubunitContactGraph() {
 	}
 
-	public static UndirectedGraph<Integer, DefaultEdge> calculateGraph(
+	public static Graph<Integer, DefaultEdge> calculateGraph(
```
josemduarte: It would be nice to do the contact calculation with the spatial hashing algorithm implemented elsewhere in biojava. The tutorial shows how to use it: https://github.com/biojava/biojava-tutorial/blob/master/structure/contact-map.md

biocryst:
Thanks for the tip! I rewrote the initialisation of a contact graph with the grid-based functions.

biojava-structure/src/main/java/org/biojava/nbio/structure/symmetry/core/QuatSymmetryParameters.java
```diff
@@ -49,14 +49,9 @@
 	private double minimumHelixRise = 1.0;
 	private double minimumHelixAngle = 5.0; // min helix angle to differentiate
 											// it from a translational repeat
-	private int maximumLocalCombinations = 50000; // max number of combinations
-													// to try for local symmetry
-													// calculation
-	private int maximumLocalResults = 1000;
-	private int maximumLocalSubunits = 20; // maximum number of subunits for
-											// local symmetry calculations
 	private double localTimeLimit = 120; // time limit for local calculations in
```
lafita: I am not sure that a time limit is a good idea here (this is how it was done before). 

I preferred calculating the number of possible combinations and only evaluating a subset of them if they were too large, since it allows to know before starting the calculation and we can use whatever heuristic we considered adequate to select the combinations we want. 
Also, it makes it easier to debug (in case a structure fails with too long runtime but very few combinations it would be hard to notice how the code is constructed now). In addition, runtime is very much system dependent and can make results less reproducible.

What is your logic to go back to using time limits to stop the calculation for large structures?

biojava-structure/src/main/java/org/biojava/nbio/structure/symmetry/core/QuatSymmetryDetector.java
```diff
@@ -50,6 +57,9 @@
 	private static final Logger logger = LoggerFactory
 			.getLogger(QuatSymmetryDetector.class);
 
+	private static final double CONTACT_GRAPH_DISTANCE_CUTOFF = 8;
+	private static final int CONTACT_GRAPH_MIN_CONTACTS = 5;
```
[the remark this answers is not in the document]
lafita: Oh I see that you copied them from another file you deleted. In that case if you know what they are you can add it. I think `CONTACT_GRAPH_MIN_CONTACTS ` was the number of residue contacts between subunits to add an edge and `CONTACT_GRAPH_DISTANCE_CUTOFF ` was the distance in A between residues to consider them as contacting (which is quite generous now that I think about it, usually contact distances are about 5A, so feel free to change the value if you think it makes sense).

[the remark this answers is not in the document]
josemduarte: The cutoff being 8A is because it is between Calphas. 5 would be too short for Calpha contacts.

pwrose:
I introduced the time limit a long time ago to make sure the PDB weekly update doesn't get stuck in pathological cases. I agree that by default this option should probably be not set.

biocryst:
Thanks for the review!

1) I added doc-strings for the protein contact cutoff constants. Indeed, they were copied from the previous implementation.

2) There is TestQuatSymmetryDetectorExamples::testLocal (in the biojava-integrationtest package). I extended it, so now it tests 15 structures which have different local symmetry arrangements. Some of those are probably bugs in bioassembly specifications, so the test might need adjustments in the future, if/when they are fixed. But for now they make nice test-cases for the algorithm.

3) Both running time and the number of combinations are limited in the current implementation, although the limit for the number of combinations typically fires first. Initially I kept only the time limit, because it is hard to predict the complexity of the graph in terms of the resulting combinations. It depends on the patterns of connectivity. For example, 4V3P (~1000+ subunits) is handled in a few seconds, while 3JAO (~300 subunits) did not finish after many hours. Moreover, some combinations take way longer to test for symmetry than others, so limiting by time appeared a more consistent solution.

Anyway, I changed the default behaviour to limit only the number of combinations, as you suggested. The time control can be enabled by the calling method through QuatSymmetryParameters (for things like PDB weekly update).

PR #728, opened 2018-02-14
Inject NCS operators into SpaceGroup

biocryst:
NCS is currently handled by duplicating chains in StructureTools::expandNcsOps, which is not optimal for further analysis (e.g., interface duplications in CrystalBuilder).

My suggestion is to combine them with the SpaceGroup transformations. The benefit of this approach is that a unit cell traversal is consistent for a client code, whether we have NCS or not. The drawback is that a SpaceGroup object is not guaranteed to be a true crystallographic space group anymore.

I also added a couple of tests for structures with NCS operators into TestCrystalBuilder.

### DIFF
--- a/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/xtal/TestCrystalBuilder.java
+++ b/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/xtal/TestCrystalBuilder.java
@@ -114,6 +114,13 @@ public class TestCrystalBuilder {
 		CrystalBuilder cb = new CrystalBuilder(s1);
 		StructureInterfaceList interfaces = cb.getUniqueInterfaces(5.5);
 		assertEquals(12,interfaces.size());
+		// kill the cell info to simulate incorrect and/or missing
+		s1.getCrystallographicInfo().setCrystalCell(null);
+		cb = new CrystalBuilder(s1);
+		interfaces = cb.getUniqueInterfaces(5.5);
+		//same number of interfaces should be found
+		// from ncs operators
+		assertEquals(12,interfaces.size());
 	}
 
 	@Test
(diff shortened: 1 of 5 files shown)


biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/xtal/TestCrystalBuilder.java
```diff
@@ -104,6 +104,30 @@ public void test4MF8() throws IOException, StructureException {
...
+	@Test
+	public void test1AUY() throws IOException, StructureException {
+		// a virus with NCS operators
+		AtomCache cache = new AtomCache();
+		StructureIO.setAtomCache(cache);
+		cache.setUseMmCif(true);
+		Structure s1 = StructureIO.getStructure("1AUY");
+		CrystalBuilder cb = new CrystalBuilder(s1);
+		StructureInterfaceList interfaces = cb.getUniqueInterfaces(5.5);
+		assertTrue(interfaces.size()==12);
```
josemduarte: Could you make this an assertEquals? Minor thing, but it makes the test reports nicer (on failures). Same below.

biocryst:
Thanks for the review! I modified the code so that a structure with NCS parameters but without a well-defined cell will look for contacts within the "expanded" asymmetric unit.

biocryst:
@sbliven Good point, thank you!

Initially I did not intend for this 'extended' space group to go beyond CrystalBuilder, but only use it locally, e.g.:
`SpaceGroup sg = new SpaceGroup(this.crystallographicInfo.getSpaceGroup());
sg.extendNCS(ncsOperators);`

Which worked on its own. But then, a StructureInterface stores the CrystalTransforms that were involved, which publicly provide their id (=serial number within SG), which is used in some places to get the transforms or their properties directly from SG, assuming SGs never change. So to make it all work I had to replace the original space group in the PDBCrystallographicInfo. I agree that it may break some assumptions about the SpaceGroup object, my hope was that SpaceGroup.isNCSExtended() method should indicate to the users that this unit cell is a bit more complicated than usual.

After sifting through the code a bit, I believe it can also be fixed by removing CrystalTransform.transformId altogether, and use other means where appropriate (in general, an object should not know its position in some list). Let me know if you agree, or there is something else I am missing!

biocryst:
Closing this for now, as I am exploring a different solution.

PR #730, opened 2018-02-20
Stoichiometry string to object

biocryst:
Proposed refactoring of SubunitClusterUtils.java into a separate object, which represents subunit clustering results, thus replaces "List<SubunitCluster>" in most places as well.

The primary motivation was to fix the stoichiometry String representation, which is currently not very usable beyond 52 clusters. The proposed 'string overflow strategies' are not exhaustive. Importantly, having an object here will allow more flexibility in dealing with assemblies in terms of grouping their subunits.

The name 'Stoichiometry' is tentative, another possibility is AssemblyComposition, ... ?

I added TestSubunitClustererExamples.testStoichiometry() which demonstrates how it can be used.


biojava-structure/src/main/java/org/biojava/nbio/structure/symmetry/core/Stoichiometry.java
```diff
@@ -0,0 +1,290 @@
...
+	 * @return Human-readable representation of this stoichiometry.
+	 */
+	@Override
+	public String toString() {
+		String formula =
+				orderedClusters.stream().
+					map((SubunitCluster r)-> {
+						String output = r.getAlpha();
+						if(output.length()>1 || r.size()>1)
+							output+=r.size();
```
josemduarte: This might be potentially slow for cases with many SubunitClusters. Ideally a StringBuilder should be used. See https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java

biocryst:
Thanks for the review! I removed SubunitClusterUtils.java and changed the toString() method to use StringBuilder.

As for the different strategies, I added a possibility to supply a function of predefined type (Function<List<SubunitCluster>,String>), with an example of how to use it in the TestSubunitClustererExamples.testStoichiometry() method. So a user may replace the entire string generation logic, if needed. Let me know if this is sufficient!

PR #732, opened 2018-03-01
NCS-aware CrystalBuilder object

biocryst:
Alternative solution to #728. 
Asymmetric unit extension is used with additional chains, as is currently done in biojava. The chain equivalence is tracked and passed on to the CrystalBuilder. As the result, it produces "clustered" list of interfaces based on NCS-equivalence (same original chains in the same relative position).

This change is not optimizing any calculations on its own (all redundant interfaces are still produced), but enables optimizations downstream by giving extra information. The client code must explicitly invoke this functionality and decide what to do with the chain/interface groups.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/StructureTools.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/StructureTools.java
@@ -2019,7 +2019,9 @@ public class StructureTools {
 	 * Expands the NCS operators in the given Structure adding new chains as needed.
 	 * The new chains are assigned ids of the form: original_chain_id+ncs_operator_index+"n"
 	 * @param structure
+	 * @deprecated use {@link org.biojava.nbio.structure.xtal.CrystalBuilder#expandNcsOps(Structure, Map, Map)} instead.
 	 */
+	@Deprecated
 	public static void expandNcsOps(Structure structure) {
 		PDBCrystallographicInfo xtalInfo = structure.getCrystallographicInfo();
 		if (xtalInfo ==null) return;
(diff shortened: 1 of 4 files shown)


biojava-structure/src/main/java/org/biojava/nbio/structure/StructureTools.java
```diff
@@ -2019,7 +2019,9 @@ public static void cleanUpAltLocs(Structure structure) {
 	 * Expands the NCS operators in the given Structure adding new chains as needed.
 	 * The new chains are assigned ids of the form: original_chain_id+ncs_operator_index+"n"
 	 * @param structure
+	 * @deprecated use {@link org.biojava.nbio.structure.xtal.CrystalBuilder#expandNcsOps(Structure, Map, Map)} instead.
 	 */
+	@Deprecated
```
josemduarte: There's no need to deprecate this one. It was an addition for 5.0, so it wasn't ever public yet. I'd say let's remove it.

biojava-structure/src/main/java/org/biojava/nbio/structure/contact/StructureInterfaceList.java
```diff
@@ -196,6 +189,62 @@ public void sort() {
...
 	}
 
+	/**
+	 * Get the interface clusters for this StructureInterfaceList grouped by NCS-equivalence.
+	 * This means that for any two interfaces in the same cluster:
+	 * 1. The chains forming the first interface are NCS-copies of the chains forming the second interface, in any order.
+	 * 2. Relative orientation of the chains is preserved, i.e. the contacts are identical.
+	 * @return list of {@link StructureInterfaceCluster} objects.
+	 */
+	public List<StructureInterfaceCluster> getClustersNcs() {
```
josemduarte: Could you add `@since 5.0.0` tags to this and any other new public methods?

biocryst:
Done, thanks!

PR #735, opened 2018-03-06
Fix for CRYST1 parsing issue

josemduarte:
Fixing PDB parsing issue. In some PDB files, valid CRYST1 records can be 58 columns only


biojava-structure/src/test/java/org/biojava/nbio/structure/io/TestNonDepositedFiles.java
```diff
@@ -408,6 +409,23 @@ public void testWaterOnlyChainCif() throws IOException {
...
 	}
+	
+	/**
+	 * Some PDB files coming from phenix or other software can have a CRYST1 line without z and not padded with white-spaces 
+	 * for the space group column.
+	 * @throws Exception
+	 * @since 5.0.0
+	 */
+	@Test
+	public void testCryst1Parsing() throws Exception {
```
biocryst: This can be specified to "throws IOException" only.

josemduarte:
Thanks, I've pushed a commit. Let's see if travis recovers now too.

josemduarte:
Tested, I'll merge

Issue #536, opened 2016-07-13
Merge Genomics and Sequencing modules


andreasprlic:
The genomics and sequencing modules are very similar. With #534  they now both depend on the Guava library. Is there any strong reason to keep them separated? I feel the overall organization would be simplified if the two modules would be merged.

josemduarte:
+1 for merging

heuermh:
Package and module names usually match.  Would this involve a package name change?

andreasprlic:
I'd say no package name change.

heuermh:
I'm +0 then, I think they're ok where they are but won't stand in the way of progress  :)

josemduarte:
Working on a pull request for this.

PR #736, opened 2018-03-07
Merging sequencing module into genome module 

josemduarte:
This solves #536. 

I did some warning cleanup as well.


biojava-genome/src/main/java/org/biojava/nbio/genome/io/fastq/AbstractFastqReader.java
```diff
@@ -18,7 +18,7 @@
  *      http://www.biojava.org/
  *
  */
-package org.biojava.nbio.sequencing.io.fastq;
+package org.biojava.nbio.genome.io.fastq;
```
heuermh: Should all these classes moved to a new package have `@since 5.0.0` tags?  Note there are other `@since` tags present on classes and methods throughout the old `o.b.n.sequencing.io.fastq` package.

biojava-genome/src/test/java/org/biojava/nbio/genome/io/fastq/AbstractFastqReaderTest.java
```diff
@@ -28,6 +28,11 @@
 import java.io.StringReader;
 import java.net.URL;
 
+import org.biojava.nbio.genome.io.fastq.Fastq;
```
heuermh: This doesn't seem right, why are these necessary?

biojava-genome/src/test/java/org/biojava/nbio/genome/io/fastq/AbstractFastqWriterTest.java
```diff
@@ -28,6 +28,9 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import org.biojava.nbio.genome.io.fastq.Fastq;
```
heuermh: ... similar here

biojava-genome/src/test/java/org/biojava/nbio/genome/io/fastq/ConvertTest.java
```diff
@@ -18,25 +18,35 @@
...
-package org.biojava.nbio.sequencing.io.fastq;
+package org.biojava.nbio.genome.io.fastq;
 
 import java.io.File;
 import java.io.FileWriter;
 import java.util.List;
 import java.util.Map;
 
-import junit.framework.TestCase;
+import org.biojava.nbio.genome.io.fastq.Fastq;
```
heuermh: ... did these all come in via an IDE?

biojava-genome/src/main/java/org/biojava/nbio/genome/io/fastq/AbstractFastqReader.java
josemduarte: Good point, I'm not sure how to do `@since` tags for these cases where something existed but was moved, rather than appeared new. Ideas anyone?

biojava-genome/src/test/java/org/biojava/nbio/genome/io/fastq/ConvertTest.java
josemduarte: Somehow eclipse considered them necessary... They are indeed redundant. I'll push a commit.

heuermh:
Ping me if this merges before #740; I'll rebase that pull request.

heuermh:
Note #740 went first, so the Guava dependency version in the genome module pom should also be bumped to `<version>24.0-jre</version>`, I didn't catch it there.

josemduarte:
Thanks @heuermh, I'll push a commit to solve the conflicts.

biojava-genome/src/main/java/org/biojava/nbio/genome/io/fastq/AbstractFastqReader.java
lafita: I would not use the `@since` tag, because it means that the class is new. Maybe just add a line like:

`moved from org.biojava.nbio.sequencing in 5.0.0`

biojava-genome/src/main/java/org/biojava/nbio/genome/util/ChromosomeMappingTools.java
```diff
@@ -610,7 +593,8 @@ public static int getCDSLengthForward(List<Integer> exonStarts, List<Integer> ex
         if ( debug)
             s = new StringBuffer();
 
-        int lengthExons = 0;
+        @SuppressWarnings("unused")
+		int lengthExons = 0;
```
lafita: Why is this variable not used? Just wondering if it would be better to remove it to eliminate confusion.

josemduarte:
Thanks for comments @lafita , I've pushed a commit addressing them.

josemduarte:
Tested, I'll merge

heuermh:
Thanks!

Issue #592, opened 2016-09-30
Merge AlignTools and AlignmentTools


lafita:
We have two different classes for **structure alignment methods**, with no apparent distinction between the methods in them. Is anyone against merging all the methods into a single class? The modification will be moving the methods in `AlignTools` to `AlignmentTools`, since it would require less changes.

sbliven:
This is pretty ancient history (predating the current module structure), so I'm having trouble remembering. I think the `structure.align.helper` package was created by Andreas and intended to port low-level procedures from CE/FATCAT from C to Java. The `structure.align.util` package was supposed to be more user-friendly, with code for post-processing AFPChains. I wrote most of the `AlignmentTools` code.

I would actually propose keeping them separate, but adding some docs to distinguish them. Maybe one should even be renamed to be less confusing? Or if we want to merge them, let's also merge the helper and util packages at the same time.

josemduarte:
Alright so it seems that it'd be better to keep them separate, possibly renaming and adding docs. I'd go for renaming `org.biojava.nbio.structure.align.helper.AlignTools` -> `org.biojava.nbio.structure.align.helper.AlignUtils`. I'll submit a PR...

PR #737, opened 2018-03-07
Refactoring and docs as per #592

josemduarte:
Solves #592


josemduarte:
Travis is having a lot of trouble still downloading external resources. I've tested this locally and it's fine. I'll merge.

PR #742, opened 2018-03-08
Avoiding creation of an empty jar file for integrationtest module

josemduarte:
More maven cleanup. There's no need to produce a jar file for integrationtest module. Maven was giving a warning for it.


josemduarte:
There were no comments here, anyway this was a minor cleanup. I'll merge it.

PR #743, opened 2018-03-09
Reset the QCP 2-step calculation

### LINKED ISSUE #725: Reset the QCP superposition method step-wise calculation
@pwrose reported this:

I was just trying to use the QCP superposition method in BioJava.  There is only one problem with the new design in that doesn't take advantage of QCP and will be slower in situations like the one below:

For example, I have the following code:

		SuperPositionQCP qcp = new SuperPositionQCP(false);
		double rmsd = qcp.getRmsd(x, y);
		double tm = 0.0;
		if (rmsd < maxRmsd) {
			qcp.superposeAndTransform(x, y);
			tm = TMScore(x, y);
		}

The getRmsd method does the majority of the work. But when superposeAndTransform(x, y) is run with the x,y arguments, it resets all intermediate results repeats the whole calculation from scratch.

In the original version (https://github.com/sbl-sdsc/QuickStructureSearch/blob/master/src/main/java/org/rcsb/structuralAlignment/SuperPositionQCP.java), you could get the RMSD first, then if you requested a superposition or the transformation matrix, it would use the intermediate results from the RMSD calculation to calculate the transformation matrix.

So QCP is conceptually different from the other superposition methods. I propose to add a QCP specific method: superposeAndTransform(), which would reuse the intermediate results from the RMSD calculation. The only downside is that it would fail if getRMSD(x, y) wasn't called before.

lafita:
Now the superposition can be calculated after the RMSD in a class
specific method.

I also added a test. This fixes #725

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/geometry/SuperPositionQCP.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/geometry/SuperPositionQCP.java
@@ -37,80 +37,79 @@ import org.slf4j.LoggerFactory;
  * <p>
  * The input consists of 2 Point3d arrays of equal length. The input coordinates
  * are not changed.
+ * 
  * <pre>
  *    Point3d[] x = ...
  *    Point3d[] y = ...
  *    SuperPositionQCP qcp = new SuperPositionQCP();
  *    qcp.set(x, y);
  * </pre>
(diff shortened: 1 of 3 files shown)


pwrose:
Looks good. Thank Aleix!

josemduarte:
Tested, I'll merge

Issue #594, opened 2016-10-14
protein-comparison-tool-5.0.0-SNAPSHOT -outputPDB error


leonqli:
I run the tool in the following way: 
./runCE.sh -pdb1 4hhb.A -pdb2 4hhb.B -autoFetch -pdbFilePath ~/pdbs/ -outputPDB  -outFile a.pdb

Thanks,
Leon 

Protein Comparison Tool 5.0.0-SNAPSHOT 20160324-1541
java.lang.NoSuchMethodException: org.biojava.nbio.structure.align.gui.DisplayAFP.getAtomArray([Lorg.biojava.nbio.structure.Atom;, java.util.List, java.util.List)
    at java.lang.Class.getMethod(Class.java:1786)
    at org.biojava.nbio.structure.align.ce.GuiWrapper.getAtomArray(GuiWrapper.java:190)
    at org.biojava.nbio.structure.align.util.AFPAlignmentDisplay.createArtificalStructure(AFPAlignmentDisplay.java:413)
    at org.biojava.nbio.structure.align.ce.AbstractUserArgumentProcessor.checkWriteFile(AbstractUserArgumentProcessor.java:591)
    at org.biojava.nbio.structure.align.ce.AbstractUserArgumentProcessor.runPairwise(AbstractUserArgumentProcessor.java:532)
    at org.biojava.nbio.structure.align.ce.AbstractUserArgumentProcessor.process(AbstractUserArgumentProcessor.java:197)
    at org.biojava.nbio.structure.align.ce.CeMain.main(CeMain.java:79)

josemduarte:
Thanks for the report! It looks like this is a bug in `org.biojava.nbio.structure.align.ce.GuiWrapper`, it tries to invoke an `org.biojava.nbio.structure.align.gui.StructureAlignmentDisplay` method (`prepareGroupsForDisplay`) that doesn't exist anymore. @lafita @sbliven could you guys have a look?

This highlights a problem in the design of the `GuiWrapper` class: it will not show errors at compile time after refactoring of method names. Could this be avoided somehow? Is there a reason to use reflection instead of direct calls here?

andreasprlic:
Reflection is used to work around maven dependency tree issues.

josemduarte:
[...]

PR #745, opened 2018-03-09
Fix the GuiWrapper class

lafita:
Fix for #594

The method `prepareGroupsForDisplay` had been moved to `AlignmentTools` in the `structure` module, but the references were not updated.


josemduarte:
Tested locally, I'll merge

PR #747, opened 2018-03-10
Standard LGPL license as provided by github template

josemduarte:
With this the license displays nicely in the top bar and a summary shown when opening the LICENSE file

### DIFF
--- a/HEADER.txt
+++ b/HEADER.txt
@@ -1,0 +1,20 @@
+/*
+ *                    BioJava development code
+ *
+ * This code may be freely distributed and modified under the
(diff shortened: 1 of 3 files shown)


heuermh:
I also find a [separate `HEADER.txt`](https://github.com/heuermh/dishevelled-bio/blob/master/HEADER.txt) to be useful.  Then checks can be added to the build using the maven-checkstyle-plugin or maven-license-plugin.

josemduarte:
Greate idea @heuermh , I've placed our header text in a separate HEADER.txt file and made the addlicense.sh script read from it, just pushed the commit.

PR #749, opened 2018-03-12
Removed v4 deprecations

josemduarte:
A further cleanup for the 5 release. Removed anything that was deprecated in v3->v4 or before, checked by comparing to 4 branch. Kept everything that was deprecated in v4->v5.


josemduarte:
Thanks @lafita , I've tested it locally and it worked. I'll merge it in later today to leave more time for comments.

PR #752, opened 2018-03-14
Update database versions

sbliven:
Add latest database versions for SCOP, CATH, and ECOD.

PR #753, opened 2018-03-15
Update ECOD test

sbliven:
This updates the ECOD tests to the latest version. The parser works fine, this just updates the test fixture to use more recent data. Since it doesn't have any public code, it can be included in either 5.0.0 or 5.0.1.

This probably should have been a part of #752 but I wanted to get the main changes out.


sbliven:
Tests pass locally.

PR #756, opened 2018-03-15
Converting junit3 to 4, thanks intellij

josemduarte:
FYI, I'll merge in a while


josemduarte:
@heuermh  I'd be happy to do it if intellij can do it :) Do you know an intellij shortcut for that?

heuermh:
> Do you know an intellij shortcut for that?

I don't know, but it appears that intellij knew how to do it sometimes.  :)

josemduarte:
It seems that optimize imports won't do it. It rather tries the opposite: changing wildcard to explicit. We'll have to leave this for next release.

emckee2006:
The opposite is better, isn't it? That way you are explicitly listing only
the things you need...

lafita:
@heuermh why is the static wildcard import better?

heuermh:
I most prefer explicit static imports, no matter how many
```java
import static org.junit.Assert.assertNotNull;
import static org.junit.Assert.assertNull;
import org.junit.Test;

//...
@Test
public void testSomething() {
  assertNull(null);
  assertNotNull(something);
}
```

but would still prefer static wildcard imports
```java
import static org.junit.Assert.*;
import org.junit.Test;

//...
@Test
public void testSomething() {
  assertNull(null);
  assertNotNull(something);
}
```

over non-static imports and qualified assert statements
```java
import org.junit.Assert;
import org.junit.Test;

//...
@Test
public void testSomething() {
  Assert.assertNull(null);
  Assert.assertNotNull(something);
}
```

Issue #574, opened 2016-09-06
Port org.biojava.bio.program.abi package from legacy biojava to current biojava series


andreasprlic:
Feature request from Scott Markel: 

Would you please have a look at porting the ABITrace class to BioJava 4.2?  The BioJava 1.9.1 class path is org.biojava.bio.program.abi.ABITrace.

see also #535

el1jaC:
Hello @andreasprlic

I've tried test file with our internal analysis software, and it works well :)

It can be a good starting point to trivially copy the classes, isn't it? And then we can think to refactor some code to integrate it in new biojava repo.

MaxGreil:
I can have a look at it.

josemduarte:
Indeed, this will go released with 5.1.0

Issue #535, opened 2016-07-13
Add support for parsing Sanger sequencing files (ABI, SCF)


andreasprlic:
see this thread on the mailing list

http://lists.open-bio.org/pipermail/biojava-l/2016-July/011489.html

A test file is here

https://github.com/biopython/biopython/blob/master/Tests/Abi/3730.ab1

andreasprlic:
see also #574

el1jaC:
Hello @andreasprlic 

I've tried test file with our internal analysis software, and it works well :)

It can be a good starting point to trivially copy the classes, isn't it?

PR #769, opened 2018-05-06
#574 Port org.biojava.bio.program.abi.ABITracer

MaxGreil:
#574 
Added 3 new classes: `ABITrace.java`, `ABITracerTest.java` and `ABITracerCompoundSet.java`.
The test file `3730.ab1` is from https://github.com/biopython/biopython/blob/master/Tests/Abi/3730.ab1.
This file was mentioned in #535.
The test data for file `3730.abi1` is from https://github.com/biopython/biopython/blob/master/Tests/Abi/test_data.


biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/ABITrace.java
```diff
@@ -0,0 +1,568 @@
...
+ * ABI files contain two sets of basecall and sequence data, one that was originally
+ * created programatically and the other, which is an editable copy. This version of this object
+ * only references the original unedited data.<br>
+ */
+public class ABITrace {
+
+    //the next three lines are the important persistent data
+    private String sequence;
+    private int A[], G[], C[], T[], Basecalls[], Qcalls[];
+    private int TraceLength, SeqLength;
```
josemduarte: Variables should start with lower case as per normal java conventions.

I know this was like that in legacy project, but it'd be a great opportunity to fix the style while porting it :)

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/ABITrace.java
```diff
@@ -0,0 +1,568 @@
...
+import java.io.InputStream;
+
+import org.biojava.nbio.core.sequence.compound.ABITracerCompoundSet;
+import org.biojava.nbio.core.sequence.compound.DNACompoundSet;
+import org.biojava.nbio.core.sequence.compound.NucleotideCompound;
+import org.biojava.nbio.core.sequence.template.AbstractSequence;
+import org.biojava.nbio.core.exceptions.CompoundNotFoundException;
+
+/**
+ * Title: ABITrace<br><br>
```
josemduarte: Paragraph separator should be `<P>` instead of `<br>`, see https://stackoverflow.com/questions/5260368/which-tag-should-be-used-as-paragraph-separator-in-javadoc?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/ABITrace.java
```diff
@@ -0,0 +1,568 @@
...
+
+    private int maximum = 0;
+
+    //the next four declaration lines comprise the file index information
+    private int MacJunk = 0; //sometimes when macintosh files are
+    //FTPed in binary form, they have 128 bytes
+    //of crap pre-pended to them. This constant
+    //allows ABITrace to handle that in a way that
+    //is invisible to the user.
+    private static int AbsIndexBase = 26; //The file location of the Index pointer
```
josemduarte: Constants should be `static final`

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/ABITrace.java
```diff
@@ -0,0 +1,568 @@
...
+        initData(bytes);
+    }
+
+    /**
+     * The URL constructor opens an ABI file from any URL.
+     * @param ABIFile is a <code>java.net.URL</code> for an ABI trace file.
+     * @throws IOException if there is a problem reading from the URL.
+     * @throws IllegalArgumentException if the URL does not contain a valid ABI file.
+     */
+    public ABITrace( URL ABIFile ) throws IOException
```
josemduarte: This and the method above are very similar. It'd be nice to extract the common part to a separate method to be shared by both.

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/ABITrace.java
```diff
@@ -0,0 +1,568 @@
...
+    }
+
+    /**
+     * Returns the original programmatically determined (unedited) sequence as a <code>AbstractSequence<NucleotideCompound></code>.
+     */
+    public AbstractSequence<NucleotideCompound> getSequence() throws CompoundNotFoundException {
+        try {
+            DNASequenceCreator creator = new DNASequenceCreator(ABITracerCompoundSet.getABITracerCompoundSet());
+            return creator.getSequence(sequence, 0);
+        } catch (Exception e) {
```
josemduarte: No need to catch, just let the `CompoundNotFoundException` go through.

Catching a general `Exception` here could cause undesired side effects in the error handling.

If you think the `CompoundNotFoundException` can't happen (other than because of a bug), then it's valid to catch it (but only that exception), log a warning and not throw it.

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/ABITrace.java
```diff
@@ -0,0 +1,568 @@
...
+     * @throws IllegalArgumentException if the data does not represent a valid ABI file.
+     */
+    public ABITrace(byte[] ABIFileData) {
+        initData(ABIFileData);
+    }
+
+    /**
+     * Returns the length of the sequence (number of bases) in this trace.
+     */
+    public int getSequenceLength() {
```
josemduarte: The `@return` tags are missing in a few javadocs

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/ABITrace.java
```diff
@@ -0,0 +1,568 @@
...
+import java.io.DataInputStream;
+import java.io.ByteArrayInputStream;
+import java.awt.Color;
+import java.awt.Graphics2D;
+import java.awt.image.BufferedImage;
+import java.net.URL;
+import java.io.InputStream;
+
+import org.biojava.nbio.core.sequence.compound.ABITracerCompoundSet;
+import org.biojava.nbio.core.sequence.compound.DNACompoundSet;
```
josemduarte: Unused import

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/ABITrace.java
```diff
@@ -0,0 +1,568 @@
...
+    }
+
+    //calculates the necessary scaling to allow the trace to fit vertically
+    //in the space specified.
+
+    /**
+     * Returns the scaling factor necessary to allow all of the traces to fit vertically
+     * into the specified space.
+     *
+     * @param - the required height in pixels.
```
josemduarte: Missing param name

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/ABITrace.java
```diff
@@ -0,0 +1,568 @@
...
+        PCONCounter = 0;
+
+        IndexBase = getIntAt(AbsIndexBase + MacJunk);
+        NumRecords = getIntAt(AbsIndexBase - 8 + MacJunk);
+
+        for (int record = 0; record <= NumRecords - 1; record++) {
+            getSubArray(RecNameArray, (IndexBase + (record * 28)));
+            RecName = new String(RecNameArray);
+            if (RecName.equals("FWO_"))
+                FWO = IndexBase + (record * 28) + 20;
```
josemduarte: Any idea what this 2 "magic" numbers mean? It'd be nice to extract them to constants with some documentation as to what they mean.

biojava-core/src/test/java/org/biojava/nbio/core/sequence/io/ABITracerTest.java
```diff
@@ -0,0 +1,97 @@
...
+import java.net.URL;
+import org.junit.*;
+
+/**
+ * Test file 3730.ab1 is from https://github.com/biopython/biopython/blob/master/Tests/Abi/3730.ab1
+ * The test data for comparing the results from ABITracer for file 3730.ab1 is from https://github.com/biopython/biopython/blob/master/Tests/Abi/test_data
+ */
+public class ABITracerTest {
+
+    private String filePath = System.getProperty("user.dir") + "/src/test/resources/3730.ab1";
```
josemduarte: This is not going to work when running as a jar. See how to do it here: https://stackoverflow.com/questions/3891375/how-to-read-a-text-file-resource-into-java-unit-test?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

MaxGreil: I have a question regarding this. 
Is it possible to add the dependency for org.apache.commons.io.IOUtils? 
Its function IOUtils can be used to safely convert the InputStream to a byte[] and this can then be used for the Constructor ABITracer (byte[] ABIFileData).

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/ABITrace.java
MaxGreil: I don't know what these 2 numbers stand for. What would be the best names to call them?

josemduarte: I'm not familiar with the format either. They seem some kind of offset, but I can't really tell.

FYI the format is described http://www6.appliedbiosystems.com/support/software_community/ABIF_File_Format.pdf

Let's leave it as is then... unless there's other suggestions, @el1jaC ?

MaxGreil:
@josemduarte 
* Refactored constructors `ABITrace(File ABIFile)` and `ABITrace( URL ABIFile )` with new shared method `private void ABITraceInit(BufferedInputStream bis)`. 
* Corrected all other requests. 
* Left request for method `private void setIndex()`.

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/ABITrace.java
```diff
@@ -0,0 +1,597 @@
...
+    private int maximum = 0;
+
+    //the next four declaration lines comprise the file index information
+    private int macJunk = 0; //sometimes when macintosh files are
+    //FTPed in binary form, they have 128 bytes
+    //of crap pre-pended to them. This constant
+    //allows ABITrace to handle that in a way that
+    //is invisible to the user.
+    private static final int absIndexBase = 26; //The file location of the Index pointer
+    private int IndexBase, PLOC, PCON;
```
josemduarte: IndexBase should be lower case, plus the variable should be local to the method where it is used.

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/ABITrace.java
```diff
@@ -0,0 +1,597 @@
...
+
+    //the next three lines are the important persistent data
+    private String sequence;
+    private int A[], G[], C[], T[], baseCalls[], qCalls[];
+    private int traceLength, seqLength;
+
+    //This is the actual file data.
+    private byte[] traceData;
+
+    private int maximum = 0;
```
josemduarte: Doesn't look that this is being used. I seems to be a caching variable but not used properly. I'd remove it.

MaxGreil:
Fixed minor changes.

MaxGreil:
@josemduarte 
Thank you. Is the PR now ready for merge with the master branch?

josemduarte:
I'd like to leave it open for a couple of days in case there are any other comments.

el1jaC:
@josemduarte sorry for the late, I didn't see the tag!

Btw, I'm very happy that classes are ported to new Biojava.

josemduarte:
> When the pull request will be merged, can I add support in ABITrace class to getImageReverted, to manage RC sequences?

Absolutely, more than welcomed!

josemduarte:
@el1jaC please go ahead and submit the new pull request when you are ready. This one is merged.

el1jaC:
Thank you  @josemduarte .

I'll do this in the next days!

Issue #731, opened 2018-02-27
OBO Parser: replaced_by


PatrickDeelen:
I'm trying to parse the HPO OBO file. Unfortunately the `replaced_by` triple is not parsed. It would be great if you can add this in a future version.

josemduarte:
Good suggestion. I've added it to 5.1.0 milestone. Note that 5.0.0 is going to be released soon, see #727

PatrickDeelen:
Great, thank you!

MaxGreil:
I can have a look at it.

josemduarte:
Now implemented with #731 . Will be available in next release 5.1.0

PR #773, opened 2018-05-26
#731 OBO Parser: replaced_by

MaxGreil:
#731 
- Modified `OBOFileHandler.java` for triple `replaced_by`
- Added new file `hp.obo` (HPO OBO file) and added new test in `TestOBOFileParsing.java` to test new functionality for HPO OBO file

### DIFF
--- a/biojava-ontology/src/main/java/org/biojava/nbio/ontology/obo/OboFileHandler.java
+++ b/biojava-ontology/src/main/java/org/biojava/nbio/ontology/obo/OboFileHandler.java
@@ -136,6 +136,10 @@ public class OboFileHandler implements OboFileEventListener {
 				logger.warn("did not find ID for Term! ");
 				return;
 			}
+			if (key.equals(NAMESPACE)){
+				Annotation anno = currentTerm.getAnnotation();
+				anno.setProperty(NAMESPACE, value);
+			}
 			else if (key.equals(NAME)){
 				currentTerm.setDescription(value);
 			} else if (key.equals(DEF)){
@@ -172,10 +176,6 @@ public class OboFileHandler implements OboFileEventListener {
 			} else if (key.equals(ALT_ID)){
 				Annotation anno = currentTerm.getAnnotation();
 				anno.setProperty(ALT_ID, value);
-			}
-			else if (key.equals(NAMESPACE)){
-				Annotation anno = currentTerm.getAnnotation();
-				anno.setProperty(NAMESPACE, value);
 			}
 			else if (key.equals(REPLACED_BY)) {
 				Annotation anno = currentTerm.getAnnotation();
--- a/biojava-ontology/src/main/java/org/biojava/nbio/ontology/obo/OboFileHandler.java
+++ b/biojava-ontology/src/main/java/org/biojava/nbio/ontology/obo/OboFileHandler.java
@@ -66,6 +66,7 @@ public class OboFileHandler implements OboFileEventListener {
 	public static final String SUBSET       = "subset";
 	public static final String INTERSECTION_OF = "intersection_of";
 	public static final String NAMESPACE = "namespace";
+	public static final String REPLACED_BY = "replaced_by";
 
 
 	public static final String ALT_ID      = "alt_id";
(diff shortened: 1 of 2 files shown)


biojava-ontology/src/test/java/org/biojava/nbio/ontology/TestOboFileParsing.java
```diff
@@ -37,23 +37,40 @@
...
 
-		Assert.assertTrue(keys.size() > 4000);
-	}
+        Assert.assertTrue(keys.size() > 4000);
+    }
+
+    @Test
+    public void testParsingHPOOBO() throws Exception {
+        OboParser parser = new OboParser();
+        InputStream inStream = parser.getClass().getResourceAsStream("/ontology/hp.obo");
```
josemduarte: If you gzip the hp.obo file, here you just need to use the gzip filter to read it, i.e.
```java
InputStream inStream = new GZIPInputStream(parser.getClass().getResourceAsStream("/ontology/hp.obo"));
```

biojava-ontology/src/test/java/org/biojava/nbio/ontology/TestOboFileParsing.java
```diff
@@ -37,23 +37,40 @@
...
+        Assert.assertNotNull(inStream);
+
+        BufferedReader oboFile = new BufferedReader ( new InputStreamReader ( inStream ) );
+
+        Ontology ontology;
+
+        ontology = parser.parseOBO(oboFile, "Human_phenotype", "the Human Phenotype ontology");
+        Set<Term> keys = ontology.getTerms();
+
+        Assert.assertTrue(keys.size() > 4000);
```
josemduarte: Could you also assert that the new replaced_by field is parsed properly?

heuermh:
> As the new test resource file is quite big I would suggest to gzip it.

It might be better to create an extract of the original file and store it in git uncompressed, that way changes can be tracked with regular line diffs.

josemduarte:
Agree @heuermh , that's a better solution

MaxGreil:
The first appearance of a `replaced_by` field is at line 574. Would it be alright if I then just extract the first 574 lines of the `hp.obo` file?

josemduarte:
@MaxGreil sounds good as long as the file stays consistent and parseable. Thanks!

MaxGreil:
Implemented requested changes:
- Extracted first 574 lines of the original `hp.obo` file

biojava-ontology/src/main/java/org/biojava/nbio/ontology/obo/OboFileHandler.java
```diff
@@ -172,10 +173,14 @@ else if (key.equals(NAME)){
 				Annotation anno = currentTerm.getAnnotation();
 				anno.setProperty(ALT_ID, value);
 			}
-            else if (key.equals(NAMESPACE)){
-                Annotation anno = currentTerm.getAnnotation();
-                anno.setProperty(NAMESPACE, value);
-            }
+			else if (key.equals(NAMESPACE)){
+				Annotation anno = currentTerm.getAnnotation();
```
heuermh: Minor style fix, could this line be brought outside the `if/else if` block?

biojava-ontology/src/test/java/org/biojava/nbio/ontology/TestOboFileParsing.java
```diff
@@ -34,26 +34,52 @@
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.util.Set;
+import java.util.Iterator;
+import org.biojava.nbio.ontology.utils.Annotation;
 
 public class TestOboFileParsing {
 
-	private static final Logger logger = LoggerFactory.getLogger(TestOboFileParsing.class);
+    private static final Logger logger = LoggerFactory.getLogger(TestOboFileParsing.class);
```
heuermh: I'm afraid all of the biojava source should use tabs (I lost that argument a while back :wink:)

MaxGreil:
Fixed style suggestions.

MaxGreil:
Fixed test.

MaxGreil:
@josemduarte there is a new compilation error in the `org.biojava.nbio.structure.tests` and that is weird because I didn't touch any of those files and tests in my last commit

josemduarte:
Yes the tests are still not stable enough. I've restarted the build, hopefully this time it will work.

josemduarte:
Ok finally passing. I'll squash commits and merge. Thanks again!

PR #774, opened 2018-06-08
Fix #703: Recover from empty structure files in PDB_CACHE_DIR

### LINKED ISSUE #703: Chemical component cache poisoning
In cases where there's problems downloading the chemical component files from PDB, corrupted or empty files are still saved producing difficult to understand errors down the line (see #702). Two things to fix:

- do not store files if the download is empty or any kind of server error (404 or others)
- better error handling when the file doesn't have valid content, trying a redownload

sbliven:
For a short period the ChemComp files didn't resolve, resulting in empty files getting cached by BioJava. The caching problem has been long fixed, but this PR should make it so BioJava can recover and if the user happened to use an old version of BioJava (e.g. older CE-Symm versions).

Fixes #703


biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/DownloadChemCompProvider.java
```diff
@@ -272,7 +275,10 @@ public  ChemComp getChemComp(String recordName) {
 
 				ChemComp chemComp = dict.getChemComp(recordName);
 
-				return chemComp;
+				// May be null if the file was corrupt. Fall back on ReducedChemCompProvider in that case
+				if(chemComp != null) {
+					return chemComp;
```
sbliven: This is a small change in behavior. If we can't read the downloaded chemcomp file, we now return a stub chemcomp rather than NULL. Previously the stub was returned for corrupt GZIP files (which threw an error) but not for malformed cif contents (which just returned null).

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/DownloadChemCompProvider.java
```diff
@@ -296,9 +302,11 @@ public  ChemComp getChemComp(String recordName) {
 
 		// see https://github.com/biojava/biojava/issues/315
 		// probably a network error happened. Try to use the ReducedChemCOmpProvider
-		ReducedChemCompProvider reduced = new ReducedChemCompProvider();
+		if( fallback == null) {
+			fallback = new ReducedChemCompProvider();
```
sbliven: creating a ReducedChemCompProvider is cheap (and idempotent), but let's cache it anyways.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/util/FileDownloadUtils.java
```diff
@@ -240,6 +246,41 @@ public static URLConnection prepareURLConnection(String url, int timeout) throws
 		connection.setConnectTimeout(timeout);
 		return connection;
 	}
+	
+	/**
+	 * Recursively delete a folder & contents
+	 *
+	 * @param dir directory to delete
+	 */
+	public static void deleteDirectory(Path dir) throws IOException {
```
sbliven: Surprisingly this isn't an nio method and I couldn't find one in BioJava

biojava-structure/src/test/java/org/biojava/nbio/structure/align/util/AtomCacheTest.java
```diff
@@ -312,4 +350,125 @@ public void testSeqRes() throws StructureException, IOException {
...
+		System.setProperty(UserConfiguration.PDB_DIR, tmpCache.toString());
+		System.setProperty(UserConfiguration.PDB_CACHE_DIR, tmpCache.toString());
+
+		FileDownloadUtils.deleteDirectory(tmpCache);
+		Files.createDirectory(tmpCache);
+		try {
+			cache.setPath(tmpCache.toString());
+			cache.setCachePath(tmpCache.toString());
+			cache.setUseMmCif(true);
+			ChemCompGroupFactory.setChemCompProvider(new DownloadChemCompProvider(tmpCache.toString()));
```
sbliven: It's annoying how much code is needed to change the cache dir.

sbliven:
Tests pass locally, but there's some danger of system-dependent bugs here since I'm hacking the cache path for the two new tests.

BTW, this could be an example of how to use AtomCache in tests without incurring network access.

sbliven:
This should be merged to master after merging

biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathInstallation.java
```diff
@@ -636,6 +636,7 @@ private void parseCathDomall(BufferedReader bufferedReader) throws IOException{
 
 	protected void downloadFileFromRemote(URL remoteURL, File localFile) throws IOException{
 //        System.out.println("downloading " + remoteURL + " to: " + localFile);
+		LOGGER.info("Downloaded file {} to local file {}", remoteURL, localFile);
```
josemduarte: Shouldn't this be "Downloading"? It isn't downloaded at this point yet

biojava-structure/src/main/java/org/biojava/nbio/structure/io/LocalPDBDirectory.java
```diff
@@ -687,6 +690,14 @@ public File getLocalFile(String pdbId) {
 				for(String ex : getExtensions() ){
 					File f = new File(searchdir,prefix + pdbId.toLowerCase() + ex) ;
 					if ( f.exists()) {
+						// delete files that are too short to have contents
+						if( f.length() < MIN_PDB_FILE_SIZE ) {
+							boolean success = f.delete();
+							if( ! success) {
+								return null;
+							}
+							assert(!f.exists());
```
josemduarte: I'd rather log a warning here. Assert is going to be ignored in normal situations

biojava-structure/src/main/java/org/biojava/nbio/structure/io/LocalPDBDirectory.java
```diff
@@ -687,6 +690,14 @@ public File getLocalFile(String pdbId) {
 				for(String ex : getExtensions() ){
 					File f = new File(searchdir,prefix + pdbId.toLowerCase() + ex) ;
 					if ( f.exists()) {
+						// delete files that are too short to have contents
+						if( f.length() < MIN_PDB_FILE_SIZE ) {
+							boolean success = f.delete();
+							if( ! success) {
+								return null;
```
josemduarte: Shouldn't this rather throw an exception (perhaps IOException)? I don't see how can this null be handled if file can't be deleted.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/ChemCompGroupFactory.java
```diff
@@ -84,6 +83,13 @@ public static void setChemCompProvider(ChemCompProvider provider) {
 	public static ChemCompProvider getChemCompProvider(){
 		return chemCompProvider;
 	}
+	
+	/**
+	 * Force the cache to be reset.
```
josemduarte: This clears the memory cache, right? Could you add that to javadoc, it's an important detail since there is both memory and file cache

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/DownloadChemCompProvider.java
```diff
@@ -337,6 +345,12 @@ private static  boolean fileExists(String recordName){
 
 		File f = new File(fileName);
 
+		// delete files that are too short to have contents
+		if( f.length() < LocalPDBDirectory.MIN_PDB_FILE_SIZE ) {
+			f.delete();
+			return false;
```
josemduarte: If the file fails to be deleted, this won't work as expected. I think the fail to delete case should throw an IOException

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/DownloadChemCompProvider.java
```diff
@@ -296,9 +302,11 @@ public  ChemComp getChemComp(String recordName) {
 
 		// see https://github.com/biojava/biojava/issues/315
 		// probably a network error happened. Try to use the ReducedChemCOmpProvider
-		ReducedChemCompProvider reduced = new ReducedChemCompProvider();
+		if( fallback == null) {
+			fallback = new ReducedChemCompProvider();
+		}
 
-		return reduced.getChemComp(recordName);
+		return fallback.getChemComp(recordName);
```
josemduarte: Is there a warning logged when the fallback is used? That'd be important, so that users are aware and can investigate if there's something wrong in their code or file system

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/SimpleMMcifConsumer.java
```diff
@@ -328,7 +328,13 @@ public void newAtomSite(AtomSite atom) {
 
 		String recordName    = atom.getGroup_PDB();
 		String residueNumberS = atom.getAuth_seq_id();
-		Integer residueNrInt = Integer.parseInt(residueNumberS);
+		Integer residueNrInt;
+		if(residueNumberS != null) {
+			residueNrInt = Integer.parseInt(residueNumberS);
+		} else {
+			String label_seq_id = atom.getLabel_seq_id();
+			residueNrInt = Integer.parseInt(label_seq_id);
```
josemduarte: `label_seq_id` is not the same as residue number. This can introduce many problems.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/SimpleMMcifParser.java
```diff
@@ -212,6 +212,9 @@ public void parse(BufferedReader buf)
 
 		// the first line is a data_PDBCODE line, test if this looks like a mmcif file
 		line = buf.readLine();
+		while( line != null && (line.isEmpty() || line.startsWith(COMMENT_CHAR))) {
+			line = buf.readLine();
+		}
```
josemduarte: Nice to handle comments in first line! 

Is an empty first line also valid CIF?

biojava-structure/src/test/java/org/biojava/nbio/structure/align/util/AtomCacheTest.java
```diff
@@ -312,4 +350,125 @@ public void testSeqRes() throws StructureException, IOException {
...
+			Group g = s.getChainByPDB("A").getAtomGroup(0);
+			assertTrue(g.getPDBName().equals("ATP"));
+			
+			// should be unknown
+			ChemComp chem = g.getChemComp();
+			assertNotNull(chem);
+			assertTrue(chem.getAtoms().size() > 0);
+			assertEquals("NON-POLYMER", chem.getType());
+		} finally {
+//			FileDownloadUtils.deleteDirectory(tmpCache);
```
josemduarte: Why commented? Is cleaning up not needed here? 

If not needed please remove the try/finally

biojava-structure/src/test/java/org/biojava/nbio/structure/align/util/AtomCacheTest.java
```diff
@@ -312,4 +350,125 @@ public void testSeqRes() throws StructureException, IOException {
...
```
josemduarte: Why commented?

biojava-structure/src/main/java/org/biojava/nbio/structure/io/LocalPDBDirectory.java
sbliven: That's intended. If the delete was successful then the file will not exist. This is just a sanity check during development to make sure delete() wasn't asynchronous or something.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/DownloadChemCompProvider.java
sbliven: I don't check the status because the delete isn't really necessary here. If this method returns false then we re-download the file and move it on top of the old one. I just added the delete defensively.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/DownloadChemCompProvider.java
sbliven: Good idea.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/SimpleMMcifParser.java
sbliven: I think blank lines are permitted anywhere in CIF, but I could be wrong. I try to write permissive parsers. It seems like it shouldn't break the spec so we might as well be robust to it either way.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/SimpleMMcifConsumer.java
sbliven: We should discuss this further.

I added it because the atp.cif.gz file I added (from pymol) has label_seq_id but not auth_seq_id. It seems reasonable to me to fall back on the sequential numbering if the authors don't specify a custom numbering.

Note that `auth_seq_id` is optional but `label_seq_id` is required. From the spec it sounds to me like `label_seq_id` is a good fallback for ResidueNumber:

[_atom_site.auth_seq_id:](http://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_atom_site.auth_seq_id.html)
> An alternative identifier for _atom_site.label_seq_id that
> may be provided by an author in order to match the identification
> used in the publication that describes the structure.

Ideally BioJava would use `Group.getId()` rather then `getResidueNumber()` everywhere. However, many things still use residue numbers, so setting a default seems prudent.

sbliven: There's another potential issue here, which is that the seq_id is stored as a `long` while ResidueNumber contains an `Integer`. I don't think there are any files with >2billion groups per entity, but if we hit it then the ResidueNumber would throw a NumberFormatException here. I think that's fine.

biojava-structure/src/test/java/org/biojava/nbio/structure/align/util/AtomCacheTest.java
sbliven: Fixed. Cleaning up /tmp is polite, but makes it harder to debug failing tests.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/SimpleMMcifConsumer.java
sbliven: I started #775 to continue this discussion.

biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathInstallation.java
sbliven: fixed

biojava-structure/src/main/java/org/biojava/nbio/structure/io/LocalPDBDirectory.java
sbliven: switched to `Files.delete` which throws the exception

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/ChemCompGroupFactory.java
sbliven: Only memory. Documented.

biojava-structure/src/test/java/org/biojava/nbio/structure/align/util/AtomCacheTest.java
sbliven: fixed

sbliven:
@josemduarte  please check again and see if I addressed all your suggestions.

The `GlobalsHelper` class that I added is pretty clean, and it would be good to use it for more tests that hack paths and factory methods.

sbliven:
Tests pass with an old cache, but 4LNC got updated Tuesday and now `TestExperimentalTechniques.test4LNC` fails on Travis.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/SimpleMMcifConsumer.java
```diff
@@ -328,7 +328,13 @@ public void newAtomSite(AtomSite atom) {
 
 		String recordName    = atom.getGroup_PDB();
 		String residueNumberS = atom.getAuth_seq_id();
-		Integer residueNrInt = Integer.parseInt(residueNumberS);
+		Integer residueNrInt;
```
josemduarte: Could you handle this in a separate pull request? as per discussion in #775

sbliven:
OK, I removed the mmcif changes (#775) and merged the result.

Issue #778, opened 2018-06-24
Altloc groups can be duplicated when using toMmcif()


josemduarte:
The altloc groups can have duplicate atoms with the main group, since at parsing we make sure that all groups contain all the default location atoms. When writing out to mmcif we are not considering that and writing out duplicates.

PR #779, opened 2018-06-25
Fixing an altloc issue

josemduarte:
This adds tests and a fix for #778

It also clarifies how altlocs are handled by BioJava by adding docs in a few places.

A couple of issues found in testAltLocs are also fixed.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/Group.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/Group.java
@@ -64,13 +64,14 @@ public interface Group extends Serializable {
 	public int size();
 
 	/**
-	 *  Return true or false, depending if this group has 3D coordinates or not.
+	 * Return true or false, depending if this group has 3D coordinates or not.
 	 *
 	 * @return true if Group has 3D coordinates
 	 */
 	public boolean has3D ();
 
-	/** Flag if group has 3D data .
+	/**
+	 * Flag if group has 3D data .
 	 *
 	 * @param flag  true to set flag that this Group has 3D coordinates
 	 */
@@ -84,13 +85,15 @@ public interface Group extends Serializable {
 	 */
 	public GroupType getType();
 
-	/** Add an atom to this group.
+	/**
+	 * Add an atom to this group.
 	 *
 	 * @param atom  an Atom object
 	 */
 	public void addAtom(Atom atom);
 
-	/** Get list of atoms.
+	/**
+	 * Get list of atoms.
 	 *
 	 * @return a List object representing the atoms
 	 * @see #setAtoms(List)
@@ -98,7 +101,8 @@ public interface Group extends Serializable {
 	public List<Atom> getAtoms() ;
 
 
-	/** Set the atoms of this group.
+	/**
+	 * Set the atoms of this group.
 	 * @see {@link Atom}
 	 * @param atoms a list of atoms
 	 */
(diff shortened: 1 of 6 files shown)


biojava-structure/src/test/java/org/biojava/nbio/structure/TestAltLocs.java
```diff
@@ -208,15 +209,20 @@ private void ensureAllAtomsSameAltCode(Group groupInputAltLocGroup, Group inputM
 	@Test
 	public void test1AAC() throws IOException, StructureException{
 
+		AtomCache cache = new AtomCache();
+		cache.setUseMmCif(false);
+		StructureIO.setAtomCache(cache);
```
sbliven: This is a good case for my new `GlobalsHelper` class. Unfortunately I haven't gotten around to merging it from 4.2 into master yet. (assignto: @sbliven)

biojava-structure/src/test/java/org/biojava/nbio/structure/TestAltLocs.java
```diff
@@ -604,6 +593,164 @@ public void test3U7Tmmcif() throws IOException, StructureException{
...
+				assertEquals('A', a.getAltLoc().charValue());
+		}
+
+		assertEquals(11, g.getAltLocs().get(0).size());
+		for (Atom a : g.getAltLocs().get(0).getAtoms()) {
+			if (a.getName().equals("C") || a.getName().equals("N") || a.getName().equals("O") || a.getName().equals("CA") || a.getName().equals("CB"))
+				assertEquals(' ', a.getAltLoc().charValue());
+			else
+				assertEquals('A', a.getAltLoc().charValue());
+		}
```
sbliven: So does the first altloc contain the same atoms as the main group? Even stronger, is `this == this.getAltLocs().get(0)`?

biojava-structure/src/test/java/org/biojava/nbio/structure/TestAltLocs.java
```diff
@@ -604,6 +593,164 @@ public void test3U7Tmmcif() throws IOException, StructureException{
...
+				assertEquals('B', a.getAltLoc().charValue());
+		}
+
+		List<AtomSite> atomSites = MMCIFFileTools.convertChainToAtomSites(c, 1, "A", "A");
+		assertEquals(17, atomSites.size());
+
+	}
+
+	@Test
+	public void testMmcifWritingAllAltlocs() throws IOException {
```
sbliven: This doesn't actually test mmcif writing. It seems like it would be good to test `Structure.toMMCif()` directly.

biojava-structure/src/main/java/org/biojava/nbio/structure/Group.java
```diff
@@ -301,41 +314,66 @@
...
+	 * @see #getAltLocs()
 	 */
 	public boolean hasAltLoc();
 
 
-	/** Get the list of alternate locations.
+	/**
+	 * Get the list of other alternate location groups.
+	 * <p>
+	 * The main group (this group) will contain the first altloc (be it the default '.' or 'A' or a mix of '.' and 'A').
```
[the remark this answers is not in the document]
sbliven: After reading the tests I see what you meant here. Maybe the documentation could be a bit clearer. Something like:

> Atoms with the default altloc (`.`) are included in all groups. Atoms with other altlocs (typically `A`, `B`, etc) will be sorted into groups by altloc. This method returns a list of all alternative groups, starting with the main group.

Question: what happens if you have the same atom name with both `.` and `A`? Is that an invalid PDBx?

Question: do groups share an instance for "duplicated" Atoms, or will they have to be cloned?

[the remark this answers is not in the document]
josemduarte: Thanks for feedback @sbliven . I'll add some commits including the docs suggestion.

> Question: what happens if you have the same atom name with both . and A? Is that an invalid PDBx?

I think that's valid.

> Question: do groups share an instance for "duplicated" Atoms, or will they have to be cloned?

It looks like they share the same instance.

biojava-structure/src/test/java/org/biojava/nbio/structure/TestAltLocs.java
josemduarte: I used the low level methods in `MMCIFFileTools` (e.g. `convertChainToAtomSites()` so that the conversion of biojava objects to mmcif beans could be tested directly. 

Indeed it would also be nice to test the writing but that's more complicated because there can be differences in spacing and sorting of records if you do a full roundtrip from mmcif string to biojava objects and back to mmcif string.

I'll add a commit to change the name of the methods to `testMmcifConversion...`

PR #781, opened 2018-07-02
Travis config to cache pdb dirs across builds

josemduarte:
Travis still fails at downloading external resources quite often (especially ftp resources). This change would cache some of the resources across builds. 

I haven't tested this yet. The build of this PR will do the test.

It is still possible to clear the cache when needed, see https://docs.travis-ci.com/user/caching/#Clearing-Caches

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,15 @@ sudo: required
 cache:
   directories:
    - "$HOME/.m2"
+   - "$HOME/pdb_cache"
 before_install:
  - if [ ! -z "$GPG_SECRET_KEYS" ]; then echo $GPG_SECRET_KEYS | base64 --decode | $GPG_EXECUTABLE --import; fi
  - if [ ! -z "$GPG_OWNERTRUST"  ]; then echo $GPG_OWNERTRUST  | base64 --decode | $GPG_EXECUTABLE --import-ownertrust; fi
 after_success:
  - '[[ $TRAVIS_BRANCH == "master" ]] && { mvn deploy --settings travis-settings.xml -DskipTests=true -B ; };'
 env:
+  - PDB_DIR=$HOME/pdb_cache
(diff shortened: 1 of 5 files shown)


sbliven:
This is not an ideal solution. We often first discover changes to databases when Travis fails. However, the frequent network failures are an even bigger problem. Let's try this and see if it cuts down on the CI failures.

josemduarte:
Indeed I'm not convinced that this is the best solution. But as we are having so many failures I was eager to add some mitigation. 

I've found out that one of the frequently failing tests is `TestSiftsParsing` which uses ftp external resources. Very frequently it just stalls and grinds to a halt, travis then kills the process after 10 minutes of inactivity. Is anyone aware of specific problems with travis and ftp resources?

sbliven:
It looks like FTP isn't supported:

https://docs.travis-ci.com/user/common-build-problems/#FTP-SMTP-other-protocol-does-not-work

It's weird that this used to work though.

sbliven:
I added a commit to your branch switching from FTP to HTTP for Sifts.

The only place BioJava still references ftp is in BlastClustReader (ftp://resources.rcsb.org/sequence/clusters/). Maybe @josemduarte or @MrColeC could enable http or https for that server? On the other hand, it doesn't appear to be tested (👎) so hopefully travis will be OK.

MrColeC:
@sbliven  It is not advertised, but has existed for ~2 years. You can get the sequence clusters here via HTTPS or HTTP

https://cdn.rcsb.org/resources/sequence/clusters/bc-95.out

You cannot get a directory listing, but if you know the file names you can pull them.

josemduarte:
Thank you @sbliven and @MrColeC ! Actually did not realise that EBI were providing the data over http too. 

Since most of the problems came from those ftp resources, I'd propose for now to close this Pull Request and create a new one with only the ftp->http change. If problems with building become bad again we can revisit adding the pdb dirs caching.

Issue #712, opened 2017-10-27
Structure.clone() doesn't copy AtomImpl.bonds lists.


larsonmattr:
This is a problem in the 4.2.x branch and the master branch.

Structure.clone() is described as "Return an identical copy of this Structure object".  

1) If the original Structure was parsed with the FileParsingParameters.setCreateAtomBonds(true), a Structure will be created there each Atom in the Structure has a list of Bonds to describing connections to other Atom instances.

2) If you then make a deep copy by Structure.clone(), the new Structure will be missing any Bond information inside the cloned Atom instances.

I would expect that connectivity would also be preserved in a deep copy.

[TestClone.java.gz](https://github.com/biojava/biojava/files/1423341/TestClone.java.gz)

josemduarte:
Thanks for the report and the test, I've committed it into existing test TestCloning in branch https://github.com/biojava/biojava/tree/issue712. Pull requests are welcomed :)

josemduarte:
Solved in #786. Closing

PR #786, opened 2018-07-29
Solved #712: Bonds are now also copied by clone()

DK-MV:
Bonds are now copied when atoms are copied in a clone(), as described in issue 712
I have also made a few tests platform independent, as they failed on my windows 10 machine.


lafita:
@josemduarte I think that what you are suggesting would not work, since to properly clone a `Bond` it is required to clone the two `Atom` objects involved and then setting their references to one another.
Cloning a single `Atom` would just copy the reference to the original `Atom` partner so we would end up with the original and the cloned `Atoms` bonding to the same object.

This just opened a question for me: what should we do with the bonds when cloning a single `Atom` object? It is not an option to clone the whole `Structure` hierarchy in order to duplicate a single `Atom`...

DK-MV:
@josemduarte, @lafita is right: The clonebonds() cannot be in the Atom.clone(), as it needs to know about other cloned atoms and in Atom it would only know the original atom itself and its clone, and not the cloned object of the atoms it is bonded to. 

@lafita, if one clone a single Atom object, I don't think we should clone any of the bonds, as it would exist on its own and not be part of any larger entity it could be bonded to.

Issue #801, opened 2018-09-03
Biological assembly expansion: chain ids should contain both operator ids in binary expression case


josemduarte:
In cases like 1m4x the assembly operators are given as cartesian products where 2 (or more) operators have to be applied in turn. E.g. is 1m4x:
```
loop_
_pdbx_struct_assembly_gen.assembly_id 
_pdbx_struct_assembly_gen.oper_expression 
_pdbx_struct_assembly_gen.asym_id_list 
1 '(1-60)(61-88)'           A,B,C 
2 '(61-88)'                 A,B,C 
3 '(1-5)(61-88)'            A,B,C 
4 '(1,2,6,10,23,24)(61-88)' A,B,C 
5 '(1-5)(63-68)'            A,B,C 
6 '(1,10,23)(61,62,69-88)'  A,B,C 
7 '(P)(61-88)'              A,B,C 
```
BioJava deals correctly with this but assigning sequential operator ids instead of the original ones in `_pdbx_struct_oper_list.id`. It'd be better to use the original operators so that the resulting identifiers can be recognised by other tools (e.g. when outputting assemblies to mmcif files).

PR #802, opened 2018-09-03
Assembly chain ids for cases with composed operators in assembly expansion

josemduarte:
A solution to problem described in #801. 
* It assigns chain ids containing all operator ids, e.g. for chain id `A` where operators 5 and 61 have been applied, it assigns new chain id `A_5x61`
* A new test is added to check correct behaviour in cartesian product `_pdbx_struct_assembly_gen.oper_expression`  cases
* Some cleanup removing jama objects from `PdbxStructOperList` mmcif model bean
* For consistency, the `Subunit` objects used in symmetry detection now use asym_ids instead of author_ids.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitExtractor.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitExtractor.java
@@ -75,7 +75,7 @@ public class SubunitExtractor {
 				logger.debug("Chain " + c.getId() + "; CA Atoms: " + ca.length + "; SEQRES: " + c.getSeqResSequence());
 				if (ca.length==0)
 					continue;
-				subunits.add(new Subunit(ca, c.getName(), null, structure));
+				subunits.add(new Subunit(ca, c.getId(), null, structure));
 			}
 		}
 
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitExtractor.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitExtractor.java
@@ -48,8 +48,8 @@ public class SubunitExtractor {
 
 	/**
 	 * Extract the information of each protein Chain in a Structure and converts
-	 * them into a List of Subunit. The name of the Subunits is set to the
-	 * {@link Chain#getName()}.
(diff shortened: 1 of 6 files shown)


pwrose:
I've filed an issue about the chain length limitation in mmtf:
https://github.com/rcsb/mmtf/issues/37

josemduarte:
Thanks for filing the issue @pwrose . Please note that for mmtf encoding of the original mmcif file (without bioassembly expansion) there wouldn't be any issues. This can present an issue for mmtf if a bioassembly output by biojava would need to be encoded in mmtf.

pwrose:
According to: https://www.wwpdb.org/deposition/preparing-pdbx-mmcif-files
PDBx/mmcif is also restricted to 4 characters: "PDB format allows only
single-character chain IDs, while PDBx/mmCIF can accommodate chain IDs of
up to four characters.".

biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitExtractor.java
```diff
@@ -75,7 +75,7 @@ private SubunitExtractor() {
 				logger.debug("Chain " + c.getId() + "; CA Atoms: " + ca.length + "; SEQRES: " + c.getSeqResSequence());
 				if (ca.length==0)
 					continue;
-				subunits.add(new Subunit(ca, c.getName(), null, structure));
+				subunits.add(new Subunit(ca, c.getId(), null, structure));
```
valasatava: Could you, please, also correct docs to reflect this change?

josemduarte: Good catch. Just pushed a commit

josemduarte:
Strictly speaking the (mm)CIF format does not constrain the number of chars to 4 (see http://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_struct_asym.id.html). In practice it is limited to 4 chars for PDB-produced files. For other producers of files there should be no issue having more than 4 chars.

With MMTF there is a limitation to 4 chars (@pwrose 's point), there's a separate issue for that here: https://github.com/rcsb/mmtf/issues/37

In any case this PR is only concerned about producing assembly data structures in memory and not on their representation in files.

Issue #804, opened 2018-09-07
Make biojava work in Java 9, 10, 11


josemduarte:
@sbliven made some progress in the [java9 branch](/biojava/biojava/tree/java9).

I've tried out the branch and it looks like the main remaining problem is with the jaxb dependency used in a few places (jaxb is apparently not included in 9+).

sbliven:
Oracle announced they will charge for java 8 updates, so this is timely!

https://www.itassetmanagement.net/2018/05/01/oracle-to-charge-for-java-from-jan-2019/

paolopavan:
Does anyone had tried if there are any major compatibility issues? Any
issue of any kind?
I am especially worried about frequent Api changes that will compel to
update my old software too frequently than what I can do. As recalled by
the article, new releases will happen every 6 months. While this could be
interesting when bringing new features in, it can be not sustainable when
those would be API breaking.

Any thoughts?

Il giorno ven 5 ott 2018 alle 08:55 Spencer Bliven <notifications@github.com>
ha scritto:

darnells:
Starting January 2019, Oracle Java (version 11 and beyond) will be a paid-for commercial product and OpenJDK will be the free open-source distribution. OpenJDK has been providing binaries for Windows, macOS, and Linux since JDK 9. I am sure @josemduarte, @sbliven, and others will successfully transition BioJava to Java 11 (the current LTS release) in due time and the majority of us will hop onto OpenJDK 11 sometime in 2019.

@paolopavan, you will want to stick to LTS releases to reduce the frequency of API changes, which are presumably one every 18 months (time will tell if Oracle/OpenJDK stick to that schedule). Otherwise, you could always stick with JDK 8 for as long as your dependencies support it.

sbliven:
@paolopavan The biggest thing was the removal of some packages (e.g. XML). These are now pulled in from maven, so there will be no change for the user. If you want products to be stable (LTS), it looks like the best solution is to migrate them to OpenJDK 11 over the next year or so.

PR #805, opened 2018-09-07
Make biojava work with java 9, 10, 11

josemduarte:
Solves #804. Thanks @sbliven for most of the work.

This does:
- Keeps Java 8 compliance (compiled byte code is 8)
- Makes biojava work with java 9, 10 or 11 JREs

At some later point we can decide to move to next version of compliance (9, 10 or 11). That only requires changing the properties `jdk.version` and `maven.enforcer.jdk-version` in main pom.xml.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
 jdk:
- - oraclejdk9
+ - oraclejdk8
 sudo: required
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
 jdk:
- - oraclejdk8
+ - oraclejdk9
 sudo: required
 cache:
   directories:
--- a/pom.xml
+++ b/pom.xml
@@ -512,6 +512,21 @@
 			    <artifactId>jaxb-api</artifactId>
 			    <version>2.3.0</version>
 			</dependency>
+			<dependency>
+				<groupId>com.sun.xml.bind</groupId>
+				<artifactId>jaxb-core</artifactId>
+				<version>2.3.0</version>
+			</dependency>
+			<dependency>
+				<groupId>com.sun.xml.bind</groupId>
+				<artifactId>jaxb-impl</artifactId>
+				<version>2.3.0</version>
+			</dependency>
+			<dependency>
+				<groupId>javax.activation</groupId>
+				<artifactId>activation</artifactId>
+				<version>1.1.1</version>
+			</dependency>
 		</dependencies>
 	</dependencyManagement>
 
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: java
 jdk:
- - oraclejdk8
+ - oraclejdk9
```
lafita: Does this mean Travis will run the tests using java 9?

josemduarte: Yes, that's right. Everything would be compiled at java 8 compliance but the run time in travis would be 9.
Running with 9 JRE is more realistic, since it would be what you get by default when you install a jre in your system these days (or perhaps it's already 10).

darnells:
[Java 9 and 10 are "non-LTS" releases](http://www.oracle.com/technetwork/java/javase/eol-135779.html). Java 9 was EOL'd in March when Java 10 was released. The same will happen to Java 10 when Java 11 is released later this month. I am glad to see newer JVMs will be supported, but be careful using EOL'd or bleeding edge releases for production builds.

josemduarte:
@darnells thanks for the extra information. Are you then suggesting reverting the travis build JRE to jdk8? I'd be ok to do that.

darnells:
@josemduarte I choose to be pretty conservative when using EOL'd or short term support software, so my preference would be to rollback for now. However, I will defer implementation decisions to the core developers and simply ask that BioJava officially supports Java 8 at least until January 2019 (Commercial end of public updates) or ideally until April or July 2019 to provide BioJava users time to transition to Java 11.

Perhaps this would be a good time to consider the future use of multi-release jar files ([JEP 247: Compile for Older Platform Versions](http://openjdk.java.net/jeps/247), [JEP 238: Multi-Release JAR Files](http://openjdk.java.net/jeps/238)) to insulate BioJava users from change while allowing developers to use new language features?

josemduarte:
@darnells to be clear: the only intention of this PR is to make the code __runnable__ under Java 8, 9, 10, 11 JREs. At the moment it is runnable only in Java 8. Other than that it does not change anything: we are still fully Java 8 compliant.

I've reverted the travis jdk9 to jdk8 so that there's no ambiguities.

I agree with you that changing our java version compliance is an important discussion and it has to be carefully thought and planned.

darnells:
@josemduarte Sounds good to me! Sorry for being a fly in the ointment.

pom.xml
```diff
@@ -171,27 +172,27 @@
...
+					<version>3.1.1</version>
 				</plugin>
 				<plugin>
 					<artifactId>maven-jar-plugin</artifactId>
 					<!-- warning: before upgrading to a newer version of this plugin
 					      there are some config changes required.
 					      see https://github.com/biojava/biojava/issues/531
 					      -->
-					<version>3.0.2</version>
+					<version>3.1.0</version>
```
heuermh: Could you remove the comment above this line?

josemduarte: Good catch, I'll push a commit

PR #820, opened 2019-01-03
Improved ASA calculation performance

josemduarte:
Now neighbors are calculated with spatial hashing, improving performance dramatically for cases with many thousands of atoms. Also improved docs, logs and added new tests.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
@@ -54,8 +54,11 @@ public class AsaCalculator {
 
 	private static final Logger logger = LoggerFactory.getLogger(AsaCalculator.class);
 
-	// Bosco uses as default 960, Shrake and Rupley seem to use in their paper 92 (not sure if this is actually the same parameter)
-	public static final int DEFAULT_N_SPHERE_POINTS = 960;
+	/**
+	 * The default value for number of sphere points to sample.
+	 * See this paper for a nice study on the effect of this parameter: https://f1000research.com/articles/5-189/v1
+	 */
+	public static final int DEFAULT_N_SPHERE_POINTS = 1000;
 	public static final double DEFAULT_PROBE_SIZE = 1.4;
 	public static final int DEFAULT_NTHREADS = 1;
 
@@ -112,10 +115,11 @@ public class AsaCalculator {
 	 * Constructs a new AsaCalculator. Subsequently call {@link #calculateAsas()}
 	 * or {@link #getGroupAsas()} to calculate the ASAs
 	 * Only non-Hydrogen atoms are considered in the calculation.
-	 * @param structure
-	 * @param probe
-	 * @param nSpherePoints
-	 * @param nThreads
+	 * @param structure the structure, all non-H atoms will be used
+	 * @param probe the probe size
+	 * @param nSpherePoints the number of points to be used in generating the spherical
+	 *                         dot-density, the more points the more accurate (and slower) calculation
+	 * @param nThreads the number of parallel threads to use for the calculation
 	 * @param hetAtoms if true HET residues are considered, if false they aren't, equivalent to
 	 * NACCESS' -h option
 	 * @see StructureTools#getAllNonHAtomArray
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
@@ -251,17 +251,21 @@ public class AsaCalculator {
 		double[] asas = new double[atomCoords.length];
 
 		if (useSpatialHashingForNeighbors) {
+			logger.debug("Will use spatial hashing to find neighbors");
 			neighborIndices = findNeighborIndicesSpatialHashing();
 		} else {
+			logger.debug("Will not use spatial hashing to find neighbors");
 			neighborIndices = findNeighborIndices();
 		}
 
 		if (nThreads<=1) { // (i.e. it will also be 1 thread if 0 or negative number specified)
+			logger.debug("Will use 1 thread for ASA calculation");
 			for (int i=0;i<atomCoords.length;i++) {
 				asas[i] = calcSingleAsa(i);
 			}
 
 		} else {
+			logger.debug("Will use {} threads for ASA calculation", nThreads);
 			// NOTE the multithreaded calculation does not scale up well in some systems,
 			// why? I guess some memory/garbage collect problem? I tried increasing Xmx in pc8201 but didn't help
 
@@ -315,7 +319,13 @@ public class AsaCalculator {
 		return asas;
 	}
 
-	public void setUseSpatialHashingForNeighbors(boolean useSpatialHashingForNeighbors) {
+	/**
+	 * Set the useSpatialHashingForNeighbors flag to use spatial hashing to calculate neighbors (true) or all-to-all
+	 * distance calculation (false). Default is {@value DEFAULT_USE_SPATIAL_HASHING}.
+	 * Use for testing performance only.
+	 * @param useSpatialHashingForNeighbors the flag
+	 */
+	void setUseSpatialHashingForNeighbors(boolean useSpatialHashingForNeighbors) {
 		this.useSpatialHashingForNeighbors = useSpatialHashingForNeighbors;
 	}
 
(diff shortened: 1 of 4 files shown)


biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
```diff
@@ -21,12 +21,13 @@
...
 import org.biojava.nbio.structure.*;
+import org.biojava.nbio.structure.contact.Contact;
+import org.biojava.nbio.structure.contact.Grid;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.vecmath.Point3d;
-import java.util.ArrayList;
-import java.util.TreeMap;
+import java.util.*;
```
sbliven: Minor point, but maybe we should try to avoid wildcard imports?

biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
```diff
@@ -36,28 +37,34 @@
...
-	public static final int DEFAULT_N_SPHERE_POINTS = 960;
+	/**
+	 * The default value for number of sphere points to sample.
+	 * See this paper for a nice study on the effect of this parameter: https://f1000research.com/articles/5-189/v1
+	 */
+	public static final int DEFAULT_N_SPHERE_POINTS = 1000;
 	public static final double DEFAULT_PROBE_SIZE = 1.4;
 	public static final int DEFAULT_NTHREADS = 1;
 
+	public static final boolean DEFAULT_USE_SPATIAL_HASHING = true;
```
sbliven: Should this be private?

biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
```diff
@@ -293,10 +325,22 @@ public AsaCalculator(Point3d[] atomCoords, double probe, int nSpherePoints, int
...
+		end = System.currentTimeMillis();
+		logger.debug("Took {} s to calculate all {} atoms ASAs (excluding neighbors calculation)", (end-start)/1000.0, atomCoords.length);
 
 		return asas;
 	}
 
+	/**
+	 * Set the useSpatialHashingForNeighbors flag to use spatial hashing to calculate neighbors (true) or all-to-all
+	 * distance calculation (false). Default is {@value DEFAULT_USE_SPATIAL_HASHING}.
+	 * Use for testing performance only.
```
sbliven: For small molecules is the old algorithm faster? If so maybe document that here.

biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
josemduarte: That's an IntelliJ feature. I do like individual imports better too... I'll have to check if I can configure IntelliJ to do it.

biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
josemduarte: Indeed, thanks for catching. I'll make private

biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
josemduarte: Performance is essentially the same for small molecules. The gain is very noticeable in very large molecules.

biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
```diff
@@ -293,10 +325,22 @@ public AsaCalculator(Point3d[] atomCoords, double probe, int nSpherePoints, int
 			while (!threadPool.isTerminated());
 
 		}
+		end = System.currentTimeMillis();
+		logger.debug("Took {} s to calculate all {} atoms ASAs (excluding neighbors calculation)", (end-start)/1000.0, atomCoords.length);
```
kanishka-azimi: I'd probably keep invoke this function form a test and time and log there, eventually move it to a benchmark test suite.

josemduarte: My use case was testing other software that uses ASA calculation deep into some function call. I thought this would be a good way of debugging it. Do you see potential performance problems with this?

kanishka-azimi: No, I don't see any real issue. I personally tend to move code from main to tests whenever possible, simply to reduce clutter in the real code.

currentTimeMillis probably cost nothing and the other computation in the logging steps seems minimal. I'm transitioning back to java, so I'm not an expert in java just yet.

Feel free to ignore my comments here. I'm mostly writing for myself to get my head into some of biojava's code.

biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
```diff
@@ -317,36 +361,118 @@ public AsaCalculator(Point3d[] atomCoords, double probe, int nSpherePoints, int
...
+
+				double dist = atomCoords[i].distance(atomCoords[k]);
+
+				if (dist < radius + radii[i]) {
+					thisNbIndices.add(i);
+				}
+			}
+
+			int[] indicesArray = new int[thisNbIndices.size()];
+			for (int i=0;i<thisNbIndices.size();i++) indicesArray[i] = thisNbIndices.get(i);
```
kanishka-azimi: Very minor note. Since random access isn't being used in the array list, could use Queue interface instead of List.

kanishka-azimi: Some cute shortcuts - https://stackoverflow.com/questions/960431/how-to-convert-listinteger-to-int-in-java . I don't know if biojava has any general preference regarding guava or apache commons.

josemduarte: Thanks! I like the answer that uses the stream API: https://stackoverflow.com/a/23945015/3914327

biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
```diff
@@ -317,36 +361,118 @@ public AsaCalculator(Point3d[] atomCoords, double probe, int nSpherePoints, int
 	}
 
 	/**
-	 * Returns list of indices of atoms within probe distance to atom k.
-	 * @param k index of atom for which we want neighbor indices
+	 * Returns the 2-dimensional array with neighbor indices for every atom.
+	 * @return 2-dimensional array of size: n_atoms x n_neighbors_per_atom
```
kanishka-azimi: adjacency list vs adjacency matrix...possible prefer one or the other universally based on expected density of graph

biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
```diff
@@ -317,36 +361,118 @@ public AsaCalculator(Point3d[] atomCoords, double probe, int nSpherePoints, int
...
+	}
+
+	private static double maxValue(double[] array) {
+		double max = array[0];
+		for (int i = 0; i < array.length; i++) {
+			if (array[i] > max) {
+				max = array[i];
+			}
+		}
+		return max;
```
kanishka-azimi: minor: some stream shortcut could replace this function.

biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
```diff
@@ -317,36 +361,118 @@ public AsaCalculator(Point3d[] atomCoords, double probe, int nSpherePoints, int
...
+			int j = contact.getJ();
+
+			List<Integer> iIndices;
+			List<Integer> jIndices;
+			if (!indices.containsKey(i)) {
+				iIndices = new ArrayList<>(initialCapacity);
+				indices.put(i, iIndices);
+			} else {
+				iIndices = indices.get(i);
+			}
```
kanishka-azimi: minor: putIfAbsent

biojava-structure/src/test/java/org/biojava/nbio/structure/asa/TestAsaCalc.java
```diff
@@ -70,12 +70,114 @@ public void testAsa3PIU() throws StructureException, IOException {
...
+//			if (listOfMatchingIndices.contains(i)) {
+//				System.out.printf("Matching     - indices %d-%d: %5.2f\n", indexToTest, i, dist);
+//			} else {
+//				System.out.printf("Not matching - indices %d-%d: %5.2f\n", indexToTest, i, dist);
+//			}
+//		}
+
+			assertEquals(nbs.length, nbsSh.length);
+
+			assertEquals(nbs.length, listOfMatchingIndices.size());
```
kanishka-azimi: I'd probably have two tests.
1. test for the functionality against the simplest implementation. 
2. test ensuring the two implementations produce equivalent results.
I am only reading the diffs, so I may be overlooking what is already there.

biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
sbliven: @kanishka-azimi Thanks for contributing! The more people reviewing PRs the better!

PR #823, opened 2019-01-16
Avoid ASA calculation for redundant NCS interfaces

josemduarte:
Making ASA calculation of interfaces more streamlined, thanks to #732


biojava-structure/src/main/java/org/biojava/nbio/structure/contact/StructureInterface.java
```diff
@@ -21,10 +21,7 @@
 package org.biojava.nbio.structure.contact;
 
 import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.TreeMap;
+import java.util.*;
```
lafita: I think we decided to avoid * imports.

josemduarte: Sorry about that again. I'll revert and set up IntelliJ properly once and for all. FYI this is the way: https://stackoverflow.com/a/3348855/3914327

josemduarte:
Just pushed some commits, the solution wasn't quite working because interfaces can be in different orders within an NCS cluster. We'll need to squash the commits before merging.

PR #825, opened 2019-01-21
Bugfix: expanded symmetry partner chains weren't added to entities

josemduarte:
This prevented things like writing out to mmcif file a bioassembly with expanded chains.

### DIFF
--- a/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/io/TestMMCIFWriting.java
+++ b/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/io/TestMMCIFWriting.java
@@ -1,0 +1,28 @@
+package org.biojava.nbio.structure.test.io;
+
+import org.biojava.nbio.structure.Structure;
+import org.biojava.nbio.structure.StructureException;
+import org.biojava.nbio.structure.StructureIO;
+import org.biojava.nbio.structure.xtal.CrystalBuilder;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.HashMap;
+
+import static org.junit.Assert.assertNotNull;
+
+public class TestMMCIFWriting {
+
+    /**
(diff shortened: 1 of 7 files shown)


biocryst:
Thanks for looking into this!
The issue with writing mmCIF is fixed. There is a related issue (which may be processed separately from this one) -- the state of a biological assembly structure EntityInfos' are not entirely consistent.

For example, here
https://github.com/biojava/biojava/blob/86bbe0a75f61a2b3cd3708ea0bfec4adcbddd57e/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/io/TestMMCIFWriting.java#L21-L22

1. entityInfos on the structure level contain references to the old chains ("A") and no references to the new chains.
2. entity on the chain level contains references to both old and new chains ("A", "A_1", "A_2", etc.)
3. this leads to warnings when a structure is cloned, and entityInfo of a copy is empty.

A solution could be resetting structure-level entityInfos here as well as models:
https://github.com/biojava/biojava/blob/86bbe0a75f61a2b3cd3708ea0bfec4adcbddd57e/biojava-structure/src/main/java/org/biojava/nbio/structure/quaternary/BiologicalAssemblyBuilder.java#L101-L103

and resetting chain-level entity of a chain copy maybe somewhere here?
https://github.com/biojava/biojava/blob/86bbe0a75f61a2b3cd3708ea0bfec4adcbddd57e/biojava-structure/src/main/java/org/biojava/nbio/structure/quaternary/BiologicalAssemblyBuilder.java#L125-L127

josemduarte:
Thanks for catching that @biocryst ! I'll try to fix.

josemduarte:
I've pushed some commits that should address the problems

josemduarte:
Thanks @biocryst , I'll merge.

Next week I will produce a 5.2.1 release. If there are any bugfixes pending please submit them as soon as possible.

PR #834, opened 2019-04-12
Uniprot short name fix

emckee2006:
Update UniprotProxySequenceReader.java

2 fixes:

1. Full Uniprot downloads contain spaces in some sequences, which cause a crash in setContents because space isn't a valid compound. I added a replaceAll to strip out the whitespace.
2. Apparently a shortName element can be repeated. In both cases where we can encounter a shortName element, get the list (perhaps of length 0 or 1) and iterate over each shortName instead of just grabbing the first one.


emckee2006:
i'm not sure how these get visibility...

biojava-core/src/main/java/org/biojava/nbio/core/sequence/loader/UniprotProxySequenceReader.java
```diff
@@ -106,6 +106,8 @@ public UniprotProxySequenceReader(String accession, CompoundSet<C> compoundSet)
 
 	/**
 	 * The xml is passed in as a DOM object so we know everything about the protein.
+	 * Some uniprot records contain white space in the sequence. We must strip it out so setContents doesn't fail.
+	 * TODO add simmilar logic to the other constructors
```
heuermh: If so, should the `replaceAll` move inside `setContents`?

emckee2006:
That actually makes more sense... I'll make that change...

emckee2006:
Done :)

heuermh:
Thanks!  I'll wait for Travis to chime in, then approve.

emckee2006:
Are we ok with this?

heuermh:
Thank you, @emckee2006

PR #839, opened 2019-05-02
Add support for BinaryCIF

JonStargaryen:
## What is BinaryCIF

- [BinaryCIF](https://github.com/dsehnal/BinaryCIF) is a compact representation of CIF files
- similar encoding strategies are used as in MMTF - however the format is fully compatible with conventional, text-based CIF files
- therefore, we get all the advantages and established handling of CIF files as well as the performance of MMTF
- [ciftools-java](https://github.com/rcsb/ciftools-java) provides the BinaryCIF codec, all model classes for the content of CIF files, and capabilities for the parsing of text-based CIF files
- content from binary and text files is stored in the same data structure (a CifFile) - thus, it does not matter where the data came from: binary and text-based CIF data is handled the same way

## What changed

- I tried to introduce no change to existing classes and behavior, the implementation rather coexists with the established and tested parsing of CIF files
- handling of BinaryCIF and CIF files is delegated to ciftools, this library provides a representation which can be used to populate Structure instances with data
- parsing behavior was taken from SimpleMMcifConsumer - no information should be missing
- the majority of tests we duplicated for now to work on the new implementation (both on binary and text-based CIF files)

## Performance

- Some basic performance tests.

### The whole PDB archive (in parallel)

- parsing 151079 BinaryCIF files takes 918 s
- parsing 151079 (text-based) CIF files takes 2502 s (a new experimental implementation for text-based CIF parsing which comes basically for free by wrapping ciftools)

### 5000 PDB structures (in parallel)

- parsing 5000 BinaryCIF files takes 35 s
- parsing 5000 (text-based) CIF files with the new CIF parser takes 132 s
- parsing 5000 (text-based) CIF files with the current BioJava CIF parser takes 310 s
- parsing 5000 MMTF files (fetched from the web) with the BioJava MMTF parser the takes 17 s

- in summary, parsing for conventional CIF files is substantially faster
- performance should match MMTF when only atom coordinates are considered, BinaryCIF however provides all content from a file

## Side notes

- SimpleMMcifConsumer uses SimpleDateFormat for date parsing which caused a number of parsing exceptions when executed in parallel - my implementation uses Java 8's LocalDate for parsing, while still reporting Date objects compatible with the data structure

## Down the road

- integrate into StructureIO/AtomCache
- omit/merge duplicated code if stable (SimpleMMcifConsumer, mmCIF model beans)
- the new code is a good chance to revisit missing categories in the SimpleMMcifConsumer (currently not parsed content) and refine the writing in FileConvert


biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/io/cif/CifFileSupplierImplTest.java
```diff
@@ -0,0 +1,176 @@
+package org.biojava.nbio.structure.test.io.cif;
+
+import org.biojava.nbio.structure.*;
```
lafita: We are trying to avoid star (*) imports. Could you try to replace them for specific Class imports? It happens in a number of cases.

biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/io/cif/CifFileConsumerImplTest.java
```diff
@@ -0,0 +1,357 @@
...
+     * MMTF: 17 s for 5000 fetched structures, 3211 µs per structure, failed for 0 entries
+     *
+     * CIF parsing using David's tokenizer approach takes roughly half the time. Binary parsing using ciftools takes
+     * roughly double the time compared to MMTF and should be even more at a disadvantage when MMTF data is stored
+     * locally. However, MMTF provides almost no metadata.
+     *
+     * @throws IOException propagated
+     */
+    @Test
+    @Ignore("ignore long-running test, do run to track performance")
```
lafita: This test could be moved to the integrationtest module, where we place longer tests that run for longer and can be skipped during regular development.

biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/io/cif/CifFileSupplierImplTest.java
JonStargaryen: Package imports are removed in 0b59a10

biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/io/cif/CifFileConsumerImplTest.java
JonStargaryen: That tests should already be in the integrationtest module - however, there is also a test class of the same name in the structure module. I've renamed the tests in the integration module (1755a4c), so that there are no duplicate name for test classes.

lafita: Oh, I see! Sorry for the confusion.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/CifFileSupplierImpl.java
```diff
@@ -0,0 +1,316 @@
...
+        private final FloatColumnBuilder<CategoryBuilder> bIsoOrEquiv;
+        private final IntColumnBuilder<CategoryBuilder> authSeqId;
+        private final StrColumnBuilder<CategoryBuilder> authCompId;
+        private final StrColumnBuilder<CategoryBuilder> authAsymId;
+        private final StrColumnBuilder<CategoryBuilder> authAtomId;
+        private final IntColumnBuilder<CategoryBuilder> pdbxPDBModelNum;
+
+        AtomSiteCollector() {
+            // TODO this doesn't really make the case for the builder ;)
+            this.groupPDB = new StrColumnBuilder<>("atom_site", "group_PDB", null);
```
josemduarte: I thought these hard-coded strings shouldn't be needed. They should come from metadata available via the CIFTools library. Or am I missing something?

heuermh:
Curious, is there a reason why the code delegated to here is in a separate repository instead of a separate module in biojava?

biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/CifFileSupplierImpl.java
JonStargaryen: Thanks for pointing this out. 196e0c5 replaces the hard-coded category and column names with the corresponding builder functions.

JonStargaryen:
> Curious, is there a reason why the code delegated to here is in a separate repository instead of a separate module in biojava?

The ciftools code is not BioJava-specific. Rather it provides generic functionality for the handling of CIF files which can also be used by other projects. Therefore, I opted to create a separate project.

heuermh:
> The ciftools code is not BioJava-specific.

Well, the scope of BioJava is what is in BioJava.  There wasn't any support for 3D structures in BioJava until there was.  :)

I would much rather have `ciftools-java` exist within BioJava than add a new dependency on an external single-developer dependency.  That will help make deprecation/refactoring/removal easier.

JonStargaryen:
> I would much rather have `ciftools-java` exist within BioJava than add a new dependency on an external single-developer project. That will help make deprecation/refactoring/removal of the current CIF implementation in BioJava easier.

That sure is a valid concern. The basic (binary) CIF handling is realized by dedicated, language-specific projects to be able to publish/promote the format as well as codec implementations. A project is more visible than a module somewhat hidden in BioJava. The same was done for MMTF.

Also, don't worry about maintenance as ciftools-java is a RCSB project. Jose has admin rights to the repo and we're literally one desk apart for the case that refactoring becomes an issue.

josemduarte:
@heuermh just want to reiterate that outsourcing to an external library is the same as was done for the mmtf format. Additionally, in this case, the new CIFtools library is more general than just parsing BinaryCIF. It provides all the tooling needed for dealing with the CIF format and CIF schemas. So it does a lot more that what we would want to have in BioJava.

josemduarte:
I'll merge this by the end of this week if there are no other concerns.

Issue #760, opened 2018-03-28
CONECT/LINK line-length bug fix #2 for PDB Parser


Sheraf-DNG:
Hi All, 

I'm working with custom PDB Files and I'm getting StringIndexOutOufBondsException from the PDBParser.
I've seen you rencently added this line in the PDBParser  (conect_helper (String line,int start,int end) method) : 

if  line.length() < end) return null;

It'b be nice if to check the line.length() do not exeed the start arg, should be fixed by changing it to this (tested):

if (line.length() < start || line.length() < end) return null;

Thanks for the good work!

josemduarte:
Good catch, could you provide a pull request so that we can review it and merge if everything looks fine? It's difficult to follow exactly what your changes are without the code context.

See here for instructions on how to do a pull request: https://help.github.com/articles/creating-a-pull-request/

r614:
Hey! I fixed the bug on my local branch but I am unable to push the branch to upstream to make a PR. I require permissions!

lafita:
Hi @r614, we do not allow directly pushing changes to the main repository. You should instead fork the biojava repository to your account, push the changes there, and then open a new PR. You can find more info in the [contributing instructions](https://github.com/biojava/biojava/blob/master/CONTRIBUTING.md).

r614:
Yikes, that's my bad. Just added a PR.

PR #841, opened 2019-06-10
Fixed Line Length bug - #760

r614:
Added the recommended solution for the line length bug from Issue #760

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/io/PDBFileParser.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/io/PDBFileParser.java
@@ -2002,7 +2002,7 @@ public class PDBFileParser  {
 
 	/** safes repeating a few lines ... */
 	private Integer conect_helper (String line,int start,int end) {
-		if (line.length() < end) return null;
+		if (line.length() < start || line.length() < end) return null;
 
 		String sbond = line.substring(start,end).trim();
 		int bond  = -1 ;


biojava-structure/src/main/java/org/biojava/nbio/structure/io/PDBFileParser.java
```diff
@@ -2002,7 +2002,7 @@ private void switchCAOnly(){
 
 	/** safes repeating a few lines ... */
 	private Integer conect_helper (String line,int start,int end) {
-		if (line.length() < end) return null;
+		if (line.length() < start || line.length() < end) return null;
```
lafita: I am a bit confused because for this function I assume that by definition `start < end`, and therefore checking `line.length() < end` should be sufficient here. @r614 do you know of a case where this is not true? Maybe the solution would be to check that indeed `start < end` always and track when this is problematic during parsing. 

@Sheraf-DNG could you provide your test case so that we include it as a unit test and make sure it is fixed?

Sheraf-DNG: This was over ago, I can't remember it sorry... I'm myself baffled by the solution I proposed :/ but yep there was a StringOutOfBound trrown from this function, and adding this seemed to solve it in my tests...

r614: @lafita I agree with you, but I think it would also depend on how BioJava wants to handle specs. 
I think the possible ways you get this error is if:

1. (start > end ) < line length 
2. end < line length < start 
3. start < line length < end 

If any of those cases are possible, or users can use the method then I think adding a checkpoint won't be a bad idea, but if the function is used internally then I don't think it would really matter, considering it passes all tests so far.

josemduarte:
I'll close this. The conclusion seems to be that this is a private function and does not need extra input validation checks.

PR #851, opened 2019-09-03
Bump guava dependency version to 28.1-jre.

heuermh:


heuermh:
Thank you @josemduarte

PR #854, opened 2019-11-12
Fixing issue in bonds between atoms with different alt locs

josemduarte:
As reported in https://github.com/rcsb/mmtf/issues/44


josemduarte:
Will merge next week if there's no comments

PR #857, opened 2020-01-07
Optimization of subunit clusterer for quaternary sym detection of PDB deposited structures

josemduarte:
This pull request introduces a new switch for the subunit clusterer used for symmetry detection. If the switch `useEntityIdForSeqIdentityDetermination` is enabled, it uses the entity id of the subunits to establish identity of sequences, saving the full Smith-Waterman alignment calculation.

This optimization is important in cases like large viral capsids, where there are many thousands of chains and all-to-all sequence alignments become a bottleneck. E.g. for 6Q1F the runtime goes from ~ 6 hours to 7 minutes.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitCluster.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitCluster.java
@@ -71,11 +71,10 @@ import java.util.stream.Collectors;
  */
 public class SubunitCluster {
 
-	private static final Logger logger = LoggerFactory
-			.getLogger(SubunitCluster.class);
+	private static final Logger logger = LoggerFactory.getLogger(SubunitCluster.class);
 
-	private List<Subunit> subunits = new ArrayList<Subunit>();
-	private List<List<Integer>> subunitEQR = new ArrayList<List<Integer>>();
+	private List<Subunit> subunits = new ArrayList<>();
+	private List<List<Integer>> subunitEQR = new ArrayList<>();
 	private int representative = -1;
 
 	private SubunitClustererMethod method = SubunitClustererMethod.SEQUENCE;
@@ -119,7 +118,7 @@ public class SubunitCluster {
 
 		subunits.add(subunit);
 
-		List<Integer> identity = new ArrayList<Integer>();
+		List<Integer> identity = new ArrayList<>();
 		for (int i = 0; i < subunit.size(); i++)
 			identity.add(i);
 		subunitEQR.add(identity);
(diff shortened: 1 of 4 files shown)


biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitClusterer.java
```diff
@@ -74,14 +74,10 @@ public static Stoichiometry cluster(List<Subunit> subunits, SubunitClustererPara
...
 								clusters.get(c1).mergeIdenticalByEntityId(clusters.get(c2))) {
 							// This we will only do if the switch is for entity id comparison is on.
-							// In some cases in can save enormous amounts of time, e.g. for clustering full
+							// In some cases it can save enormous amounts of time, e.g. for clustering full
 							// chains of deposited PDB entries. For instance for 6NHJ: with pure alignments it
 							// takes ~ 6 hours, with entity id comparisons it takes 2 minutes.
 							clusters.remove(c2);
-						} else if (clusters.get(c1).mergeIdentical(clusters.get(c2))) {
-							// This always makes sense as an optimization: it's far cheaper to compare the sequence
-							// string than doing a full S-W alignment
```
lafita: I think you are correct here but there are some problems like missing residues at the ends or modified amino acids, so using S-W alignment was an easy solution at the time, but clearly not optimal. Do you have an idea why the test is failing?

biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitCluster.java
```diff
@@ -71,11 +74,10 @@
  */
 public class SubunitCluster {
 
-	private static final Logger logger = LoggerFactory
-			.getLogger(SubunitCluster.class);
+	private static final Logger logger = LoggerFactory.getLogger(SubunitCluster.class);
 
-	private List<Subunit> subunits = new ArrayList<Subunit>();
-	private List<List<Integer>> subunitEQR = new ArrayList<List<Integer>>();
+	private List<Subunit> subunits = new ArrayList<>();
```
lafita: Why did you remove the Class of elements in the List when it is defined? Has something changed in Java8+?

biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitClusterer.java
josemduarte: With this optimization enabled, the test that fails is `TestQuatSymmetryDetectorExamples.testLocal()`, but I have no idea why. I then decided to remove this optimization as anyway for my use-case the entity id comparison takes care of it.

biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitCluster.java
josemduarte: Since Java 7 it is redundant to use the types in the variable initialization. They `<>` is called "diamond". See "The diamond" section here: https://docs.oracle.com/javase/tutorial/java/generics/types.html

I think it is good style to remove them, since they are redundant. It helps readability.

josemduarte:
FYI for all: after this is merged I'd like to cut a 5.4.0 release (bump in minor, since this one introduces a tiny new feature).

biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitClusterer.java
lafita: Ok makes sense! Thanks

PR #858, opened 2020-01-08
use static analysis tool

bestborn14:
SonarLint is an IDE extension that helps you detect and fix quality issues as you write code.
Like a spell checker, SonarLint squiggles flaws so that they can be fixed before committing code.


heuermh:
Closing this pull request, as no significant code changes are requested, and this reads like an product advertisement.

PR #859, opened 2020-01-23
Fixes on optimization of subunit clusterer for quaternary sym detection of PDB deposited structures

josemduarte:
A follow up to #857 

This adds tests and fixes a few problems in the implementation (wasn't updating equivalent residues indices for chains of same entity). Also added docs and a new utility method in StructureTools.

In any case as of this PR I can't reproduce anymore the dramatic difference in runtime mentioned in #857 . One of the new (ignored) tests in this PR is to demonstrate the runtime difference, but it's only ~ 5x

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitCluster.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitCluster.java
@@ -192,22 +192,35 @@ public class SubunitCluster {
 	 * @return true if the SubunitClusters are identical, false otherwise
 	 */
 	public boolean isIdenticalByEntityIdTo(SubunitCluster other) {
-		Structure thisStruct = this.subunits.get(this.representative).getStructure();
-		Structure otherStruct = other.subunits.get(other.representative).getStructure();
-		String thisName = this.subunits.get(this.representative).getName();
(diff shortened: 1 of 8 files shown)


lafita:
Even if two chains are from the same entity, they might have different number of residues in the crystal (e.g. disorder). How do you make sure that their alignment is correct when adding them into the cluster?
I think that was the reason why I did not use entities originally and just aligned the two sequences if the they were not identical.

biojava-structure/src/main/java/org/biojava/nbio/structure/cluster/SubunitCluster.java
```diff
@@ -252,12 +254,49 @@ public boolean mergeIdenticalByEntityId(SubunitCluster other) {
...
+
+			Group g = thisSub.getRepresentativeAtoms()[thisIndex].getGroup();
+
+			int seqresIndex = entityInfo.getAlignedResIndex(g, thisChain);
+
+			Group otherG = otherChain.getSeqResGroups().get(seqresIndex - 1);
+
+			if (!otherChain.getAtomGroups().contains(otherG)) {
+				// skip residues that are unobserved in other sequence ("gaps" in the entity alignment)
+				continue;
```
lafita: Ok I think that answers my question before, makes sense! Thanks Jose

PR #867, opened 2020-04-12
New interface finder functionality

josemduarte:
New feature to find interfaces in a given`Structure` object. 

Some doc corrections are included.


josemduarte:
@lafita it seems that test `TestQuatSymmetryDetectorExamples.testLocal` is failing because the bioassemblies in 3R8R were modified in January 2020. Could you have a quick look how to modify the test to pass?

Also would appreciate a quick check on this PR :)

PR #870, opened 2020-04-15
 Add option to return sequences from the Genbank file as a stream of sequences

emckee2006:


biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenbankReader.java
```diff
@@ -153,35 +154,56 @@ public GenbankReader(
...
 	}
+	
+	public Stream<S> getSequencesAsStream() {
+		return Stream.generate(() -> {
+			try {
+				return getSequence();
+			} catch (IOException | CompoundNotFoundException e) {
+				// TODO Auto-generated catch block
+				e.printStackTrace();
+				return null;
```
josemduarte, line 178: I'd rather throw the exception forward and let the caller handle it. Returning null means all callers have to handle the null and that they lose the ability to provide a good error message.

josemduarte:
Could you merge master into your branch? The tests should pass then. A problem with them was fixed in #871

emckee2006:
I thought I had... I'll double check.

As far as the exception goes, can you actually pass that exception on? If I
try, Eclipse tells me it's an error...

emckee2006:
I successfully merged the new stuff in.

There are ways to get the exception out of the lambda. They are a bit
tedious, but if it is helpful, I can give that a try...

josemduarte:
There seems to be an issue with the new test. See logs from travis, there's a timeout after running GenbankReaderTest

emckee2006:
I'll take a look... if I run it in junit in eclipse it runs fine... that's
odd

emckee2006:
How do i rerun Travis after adding a new commit? (or what is a better debugging mechanism here?)

josemduarte:
I've just triggered the travis tests again

emckee2006:
I commented out the new test to verify that it is the cause. Is there an
easy way to debug?

emckee2006:
OK, it still hangs even if not executing my test...

lafita:
> I commented out the new test to verify that it is the cause. Is there an easy way to debug?

Can you try to run `mvn install` locally from the command line or using Maven Install in eclipse?

biojava-core/src/test/java/org/biojava/nbio/core/sequence/io/GenbankReaderTest.java
```diff
@@ -163,6 +164,27 @@ public void testProcess() throws Exception {
 		assertEquals(3, dnaSequence.getAccession().getVersion().intValue());
 		assertTrue(genbankDNA.isClosed());
 	}
+	
+	@Test
+	public void testSequenceStream() {
+		CheckableInputStream inStream = new CheckableInputStream(this.getClass().getResourceAsStream("/two-dnaseqs.gb"));
```
lafita, line 170: Did you also include this file in the test resources directory? It seems not to be in this PR

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenbankReader.java
```diff
@@ -153,35 +154,56 @@ public GenbankReader(
 		}
 
 		LinkedHashMap<String,S> sequences = new LinkedHashMap<>();
-		@SuppressWarnings("unchecked")
 		int i=0;
 		while(true) {
```
lafita, line 158: This is what is hanging the build. Because the stream returned is null (the file does not exist, see comment below), sequence is always null and this while loop never finishes.

The while loop is not needed anymore, simply take an action if the sequence is null.

emckee2006:
Isn't that the same file used by one of the existing tests?

lafita:
True the file exists, but the while loop is not needed anymore and probably the one causing the timeout in travis. @emckee2006 does maven install work locally for you?

emckee2006:
Why doesn't that method need the while loop? It still needs to support
grabbing multiple sequences...

emckee2006:
This needs help... closing for now...

Issue #855, opened 2019-11-22
make or read genbank file error when feature spans zero point of circular sequence


chenzhiw:
Download a [genbank file](https://media.addgene.org/snapgene-media/v1.6.2-0-g4b4ed87/sequences/74/51/47451/addgene-plasmid-37845-sequence-47451.gbk) from [this page](http://www.addgene.org/37845/sequences/#depositor-full).
Read it  with biojava and error as below:

Exception in thread "main" java.lang.IllegalStateException: Start (7028) is greater than end (286); this is an incorrect format
	at org.biojava.nbio.core.sequence.location.template.AbstractLocation.assertLocation(AbstractLocation.java:133)
	at org.biojava.nbio.core.sequence.location.template.AbstractLocation.<init>(AbstractLocation.java:119)
	at org.biojava.nbio.core.sequence.location.template.AbstractLocation.<init>(AbstractLocation.java:92)
	at org.biojava.nbio.core.sequence.location.SimpleLocation.<init>(SimpleLocation.java:64)
	at org.biojava.nbio.core.sequence.location.SimpleLocation.<init>(SimpleLocation.java:55)
	at org.biojava.nbio.core.sequence.location.InsdcParser.parseLocationString(InsdcParser.java:241)
	at org.biojava.nbio.core.sequence.location.InsdcParser.parseLocationString(InsdcParser.java:189)
	at org.biojava.nbio.core.sequence.location.InsdcParser.parse(InsdcParser.java:143)
	at org.biojava.nbio.core.sequence.io.GenbankSequenceParser.parse(GenbankSequenceParser.java:283)
	at org.biojava.nbio.core.sequence.io.GenbankSequenceParser.getSequence(GenbankSequenceParser.java:402)
	at org.biojava.nbio.core.sequence.io.GenbankReader.process(GenbankReader.java:153)
	at org.biojava.nbio.core.sequence.io.GenbankReader.process(GenbankReader.java:121)
	at org.biojava.nbio.core.sequence.io.GenbankReaderHelper.readGenbankDNASequence(GenbankReaderHelper.java:178)
	at org.biojava.nbio.core.sequence.io.GenbankReaderHelper.readGenbankDNASequence(GenbankReaderHelper.java:190)

chenzhiw:
error part as below:

     CDS             complement(7028..286)
                     /codon_start=1
                     /gene="bla"
                     /product="beta-lactamase"
                     /label=AmpR

josemduarte:
[...]

PR #872, opened 2020-04-27
#855 make or read genbank file error when feature spans zero point of circular sequence

paolopavan:
This PR addresses issue #855

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/sequence/location/InsdcLocations.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/sequence/location/InsdcLocations.java
@@ -147,6 +147,10 @@ public class InsdcLocations {
 			this(Arrays.asList(subLocations));
 		}
 
+		public GroupLocation(boolean isCircular, Location... subLocations) {
+			this(Arrays.asList(subLocations));
+		}
+
 		public GroupLocation(Point start, Point end, Strand strand,
 				boolean circular, Location... subLocations) {
 			super(start, end, strand, circular, subLocations);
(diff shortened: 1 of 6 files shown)


biojava-core/src/main/java/org/biojava/nbio/core/sequence/location/InsdcLocations.java
```diff
@@ -147,6 +147,10 @@ public GroupLocation(Location... subLocations) {
 			this(Arrays.asList(subLocations));
 		}
 
+		public GroupLocation(boolean isCircular, Location... subLocations) {
```
josemduarte, line 150: The new parameter `isCircular` is not used.

josemduarte:
@chenzhiw  since you opened the original issue #855 ,  could you have a look at this?

biojava-core/src/main/java/org/biojava/nbio/core/sequence/location/InsdcLocations.java
paolopavan: Good catch, thank you.
Done.

paolopavan:
By the way, I may just miss some pieces or a different way to represent it, does AbstractSequence misses a kind of isCircular property? Would it be good to add?

paolopavan:
> @chenzhiw since you opened the original issue #855 , could you have a look at this?

Note that the very same sequence mentioned in the issue has been added in test for the new feature.
Still I am wondering on how many sequences out there fall into this case. Any opinion?

josemduarte:
I will merge this by end of week if there are no other comments.

I'll then produce a new release.

heuermh:
Thank you @josemduarte @paolopavan, I will review and test this week.

josemduarte:
@heuermh any comments? or shall we merge?

sbliven:
Looks good to me. Thanks @paolopavan!

PR #873, opened 2020-04-30
#855 code cleaning not related to issue

paolopavan:
This PR contains general code cleaning not specifically related to the issue. Just a bit of improving quality of the files touched by #855, left apart to ease the review of the main issue.

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/sequence/features/DatabaseReferenceInterface.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/sequence/features/DatabaseReferenceInterface.java
@@ -24,6 +24,8 @@ package org.biojava.nbio.core.sequence.features;
 
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
 
 /**
  * If a SequenceProxyReader implements this interface then that external source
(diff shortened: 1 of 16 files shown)


heuermh:
> Anyone not ok with that?

Could we run a binary compatibility check using [Clirr](https://www.mojohaus.org/clirr-maven-plugin/) or another similar tool?  The problem is not that changes are big or small, it is rather that users run may run into `IncompatibleClassChangeError` and related exceptions at runtime.

josemduarte:
@heuermh  The changes here would mean a compile time error for downstream users (that's why it should be a Major version bump). 

I was simply saying that for this case it's a minimal change since the interface is mostly conserved when switching `ArrayList`->`List` or `HashMap`->`Map`. So it would be an easy-to-fix compile time error.

paolopavan:
You are right @josemduarte , I didn't put too much attention to that since I had the feeling I was not changing API exposed classes, but eventually this is not the case, sorry about that.
Indeed I got lost in this cleaning and unfortunately it is also far away to be completed.

And yes, when I started I didn't suppose to make a lot of changes and unfortunately I didn't create another branch from master. That would have prevented those changes on the top of #872 but in fact my initial purpose was just cleaning touched files. Unfortunately I don't think what you ask is a very easy task. 
It would be an option merging #872 first and #873 after?

josemduarte:
Ok, thanks @paolopavan . Indeed, I think we can merge #872 to include it in next release (5.4.0), whilst keeping this one open and defer for next release. 

We are likely going to need a 6.0.0 release soon for other reasons, so we can do all these API changes together.

paolopavan:
Fair enough, @josemduarte .
There is a roadmap for 6.0.0?

josemduarte:
Ok I'll close the PR. @paolopavan please make a new one after the 5.4.0 release happens.

PR #875, opened 2020-05-21
Enable access to the full genbank record

emckee2006:


biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenbankSequenceParser.java
```diff
@@ -440,4 +442,22 @@ public void parseFeatures(AbstractSequence<C> sequence) {
 	public CompoundSet<?> getCompoundType() {
 		return compoundType;
 	}
+
+
+
+	/**
+	 * @return the srcRecord
+	 */
+	public String getSrcRecord() {
```
josemduarte, line 451: It'd help to have javadocs here. What exactly is the source record? The whole genbank file? a part of it? what format?

biojava-core/src/main/java/org/biojava/nbio/core/sequence/template/AbstractSequence.java
```diff
@@ -732,5 +734,18 @@ public int countCompounds(C... compounds) {
 		return SequenceMixin.inverse(this);
 	}
 
+	/**
+	 * @return the srcRecord
+	 */
+	public String getSrcRecord() {
```
josemduarte, line 740: Same comment about javadocs

sbliven:
What's the memory footprint of this change? Does it double the storage size?

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenbankSequenceParser.java
sbliven: Thou hast a typo :)

sbliven:
What's the use case for this? It seems very strange to me for a parser to store the unparsed input.

emckee2006:
I must be missing the typo...

emckee2006:
We are stepping through the refseq/genbank release files one record at a
time. We do some checksumming and such in an attempt to determine if the
record has been modified since we last loaded it to our database...

heuermh:
> What's the use case for this? It seems very strange to me for a parser to store the unparsed input.

I would have to agree here.

Is it possible for this use case to break the parser up into a record splitter (which one could use to checksum the entire record text) and a record tokenizer (which then parses the record into the java objects)?

emckee2006:
Or add in a flag to turn off the full record storing?

sbliven:
I'd suggest leaving this out of biojava. I think you can implement a splitter class as @heuermh suggested, and then use biojava to parse the split records.

emckee2006:
I'm planning on something like that... it does sound like a better
approach.

josemduarte:
Thanks @emckee2006 and thanks all for reviewing. I'll close then

PR #876, opened 2020-06-11
refactored code to use java 8 streams

balaji-s:
Hello-
I have written three for loops in java 8 streams, if you like it, kindly merge. Thanks

### DIFF
--- a/biojava-aa-prop/src/main/java/org/biojava/nbio/aaproperties/CommandPrompt.java
+++ b/biojava-aa-prop/src/main/java/org/biojava/nbio/aaproperties/CommandPrompt.java
@@ -327,9 +327,7 @@ public class CommandPrompt {
 			}
 		}
 		output.print(header.replace(delimiter, "_"));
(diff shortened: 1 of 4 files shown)


biojava-aa-prop/src/main/java/org/biojava/nbio/aaproperties/CommandPrompt.java
```diff
@@ -327,9 +327,7 @@ private static void compute(PrintStream output, String header, String sequence,
 			}
 		}
 		output.print(header.replace(delimiter, "_"));
-		for(int i = 0; i < dList.size(); i++){
-			output.print(delimiter + Utils.roundToDecimals(dList.get(i), decimalPlace));
-		}
+		dList.stream().forEach(item -> System.out.print(delimiter + Utils.roundToDecimals(item, decimalPlace)));
```
josemduarte: This is not the same. The original code was printing to `output` and not to `System.out`

PR #878, opened 2020-06-27
Case insensitive chem.comp type lookup

valasatava:
mmtf has it all upper-case - fails to lookup for chem.comp type because of different case


pwrose:
Thanks, @valasatava for fixing the issue and @lafita and @joseduarte for
the review!

PR #880, opened 2020-06-30
#855 code cleaning not related to issue

paolopavan:
As per our conversation.
Cheers,
Paolo

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/sequence/features/DatabaseReferenceInterface.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/sequence/features/DatabaseReferenceInterface.java
@@ -24,6 +24,8 @@ package org.biojava.nbio.core.sequence.features;
 
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
 
 /**
  * If a SequenceProxyReader implements this interface then that external source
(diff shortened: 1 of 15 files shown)


paolopavan:
Thank you @josemduarte !

heuermh:
Do these changes need a cycle of `@deprecated` javadoc tag/`@Deprecated` annotations in a 5.x release and then `@since 6.0.0` javadoc tags when merged to the main branch?

paolopavan:
Well, the breaking changes are pretty trivial, will require just moving from implementation to interface in a very specific case but there is no need to change the structure of the client code or even a method renaming.
A deprecated annotation may have been a good idea but was to send out with the already pushed version? May worth to push a new minor version just for that?
Anyway this change is not terrible and is also very limited. If I should be honest, I would commit much more reworking of this kind, in my opinion. If you agree, there is a chance that I could pick something more.

josemduarte:
I agree with @paolopavan . These changes are trivial. Any downstream project using the code would find it trivial to fix the compilation errors by changing the interfaces. I don't think this deserves deprecation tagging. In my opinion it only needs appropriate documentation in CHANGELOG.

In general I am not a fan of deprecation tagging, except for some very clear cut cases (e.g. method renames). I think good docs in CHANGELOG or a migration guide are a lot more useful. Deprecations can become a liability with time, because they can be forgotten or become a source of bugs. By the time they need to be removed (after the next major release), developers are likely to not understand anymore why they were deprecated in the first place.

josemduarte:
Any other comments? If not, I'll merge tomorrow

PR #882, opened 2020-07-06
Removing code that uses deprecated web service source.rcsb.org

josemduarte:
This removes quite a lot of legacy code from the structure module. The change is needed because RCSB is deprecating the All-vs-All structure alignment system and the source.rcsb.org web services that provided domain information.

The most important changes are:

- removal of code related to All-vs-All data retrieval and calculation
- removal of PDP and SCOP providers that depended on source.rcsb.org
- removal of JFatCatClient, a client for the webservices at source.rcsb.org
- bump to 6.0.0-SNAPSHOT to make clear that master branch is dev towards next major release
- a stub of 6.0.0 changelog to document the changes from this PR


biojava-structure/src/test/java/org/biojava/nbio/structure/align/ce/CeCPMainTest.java
```diff
@@ -405,16 +407,31 @@ public void testCalculateMinCP() throws SecurityException, NoSuchMethodException
...
-		CeCPMain algorithm = new CeCPMain();
+		// since BioJava 6.0.0, there's no PDP provider. The below corresponds to domain "PDP:3A2KAc"
+		List<ResidueRange> ranges = new ArrayList<>();
+		ranges.add(new ResidueRange("A", new ResidueNumber("A",234, null), new ResidueNumber("A", 333, null)));
+		SubstructureIdentifier ssi = new SubstructureIdentifier(pdb1, ranges);
+		Structure structure1 = cache.getStructure(pdb1);
+		ssi.reduce(structure1);
 
-		AtomCache cache = new AtomCache();
+		// since BioJava 6.0.0, there's no RemoteSCOP provider. The below corresponds to domain "d1wy5a2"
```
lafita, line 424: Does this mean it is no longer possible to pass a SCOP identifiers directly?

josemduarte: The BerkeleyScopInstallation should still be working. The RemoteScopInstallation was grabbing all data from source.rcsb.org and there was no other solution than removing it.

For the test I wanted to avoid external resources as much as possible, so I directly hard-coded the residue range. I'm not sure how exactly the Berkeley installation works, but most likely it would have to download a lot of data, which wouldn't have been good for the test.

lafita: Ok, that is what I imagined but I wasn't sure why you were using hard-coded ranges. Makes sense!

sbliven:
It's sad to see all-vs-all disappear. I was proud of that project, although I would approach it differently if I did it again.

Is there a replacement for the domain providers? I thought that was a useful service. Mashing SCOP+PDP was a little inconsistent, but we really need 100% coverage for many applications, and I'm not aware of any third parties that can provide that.

biojava-structure/src/test/java/org/biojava/nbio/structure/align/ce/CeCPMainTest.java
sbliven: Yes, Berkeley provides all data in flat files, so it's a big download.

josemduarte:
@sbliven the RemoteScopInstallation was essentially a client of the provider http://source.rcsb.org. Once the provider is gone, the client doesn't make sense anymore. 

However I did not remove the interfaces (`ScopDatabase` and `LocalScopDatabase`). The day someone starts providing a service for scop domains, we could implement a new scop installation in biojava. 

> Did you check the impact of this on test times? We used the RemoteScopInstallation pretty extensively to avoid downloading the full database files.

It seems that the only test that was using it was the one I modified (see comments from @lafita above). But I will double check, perhaps I'm missing something

sbliven:
@josemduarte 

> @sbliven the RemoteScopInstallation was essentially a client of the provider http://source.rcsb.org. Once the provider is gone, the client doesn't make sense anymore.
>
>However I did not remove the interfaces (ScopDatabase and LocalScopDatabase). The day someone starts providing a service for scop domains, we could implement a new scop installation in biojava.

I was partially mourning the deprecation of the provider (but other parts of source.rcsb.org were ripe for retirement). There's not a good model for open source services; no one can fork source.rcsb.org and continue hosting it. I wonder if OBF might be able to take a role providing this (e.g. if someone ported the domain provider to a microservice).

> It seems that the only test that was using it was the one I modified (see comments from @lafita above). But I will double check, perhaps I'm missing something

I was thinking about all the tests that use scop domains: OptimalCECPMainTest, CeCPMainTest, TestStructureName, AtomCacheTest, FastaAFPChainConverterTest, TestSequenceFunctionOrderDetector. However, these seem to be using the default ScopInstallation, so they shouldn't add additional overhead (beyond BerkeleyScopInstallationTest, which it seems is already downloading 4 scop versions).

josemduarte:
Ok I've just double checked and it seems that all these tests use the scop provider (I ran the tests on master while source.rcsb.org was down):
```
[ERROR] Errors:
[ERROR]   CeCPMainTest.testCECP1:417 » Runtime Unable to reach http://source.rcsb.org/jf...
[ERROR]   AtomCacheTest.testGetStructureForChainlessDomains:171 » Runtime Unable to reac...
[ERROR]   AtomCacheTest.testGetStructureForDomain1:109 » Runtime Unable to reach http://...
[ERROR]   AtomCacheTest.testGetStructureForDomain2:128 » Runtime Unable to reach http://...
[ERROR]   AtomCacheTest.testGetStructureForDomain3:152 » Runtime Unable to reach http://...
[ERROR]   FastaAFPChainConverterTest.testBug1:137 » Runtime Unable to reach http://sourc...
[ERROR]   TestSequenceFunctionOrderDetector.testGetSymmetryOrder:60 » Runtime Unable to ...
```
These tests do work in this branch, which means that berkeley scop provider works. In any case I'm going to try to remove the dependency on berkely scop in tests, by hard-coding the residue ranges corresponding to scop domains (what I did for other tests already)

PR #884, opened 2020-08-03
Remove code that uses legacy RCSB PDB APIs

josemduarte:
The RCSB PDB legacy APIs at `rcsb.org/pdb/rest` are to be shutdown in November 2020 (see [announcement](https://www.rcsb.org/news?year=2020&article=5eb18ccfd62245129947212a&feature=true)). This removes biojava code that made use of those APIs

Also removing PDB xml validation report parsing code that is now obsolete because the schema was upgraded.


josemduarte:
Thanks for reviewing @sbliven 

The build didn't work because a test was broken after carbohydrate remediation project went out (https://www.rcsb.org/news?year=2020&article=5f17634f0b582626e650e48b&feature=true). Now fixed.

I'll merge.

PR #885, opened 2020-08-12
Upgrade to latest ciftools

JonStargaryen:
- upgrade ciftools version (dedicated release that supports Java 8, properly supports arbitrary schemata)
- removing the old mmCIF parser requires some discussion and is probably better of in a separate PR


biojava-structure/src/test/java/org/biojava/nbio/structure/io/cif/CifFileSupplierTest.java
```diff
@@ -0,0 +1,192 @@
...
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Arrays;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+public class CifFileSupplierTest {
```
josemduarte, line 27: It'd be better to put this test under the integration test module. We'd like to have a clearer separation of unit tests and integration tests going forward: the structure module should have unit tests, and the integrationtest module should have integration tests.

JonStargaryen: Seems like we had that discussion already. Just saw that the same test exists in the integrationtest module :) (https://github.com/biojava/biojava/blob/master/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/io/cif/CifFileSupplierIntegrationTest.java). I deleted the redundant copy in biojava-structure.

PR #890, opened 2020-09-17
Add GFF3Reader overload that accepts Path

jean-philippe-martin:
The String-based functions are still here, and almost all the code is shared.
The Path overload allows everyone to read files via NIO providers, such as e.g.
the GCS NIO provider for reading files on Google Cloud Storage.

Key change: new BufferedReader(new FileReader(filename)) -> Files.newBufferedReader(path)

### DIFF
--- a/biojava-genome/src/main/java/org/biojava/nbio/genome/parsers/gff/GFF3Reader.java
+++ b/biojava-genome/src/main/java/org/biojava/nbio/genome/parsers/gff/GFF3Reader.java
@@ -20,6 +20,9 @@
  */
 package org.biojava.nbio.genome.parsers.gff;
 
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


josemduarte (approved):
Nice addition, thank you!

A side note: if you see opportunity to improve interfaces related to this and there's a good case for it, now is a good time because we are working towards a 6.0.0 (breaking changes) release.

jean-philippe-martin:
Thank you @josemduarte! I see I do not have write access, so you'll have to do the merge (or squash+merge) for me it looks like. Let me know whether there's something you'd like me to do to help.

I'm sure there are other places that are using File or String instead of Path, but someone else will have to update those.

Good luck for the 6.0.0 release!

josemduarte:
Yes merges have to come from developers with write access. There's a few of us. In any case we leave PRs opened for a few days to allow enough time for anyone else to review.

sbliven (approved):
Nice addition, thanks!

PR #893, opened 2020-10-19
Added github actions to verify the build on each push

HanSolo:
On each push mvn verify will be executed on Ubuntu 18.04 and the latest Mac OS X using JDK 11

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,6 +21,7 @@ jobs:
         with:
           java-version: ${{ matrix.java }}
       - name: Test with Maven
-        run: mvn test -B --file pom.xml
+        #run: mvn test --batch-mode verify
+        run: mvn verify
 
 ...
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,26 @@
+---
+name: Java CI
+
+on: [push]
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


josemduarte:
Thanks, this is great!

A few questions, since I'm not familiar with github actions:

- how is this different from travis CI? are they complementary? or can this cover everything that travis does?
- what does `mvn verify` do?

HanSolo:
Hi there,
Yep it can do in principle everything that travis can do just directly on github which is really nice.
Maven mvn verify just runs any checks on results of integration tests. So there is no need to run mvn clean install when you can run mvn verify to run your integration tests.

josemduarte:
Thanks for explaining and sorry for the delay. I'm not finding a lot of time for biojava these days...

One thing about integration tests: we do have integration tests, but we have set them up as a separate module (biojava-integration-tests). I think we should try to move towards something like [this blog post explains](https://www.petrikainulainen.net/programming/maven/integration-testing-with-maven/) (integration-tests as part of each module in a separate dir from unit tests).

As things are at the moment, if I understand it right `mvn verify` will simply run all tests. I think that should be fine for now.

josemduarte (approved):
I'd say we can try this setup and gain some experience on how it works. Perhaps we can replace travis by this later on.

sbliven:
Interesting blog post. However the argument against our current integration-test-as-module approach seems to be that it's hard to set up. Since it's working fine for us now I'm not sure what we gain from refactoring.

PR #900, opened 2020-11-16
fix url null pointer issue

JoshSu:
I used mockito to mock the PDBStatus class in order to fix the null pointer url issue from following link 

https://travis-ci.org/github/biojava/biojava/builds/743534185

After the change, error got fixed 

https://travis-ci.org/github/JoshSu/biojava


josemduarte (commented):
I think the test failed originally because the external resource was down.

In any case, it looks like with this change the test does not actually do any testing of the functionality but instead just fakes the return.

PR #902, opened 2020-11-24
Rewrite of PDB status to use new RCSB PDB data API

josemduarte:
This replaces the legacy RCSB PDB API by the current one, so that `PDBStatus` continues working after legacy APIs are shutdown in December.

This is a breaking change with a changed `Status` enum and 2 fewer methods (getReplace, getReplaces). All documented in CHANGELOG

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/PDBStatus.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/PDBStatus.java
@@ -70,7 +70,7 @@ public class PDBStatus {
 		 * @throws IllegalArgumentException If the string is not recognized
 		 */
 		public static Status fromString(String statusStr) {
-			if (statusStr == null) return null;
+			if (statusStr == null) throw new IllegalArgumentException("Status string can't be null");
 			if(statusStr.equalsIgnoreCase("REMOVED"))
 				return Status.REMOVED;
 			else if(statusStr.equalsIgnoreCase("CURRENT"))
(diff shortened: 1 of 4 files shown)


biojava-structure/src/main/java/org/biojava/nbio/structure/PDBStatus.java
```diff
@@ -18,134 +18,66 @@
...
-import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.InputStreamReader;
 import java.net.URL;
 import java.util.*;
 
 /**
- * Methods for getting the status of a PDB file (current, obsolete, etc)
+ * Methods for getting the status of a PDB file (current, obsolete, unreleased)
```
JonStargaryen, line 36: I'd change 'obsolete' to 'removed'.

biojava-structure/src/main/java/org/biojava/nbio/structure/PDBStatus.java
```diff
@@ -18,134 +18,66 @@
...
 
 		/**
-		 *
-		 * @param statusStr
-		 * @return
 		 * @throws IllegalArgumentException If the string is not recognized
 		 */
 		public static Status fromString(String statusStr) {
 			Status status;
 			String statusStrUpper = statusStr.toUpperCase();
```
JonStargaryen, line 74: Nitpicking here: I'd remove the `toUpperCase` call (as `equalsIgnoreCase` is used anyway) and change the order to `if("REMOVED".equalsIgnoreCase(statusStr))`. That way `statusStr` can be `null` which is nice to have for a method with public visibility.

JonStargaryen (approved):
Thanks, LGTM. 2 tiny comments to consider.

josemduarte:
Thanks for reviewing. I've pushed the requested changes.

PR #903, opened 2020-11-25
Flaky Test Fixed in TestStructureSerialization

Vidishab18:
The test `org.biojava.nbio.structure.io.TestStructureSerialization#testSerializeStructure` is flaky due to the order in which Methods are appended into the StringBuilder `str.append(pdbHeader);` which is built in the toString method of the StructureImpl class.

The flakiness is observed due to the non-deterministic order of pdbHeader that can produce different structures.

The flaky test is fixed by sorting the array of Methods in PDBHeader.java class using a custom comparator in the toString method.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/PDBHeader.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/PDBHeader.java
@@ -108,6 +108,7 @@ public class PDBHeader implements PDBRecord {
 			Class<?> c = Class.forName(PDBHeader.class.getName());
 			Method[] methods  = c.getMethods();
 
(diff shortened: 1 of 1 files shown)


josemduarte (approved):
LGTM, thanks!

Vidishab18:
@josemduarte Thank you. Is this good to be merged ?

sbliven:
Thanks for the contribution, @Vidishab18!

sbliven:
I merged this quickly, but after reviewing #905 I'm worried about it. This changes the order of mmcif columns in the pdbheader section. This order seems important. I think I will roll this back until we have more discussion.

josemduarte:
Thanks for rolling back @sbliven . I did not see the possible side issues. It will require more reviewing/discussion.

PR #906, opened 2020-12-01
Fixing flaky test in AlignmentToolsTest.java

Vidishab18:
The test `AlignmentToolsTest#testToConciseAlignmentString` is flaky due to the order dependence in `AlignmentTools` class `toConciseAlignmentString` method in `S seedNode = alig.keySet().iterator().next();` to which the HashMap is passed from the test `testToConciseAlignmentString`.  Due to the non-deterministic order of iteration, it can be produce different structures since the Oracle specification about HashMap says that "This class makes no guarantees as to the order of the map; in particular, it does not guarantee that the order will remain constant over time." The documentation is here for your reference:https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html

The retrieval of the elements using keySet() can be non-deterministic for a HashMap and can cause `AlignmentToolsTest#testToConciseAlignmentString` to be a flaky test. The HashMap is now replaced with the LinkedHashMap so that the non-deterministic iteration order of HashMap is eliminated.

### DIFF
--- a/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/align/AlignmentToolsTest.java
+++ b/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/align/AlignmentToolsTest.java
@@ -320,7 +320,7 @@ public class AlignmentToolsTest {
 		String result,expected;
 		int i=0;
 
-		test = new HashMap<Integer, Integer>();
+		test = new LinkedHashMap<Integer, Integer>();
 		test.put(1, 2);
 		test.put(2, 3);
 		test.put(3, 4);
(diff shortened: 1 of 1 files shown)


sbliven:
I created #907 to discuss the issue behind this PR.

sbliven (changes requested):
Did you run this patch against your test tool? The first thing `toConciseAlignmentString` does is copy the input to a HashMap, so changing the input to LinkedHashMap should not make a difference.

I think a better solution is to copy input into a TreeMap within `toConciseAlignmentString` to ensure ordered enumeration.

josemduarte:
This has been inactive a long time. I'll close

PR #908, opened 2020-12-07
Ensure deterministic order in toConciseAlignmentString

### LINKED ISSUE #907: toConciseAlignmentString is non-deterministic
In #906, @Vidishab18 showed some test cases using [AlignmentTools.toConciseAlignmentString](https://github.com/biojava/biojava/blob/e3e3a1f50964b4fc076c47cbe8a05ea97720a0e0/biojava-structure/src/main/java/org/biojava/nbio/structure/align/util/AlignmentTools.java#L990) can produce different results depending on HashMap iteration order.

The method takes an alignment map (a way of representing a pairwise alignment as a map between residue indices) and outputs a string representation. The representation is not unique. For instance "1>2>3>2 4>3" and "4>3 1>2>3>2" are the same alignment. Likewise, "1>2>1" and "2>1>2" are the same alignment. Outputs potentially hinge on the enumeration order of the underlying map.

Possible solutions:
1. Create a canonical form (e.g. sorted some way). I'm not sure whether this is feasible; general graph canonization is computationally hard, but perhaps the special structure of alignment graphs gives an obvious order
2. Use a SortedMap (TreeMap) internally. This should make it deterministic, although perhaps not in a simple form
3. Merge #906, which apparently fixes the tests even though it leaves the function itself unchanged.

The method was intended as a concise human-readable output of the map, so the order doesn't really matter. Option 2 appears to be the simplest solution. Performance on this method is not critical.

sbliven:
Fixes #907. Uses Option 2 (TreeMap)

This is an alternative to #906.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/ZipChemCompProvider.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/ZipChemCompProvider.java
@@ -293,7 +293,7 @@ public class ZipChemCompProvider implements ChemCompProvider{
 		*/
 
 		// Copy in each file.
-		try (FileSystem zipfs = FileSystems.newFileSystem(zipFile, null)) {
+		try (FileSystem zipfs = FileSystems.newFileSystem(zipFile, (ClassLoader) null)) {
 			Files.createDirectories(pathWithinArchive);
 			for (File f : files) {
 				if (!f.isDirectory() && f.exists()) {
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/align/util/AlignmentTools.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/align/util/AlignmentTools.java
@@ -20,7 +20,37 @@
  */
 package org.biojava.nbio.structure.align.util;
 
-import org.biojava.nbio.structure.*;
+import java.io.IOException;
+import java.io.Writer;
+import java.util.AbstractMap;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeMap;
+import java.util.TreeSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import javax.vecmath.Matrix4d;
+
+import org.biojava.nbio.structure.Atom;
+import org.biojava.nbio.structure.AtomImpl;
+import org.biojava.nbio.structure.Calc;
+import org.biojava.nbio.structure.Chain;
+import org.biojava.nbio.structure.ChainImpl;
+import org.biojava.nbio.structure.Group;
+import org.biojava.nbio.structure.ResidueNumber;
+import org.biojava.nbio.structure.Structure;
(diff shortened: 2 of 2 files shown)


sbliven:
@Vidishab18 Could you please run your tool against this version and verify that `AlignmentToolsTest#testToConciseAlignmentString` is deterministic now?

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/ZipChemCompProvider.java
```diff
@@ -293,7 +293,7 @@ private synchronized boolean addToZipFileSystem(Path zipFile, File[] files, Path
 		*/
 
 		// Copy in each file.
-		try (FileSystem zipfs = FileSystems.newFileSystem(zipFile, null)) {
+		try (FileSystem zipfs = FileSystems.newFileSystem(zipFile, (ClassLoader) null)) {
```
sbliven, line 296: This line fixes an error I was getting compiling with Java 1.8:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project biojava-structure: Compilation failure
[ERROR] /Users/bliven_s/git/biojava_ws/biojava/biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmcif/ZipChemCompProvider.java:[296,52] reference to newFileSystem is ambiguous
[ERROR]   both method newFileSystem(java.nio.file.Path,java.lang.ClassLoader) in java.nio.file.FileSystems and method newFileSystem(java.nio.file.Path,java.util.Map<java.lang.String,?>) in java.nio.file.FileSystems match
```

The second form isn't documented, so it could be that this change will cause a warning with different JDKs.

Vidishab18: @sbliven 
The change to TreeMap still shows the test to be flaky. I did try this initially, but it doesn't fix the flaky error.

Vidishab18: @sbliven 
I did initially try to fix this error with the same change. However the TreeMap here, is not fixing this flaky error. The tool still marks it as flaky with error of the sort -
```
[ERROR] Failures: 
[ERROR]   AlignmentToolsTest.testToConciseAlignmentString:342 1. Cycles. expected:<[1>2>3>1] 7>7> but was:<[2>3>1>2] 7>7>
[INFO] 
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0
```

Vidishab18: @sbliven @josemduarte 
Would like just to link the Nondex tool here : https://github.com/TestingResearchIllinois/NonDex

The tool requires to add the plugin as specified in the documentation and run with maven for testing. Ideally I would run this as - 
` mvn -pl biojava-integrationtest nondex:nondex -Dtest=AlignmentToolsTest#testToConciseAlignmentString`

It could be used to run for the whole module as well -
`mvn -pl biojava-integrationtest nondex:nondex`.

This could be added into travis.yml, so it can be run on new tests too.

It also has the debug feature that can be used to point to the file(s) that could require changes. The link to the tool has all the documentation.

sbliven:
It looks like more effort is needed to make this deterministic.

sbliven:
I intend to figure this out (and #909) but I got sidetracked this week. It might not get fixed until after Christmas.

PR #913, opened 2021-01-29
Performance improvement in ASA calculation

josemduarte:
Implemented some of the ideas exposed in Eisenhaber et al 1994 (https://onlinelibrary.wiley.com/doi/epdf/10.1002/jcc.540160303). This results in ~x5 runtime improvement


JonStargaryen (approved):
LGTM. Thank you for this set of optimizations!

PR #915, opened 2021-02-22
Minor updates to fastq package

heuermh:


biojava-genome/src/test/java/org/biojava/nbio/genome/io/fastq/FastqBuilderTest.java
```diff
@@ -36,6 +36,41 @@ public void testConstructor()
...
+	public void testConstructorNullFastq()
+	{
+		try
+		{
+			new FastqBuilder(null);
+			Assert.fail("builder(null) expected IllegalArgumentException");
+		}
+		catch (IllegalArgumentException e)
+		{
+			// expected
```
josemduarte, line 70: I think it'd be cleaner to use the `Assert.assertThrows()` call available in junit 4.13+. See https://stackoverflow.com/questions/156503/how-do-you-assert-that-a-certain-exception-is-thrown-in-junit-4-tests/46514550#46514550

biojava-genome/src/test/java/org/biojava/nbio/genome/io/fastq/FastqTest.java
```diff
@@ -111,6 +111,20 @@ public void testBuilder()
...
+	public void testBuilderNullFastq()
+	{
+		try
+		{
+			Fastq.builder(null);
+			Assert.fail("builder(null) expected IllegalArgumentException");
+		}
+		catch (IllegalArgumentException e)
+		{
+			// expected
```
josemduarte, line 124: Same comment as above

josemduarte (changes requested):
Thanks! One comment above exception assertion

biojava-genome/src/test/java/org/biojava/nbio/genome/io/fastq/FastqBuilderTest.java
heuermh: I agree, and perhaps it'd be even more preferable to use `@Test(expected=IllegalArgumentException.class)` but this stuff is old and I followed the pattern used elsewhere in the same file

josemduarte: > more preferable to use @Test(expected=IllegalArgumentException.class)

That seems to be so before 4.13. I just read in the SO thread that from 4.13+, it is preferrable to do `Assert.assertThrows()`. Anyway since these tests are new, it's a good idea to follow the latest best practices.

josemduarte (approved):
Thank you!

heuermh:
Thank you, @josemduarte

PR #918, opened 2021-03-01
Fix #917: Symmetry doesn't display with non-US locales

### LINKED ISSUE #917: Symmetry doesn't display with non-US locales
Running CeSymm or QuatSymm on a system with a non-US locale breaks the nice Jmol axes & polygon display.

This is due to some locales (e.g DE) using commas in floats, which breaks Jmol parsing.

sbliven:
The core of this is the locale fix, which is necessary for CESymm/QuatSymm to function properly on non-US systems.

I've also included some commits for general cleanup and tests. If any of these are controversial I can revert them or move them to separate PRs.

### DIFF
--- a/biojava-structure-gui/pom.xml
+++ b/biojava-structure-gui/pom.xml
@@ -42,7 +42,7 @@
         <dependency>
             <groupId>net.sourceforge.jmol</groupId>
             <artifactId>jmol</artifactId>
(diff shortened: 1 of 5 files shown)


heuermh (approved):

josemduarte (approved):
LGTM, thanks!

PR #919, opened 2021-03-01
Github issue 843 - Genebank parser

### LINKED ISSUE #843: Biojava fails to parse Genbank and EMBL format
Biojava fails to parse `anticodon` and `transl_except` feature qualifiers when they line wrap. 
Biojava expects the values to be quoted, this is invalid.

This causes applications like Mauve and Colombo/SigiHMM to emit 
```
This line could not be parsed:                 seq:caa)
```
and discard large portions of the dataset.

This line is from 15584_genome.embl from Biopython: 
```
FT                   /anticodon=(pos:complement(1123552..1123554),aa:Leu,
FT                   seq:caa)
```
The matching line from 15584_genome.embl from Bioperl:
```
FT                   /anticodon="(pos:complement(1123552..1123554),aa:Leu,seq:ca
FT                   a)"
```

The difference between biopython and bioperl is that bioperl quotes anticodons if they wrap.

Copying NIH statement here for reference. Some messages were removed/edited for brevity.

> Case #: CAS-380254-C8B2Z2 
> The document here: http://www.insdc.org/files/feature_table.html does not describe how and where continuation lines should be created. There is divergence between biopython, bioperl, and other tools on how this is handled causing datasets to fail to be parsed.
> For example:
> The reference genome for NC_018080 downloaded from RefSeq fails to parse.
> Features containing:
>     /anticodon=(pos:complement(1123552..1123554),aa:Leu,
>     seq:caa)
> are failing to parse, this output is consistent with biopython.
> Bioperl wraps at column 80:
>     /anticodon="(pos:complement(1123552..1123554),aa:Leu,seq:caa
>     )"
> Because there is no official statement on how wrapping should occur I can't raise issues with the appropriate software developer.
> Any input on the matter would be appreciated. 
> Case Created:  6/6/2019 4:08 PM 
> 
> ----------------------------
> 
> Dear Nolan,
> Thanks for the additional info. It is expected that long anticodon qualifiers will wrap if they exceed the 80-character max line length. Furthermore, any of the following are legal within the spec:
>     /anticodon=(pos:complement(1123552..1123554),aa:Leu,
>     seq:caa)
>     /anticodon=(pos:complement(1123552..1123554),aa:Leu,seq:
>     caa)
>     /anticodon=(pos:complement(1123552..1123554),
>     aa:Leu,seq:caa)
> What might be happening is that the parsers you're testing may be combining the /anticodon qualifier and its continuation line, but adding an extra space, which would then invalidate the qualifier. That would be a bug in the parser code. For parsing an anticodon qualifier with a continuation line, you have to:
>     Realize that the you aren't yet at the end of the qualifier value (terminal right paren)
>     Read the next line, and strip all leading white space
>     Directly append (2) to the qualifier value
> For other qualifiers with text content, a parser generally needs to include a space when appending a continuation line, unless the row ends in a hyphen or is an unbroken string (no spaces, such as with a long chemical formula or the protein sequence in a /translation qualifier). But /anticodon may need an exception, depending on how the parser code is written.
> I hope that helps clear up the bug. If you file a bug report, please let us know the ticket for our records.
> 
> ...
> 
> As our flatfile does not have quotes, the quoting is not valid. The quotes are a bug. 
> 
> Best regards,
> -Terence Murphy, Ph.D.
> RefSeq and Gene Team Lead
> NCBI/NLM/NIH/DHHS
> 
> ----------------------------

https://github.com/bioperl/bioperl-live/issues/321

MaxGreil:
#### Reference Issue
Partly fixes #843 

#### What does this implement/fix? Explain your changes.
- Corrected GenbankReader.java to parse anticodon qualifiers when they line wrap.
- Wrote new testGithub843 in GenbankReaderTest.java
- Added mentioned file NC_018080.gb downloaded from https://www.ncbi.nlm.nih.gov/nuccore/NC_018080

#### Additional comments
@josemduarte , can you please tell me where I can get NC_018080 in .embl file format? I couldn't find it.

### DIFF
--- a/biojava-core/src/test/java/org/biojava/nbio/core/sequence/io/GenbankReaderTest.java
+++ b/biojava-core/src/test/java/org/biojava/nbio/core/sequence/io/GenbankReaderTest.java
@@ -333,7 +333,7 @@ public class GenbankReaderTest {
 	/**
 	 * Biojava fails to parse anticodon and transl_except feature qualifiers when they line wrap.
 	 * https://github.com/biojava/biojava/issues/843
-	 * File NC_018080.gb downloaded from https://www.ncbi.nlm.nih.gov/nuccore/NC_018080
+	 * Except file NC_018080.gb from https://www.ncbi.nlm.nih.gov/nuccore/NC_018080
 	 */
 	@Test
 	public void testGithub843() throws Exception {
--- a/biojava-core/src/test/java/org/biojava/nbio/core/sequence/io/GenbankReaderTest.java
+++ b/biojava-core/src/test/java/org/biojava/nbio/core/sequence/io/GenbankReaderTest.java
@@ -353,11 +353,10 @@ public class GenbankReaderTest {
 		DNASequence dna = new ArrayList<>(dnaSequences.values()).get(0);
 		assertNotNull(dna);
 
-		for (Iterator<FeatureInterface<AbstractSequence<NucleotideCompound>, NucleotideCompound>> it = dna.getFeaturesByType("tRNA").iterator(); it.hasNext();) {
-			FeatureInterface<AbstractSequence<NucleotideCompound>, NucleotideCompound> tRNAFeature = it.next();
-			String anticodon = tRNAFeature.getQualifiers().get("anticodon").get(0).getValue();
-			assertFalse(anticodon.contains(" "));
-		}
+		FeatureInterface<AbstractSequence<NucleotideCompound>, NucleotideCompound> tRNAFeature = dna.getFeaturesByType("tRNA").get(0);
+		String anticodon = tRNAFeature.getQualifiers().get("anticodon").get(0).getValue();
+		assertEquals("(pos:complement(1123552..1123554),aa:Leu,seq:caa)", anticodon);
+
 	}
 
 	/**
--- a/biojava-core/src/test/java/org/biojava/nbio/core/sequence/io/GenbankReaderTest.java
+++ b/biojava-core/src/test/java/org/biojava/nbio/core/sequence/io/GenbankReaderTest.java
@@ -36,10 +36,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.*;
-import java.util.ArrayList;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.*;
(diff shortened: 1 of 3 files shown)


josemduarte (changes requested):
Thanks @MaxGreil ! this is great progress.

Adding the full 14MB genbank file is a bit too much for a unit test. We should avoid adding a lot of data if not needed. The ideal unit test is a test that goes through a few lines only and asserts that the behaviour is as expected. For instance you can get the few lines excerpt from the genbank record to demonstrate the issue. You can even keep the few lines as a string constant inside the test.

MaxGreil:
Hi @josemduarte , I reduced the size of file NC_018080.gb to contain only one relevant case with qualifier anticodon.

heuermh:
Is the fix for `transl_except` as simple?  Though here aren't any examples of these kind of failing lines in the linked issue

aalhossary:
ور

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

MaxGreil:
I added a fix for qualifier `transl_except` in the genebank parser according to the documentation.

@heuermh thank you for posting the documentation for qualifier `transl_except`. Do you know an example file for qualifier `transl_except` with line wrap? Or is testGithub843 sufficient as it is?

@josemduarte does it make sense to split this PR for issue #843 into 2 separate PRs, i.e. one PR for the genebank parser and one PR for the embl parser?

biojava-core/src/test/java/org/biojava/nbio/core/sequence/io/GenbankReaderTest.java
```diff
@@ -333,6 +330,36 @@ public void readSequenceWithZeroSpanFeature() throws IOException, CompoundNotFou
...
+		LinkedHashMap<String, DNASequence> dnaSequences = genbankDNA.process();
+		assertNotNull(dnaSequences);
+
+		DNASequence dna = new ArrayList<>(dnaSequences.values()).get(0);
+		assertNotNull(dna);
+
+		for (Iterator<FeatureInterface<AbstractSequence<NucleotideCompound>, NucleotideCompound>> it = dna.getFeaturesByType("tRNA").iterator(); it.hasNext();) {
+			FeatureInterface<AbstractSequence<NucleotideCompound>, NucleotideCompound> tRNAFeature = it.next();
+			String anticodon = tRNAFeature.getQualifiers().get("anticodon").get(0).getValue();
+			assertFalse(anticodon.contains(" "));
```
josemduarte, line 359: How about also asserting that the string is exactly as expected? If I understand it right, it should be: `(pos:complement(1123552..1123554),aa:Leu,seq:caa)`

josemduarte:
> Do you know an example file for qualifier transl_except with line wrap? Or is testGithub843 sufficient as it is?

I would simply fabricate an example for transl_except with wrapping in the same gb file excerpt text you have added.

>  does it make sense to split this PR for issue #843 into 2 separate PRs, i.e. one PR for the genebank parser and one PR for the embl parser?

Do we already have an EMBL file parser somewhere in BioJava? Sorry I don't know well this part of BioJava. If we do and it's only about fixing it, then please go ahead and make it part of this PR. It will only need an additional unit test (again, it's ok to have fabricated data).

MaxGreil:
I added a line in file `NC_018080.gb` qualifier `transl_except` and tested for correct parsing.

However, I have an additional question:

I had a look at the EMBL file parser with test `givenAnEmilFileWhenProcessEmilFileThanTheSequenceShouldReturnAsExpected` in `biojava-core/src/test/java/org/biojava/nbio/core/sequence/io/embl/EmblReaderTest.java`. 
To my understanding, the FeatureTable in file `embl.test`  is not parsed at all. 
Please have a look at `biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/embl/EmblRecord.java`. Variable `featureTable` is just a String that gets assign and overwritten by the last value.

@josemduarte , should I open a new issue for this problem?

josemduarte:
> I added a line in file NC_018080.gb qualifier transl_except and tested for correct parsing.

Great, thanks @MaxGreil ! I think this looks good now. @heuermh could you double check it too?

> I had a look at the EMBL file parser with test givenAnEmilFileWhenProcessEmilFileThanTheSequenceShouldReturnAsExpected in biojava-core/src/test/java/org/biojava/nbio/core/sequence/io/embl/EmblReaderTest.java.
To my understanding, the FeatureTable in file embl.test is not parsed at all.
Please have a look at biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/embl/EmblRecord.java. Variable featureTable is just a String that gets assign and overwritten by the last value.

Thanks for investigating. Indeed that deserves another issue separate from #843

josemduarte:
I'll merge this by the end of the week if there's no other comments

PR #921, opened 2021-03-04
Bugfix: unknown residues were not looked up properly

josemduarte:
`ChemCompGroupFactory.getOneLetterCode(Chemcomp)` would try to get a chemcomp for an empty identifier ".cif", because parent chemcomp from the new `ChemComp` object is an empty string rather than null. 

This fixes the problem. But I'd like @JonStargaryen to have a look at this and see if we need to make `ChemComp.monNstdParentCompId` default to null instead.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/chem/ChemCompGroupFactory.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/chem/ChemCompGroupFactory.java
@@ -111,7 +111,7 @@ public class ChemCompGroupFactory {
         String oneLetter = cc.getOneLetterCode();
         if (oneLetter == null || oneLetter.equals("X") || oneLetter.equals("?")) {
             String parentId = cc.getMonNstdParentCompId();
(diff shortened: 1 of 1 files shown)


JonStargaryen (commented):
Thanks for spotting this!

> see if we need to make ChemComp.monNstdParentCompId default to null instead.

I think this is the way to go. It's one less breaking change when we keep `null` here. Actually, there are other currently broken things, see for example:
https://github.com/biojava/biojava/blob/53c2dc0c3807161707bfecd4302b906c6ad5d84e/biojava-structure/src/main/java/org/biojava/nbio/structure/chem/ChemComp.java#L74

Something along the lines of:
```
public void setMonNstdParentCompId(String monNstdParentCompId) {
    this.monNstdParentCompId = monNstdParentCompId.isEmpty() ? null : monNstdParentCompId;
    setStandardFlag();
}
```
would fix this.

josemduarte:
Ok thanks. Then I think it's best that I close this one and you submit that change as a new PR.

PR #922, opened 2021-03-05
ChemComp#monNstdParentCompId now returns null instead of empty String

JonStargaryen:
ChemComps without parent would erroneously return an empty String. This restores the old behavior and will return `null`.
This fixes #921.


josemduarte (approved):
LGTM, thanks!

josemduarte:
Actually, how about if the argument `monNstdParentCompId` is null?

JonStargaryen:
> Actually, how about if the argument `monNstdParentCompId` is null?

Anything parsed from CIF is non-null. I've added an explicit check for the case that this function is called from elsewhere.

PR #926, opened 2021-04-13
fix logging in CifStructureConsumerImpl if type symbol is unknown

JonStargaryen:
- log statement mistakenly referenced `typeSymbol` (which is now the CifColumn object)
- led to cryptic log messages like: `org.biojava.nbio.structure.io.cif.CifStructureConsumerImpl - Element org.rcsb.cif.schema.DelegatingStrColumn@13353fd2 was not recognised as a BioJava-known element, the element will be represented as the generic element R`
- fixed to log like previously: `org.biojava.nbio.structure.io.cif.CifStructureConsumerImpl - Element X was not recognised as a BioJava-known element, the element will be represented as the generic element R`


josemduarte (approved):
LGTM thanks!

PR #927, opened 2021-04-14
Cif reader file parsing parameters fix

josemduarte:
The file parsing parameters weren't respected by the new Cif reader.


JonStargaryen (approved):
LGTM, thanks for spotting this!

PR #936, opened 2021-06-16
Fix #933 URLConnection to invalid HTTP URL is not detected as failure

### LINKED ISSUE #933: URLConnection to invalid HTTP URL is not detected as a failure
TestDownloadChemCompProvider.testIfWeCachedGarbageWeCanDetectIt() still fails to detect when the target URL is invalid or the file has no contents. Instead of deleting the local cache file, a 20 bytes file (effectively an empty one) is saved.
Related to #703

aalhossary:
When the httpConnection is invalid, the application does not detect it, and it saves an empty (precisely a 20 bytes) file.
This fix detects when the connection is invalid or the file has no contents.
Related to #703
Adding a flag to be successful only when at least one byte is successfully read from the server

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/chem/DownloadChemCompProvider.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/chem/DownloadChemCompProvider.java
@@ -325,7 +325,7 @@ public class DownloadChemCompProvider implements ChemCompProvider {
                     success = true;
                 }
                 if(!success) {
-                	throw new IOException("Could not read from URL "+url.toString());
+                	throw new IOException("Malformed URL or no content found in "+url.toString());
                 }
 
                 pw.flush();
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/chem/DownloadChemCompProvider.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/chem/DownloadChemCompProvider.java
@@ -14,7 +14,6 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.FilenameFilter;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


biojava-structure/src/main/java/org/biojava/nbio/structure/chem/DownloadChemCompProvider.java
```diff
@@ -320,8 +319,13 @@ private static boolean downloadChemCompRecord(String recordName) {
             try (PrintWriter pw = new PrintWriter(new GZIPOutputStream(new FileOutputStream(newFile)));
                  BufferedReader fileBuffer = new BufferedReader(new InputStreamReader(uconn.getInputStream()))) {
                 String line;
+                boolean success = false;
                 while ((line = fileBuffer.readLine()) != null) {
                     pw.println(line);
+                    success = true;
+                }
+                if(!success) {
+                	throw new IOException("Could not read from URL "+url.toString());
```
josemduarte, line 328: Could you be more specific in the error message: `"Found no content in URL" + ...`

josemduarte (changes requested):
Thanks! One comment about the error message

biojava-structure/src/main/java/org/biojava/nbio/structure/chem/DownloadChemCompProvider.java
aalhossary: OK. I will mention the two possible reasons

josemduarte (approved):
Thanks!

josemduarte:
@aalhossary , we are having some trouble with the CI tests. Could you try pushing an empty commit (`git commit --allow-empty -m "Trigger Build"
`) to see if that triggers the tests?

PR #937, opened 2021-06-16
Fix #934 Several file and path issues on windows

### LINKED ISSUE #934: Several file and path issues on windows: space, colon ':', and parameters string
Fix several file and path issues on windows

MS Windows has file and path problems:
1- spaces are not handled well in URL. Converting it to a URI handles decoding the path well.
2- TestURLIdentifier fails because when the connection is to a file and it contains parameters (e.g. ?format=PDB), they are perceived as part of the file name. getting the path first fixes it. 
3- TestMmtfStructureReader and TestMmtfStructureWriter fail because Path does not handle drive specifier (e.g. C:) well, while File does.

aalhossary:
MS Windows has file and path problems:
1- spaces are not handled well in URL. Converting it to a URI handles decoding the path well.
2- TestURLIdentifier fails because when the connection is to a file and it contains parameters (e.g. ?format=PDB), they are perceived as part of the file name. getting the path first fixes it.
3- TestMmtfStructureReader and TestMmtfStructureWriter fail because Path does not handle drive specifier with a colon ':' (e.g. C:) well, while File does.
fixes #934

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/util/InputStreamProvider.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/util/InputStreamProvider.java
@@ -23,6 +23,7 @@
 package org.biojava.nbio.core.util;
 
 import java.io.*;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.Enumeration;
 import java.util.jar.JarEntry;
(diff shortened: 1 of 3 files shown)


josemduarte:
@aalhossary , we are having some trouble with the CI tests. Could you try pushing an empty commit (`git commit --allow-empty -m "Trigger Build"
`) to see if that triggers the tests?

aalhossary:
Well, it is obvious that this triggered the tests but did not report the statuses.

PR #939, opened 2021-07-02
ABITrace test coverage

richarda23:
Added test for ABITrace#getTrace() method, which  lacked test coverage.

Thanks, great project.

### DIFF
--- a/biojava-core/src/test/java/org/biojava/nbio/core/sequence/io/ABITracerTest.java
+++ b/biojava-core/src/test/java/org/biojava/nbio/core/sequence/io/ABITracerTest.java
@@ -21,7 +21,10 @@
 
 package org.biojava.nbio.core.sequence.io;
 
+import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 
 import java.awt.image.BufferedImage;
 import java.io.File;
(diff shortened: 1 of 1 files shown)


josemduarte (approved):
Thank you! LGTM

Note that the CI test running from github workflows is not working at the moment. However travis does and all is well there.

richarda23:
no problem - if I do more of these sorts of test improvements, so you prefer more frequent small PRs or less frequent larger ones? Do you have any plans to start using junit5?

josemduarte:
I think PRs should be about a single topic (e.g. a new feature or a bugfix). They can be large if the feature is large. In any case, I think it'd be acceptable to have several unit tests in a certain biojava area packed together.

Regarding junit5 I don't have strong opinions. I think it'd be nice to move forward, but I don't know how much effort would be required to convert the existing tests. I remember that it wasn't difficult for junit3->4 because intellij had an automatic convertor. Is there something similar available for 4->5?

richarda23:
Yes, junit5 has a compatibility layer ('vintage'), so new tests can be written in junit5 and older junit 3/4 tests can be migrated later (or not). We just updated at work, and it's nice to start using test display names, parameterized tests, nested tests etc. Java 8  is required and version 2.22 of maven surefire test runner plugin. 
   
  
  
   
  
  
   If you like I can submit a PR with the pom updates and a few examples, and you can see what you think?
   
  
  
   Richard
   
  
  
   
    On 02/07/2021 22:01 Jose Manuel Duarte ***@***.***> wrote:
   
   
    
   
   
    
   
   
   I think PRs should be about a single topic (e.g. a new feature or a bugfix). They can be large if the feature is large. In any case, I think it'd be acceptable to have several unit tests in a certain biojava area packed together.
   Regarding junit5 I don't have strong opinions. I think it'd be nice to move forward, but I don't know how much effort would be required to convert the existing tests. I remember that it wasn't difficult for junit3->4 because intellij had an automatic convertor. Is there something similar available for 4->5?

josemduarte:
> If you like I can submit a PR with the pom updates and a few examples, and you can see what you think?

That sounds good to me, thanks!

PR #940, opened 2021-07-04
Junit5 upgrade

richarda23:
This PR demonstrates using JUnit 5(Jupiter) in BioJava

Work done:

- added dependency declarations to <dependencyManagement> section of top-level pom
- added junit5 dependency to biojava-core
- refactored 2 unit tests - ABITracerTest and TestAmbiguityCompoundSet to  use junit5 and demo a few of the junit5 features such as parameterised tests and  displayName.
Screenshot 2021-07-04 at 10 57 49

- ran entire test suite, all passed on my dev machine
- setup follows guidelines here: https://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4
- So far I've just tested in biojava-core. 
- Like the junit3/4 migration, idea is to enable possibility to write new tests using JU5 and update existing tests over time at own pace.


biojava-core/src/test/java/org/biojava/nbio/core/util/SequenceToolsTest.java
```diff
@@ -0,0 +1,74 @@
...
+		String[] nucs = new String[] { "A", "T", "C", "G" };
+		Random r = new Random();
+		StringBuilder sb = new StringBuilder();
+		for (int i = 0; i < n; i++) {
+			sb.append(nucs[r.nextInt(4)]);
+		}
+		return sb.toString();
+	}
+
+	@Nested
```
richarda23, line 28: I thought using nested tests  for these utility classes of  diverse static methods would help group related tests together better.

biojava-core/pom.xml
```diff
@@ -42,8 +42,19 @@
 		<dependency>
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
-			<scope>test</scope>
 		</dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
```
josemduarte, line 48: I guess there's no need to set scope=test here, because it's done in parent pom, right?

biojava-core/src/main/java/org/biojava/nbio/core/util/SequenceTools.java
```diff
@@ -104,13 +119,31 @@ public static boolean isNucleotideSequence(String sequence)
...
+	 * peptide sequence and a DNA sequence, will always be returned 
+	 * as a DNA sequence.
+	 * </p>
+	 * <p>
+	 * An empty string argument returns a ProteinSequence of length 0.
+	 * A null argument throws a {@link NullPointerException}
+	 * @param sequence
+	 * @return Either a DNASequence or a ProteinSequence
+	 * @throws CompoundNotFoundException
+	 */
```
josemduarte, line 138: Nice! thanks for the docs!

josemduarte (approved):
LGTM, thank you! Just one question, see below

biojava-core/pom.xml
richarda23: yes, that's right, in the <dependencyManagement> section you just need to set the version and scope once, then sub-modules just need to add  the dependency

biojava-core/src/main/java/org/biojava/nbio/core/util/SequenceTools.java
richarda23: Yes.. although whether that behaviour is what is desired though might be another matter... I guess this is a non-standard way to create a new Sequence since the Cookbook/Docs uses more rigorous methods to accomplish that.

josemduarte:
Thanks again. I'll merge. BTW I guess you had to do the test conversion manually, right? Or are there some tools to help in that? I remember IntelliJ had a neat 3->4 convertor.

richarda23:
> Thanks again. I'll merge. BTW I guess you had to do the test conversion manually, right? Or are there some tools to help in that? I remember IntelliJ had a neat 3->4 convertor.

Yes, I did these manually to use some  of junit5 features. But there may well be an automated way to just changing import names and some annotations, e.g. @Before -> @BeforeEach

PR #941, opened 2021-07-09
update junit3->junit4 test, remove doc references to ju3

richarda23:
Thought it might be useful to update the few remaining junit3 tests to use junit4; so that in the event the junit5 PR is accepted, there won't be 3 concurrent versions of junit tests, just 2.


josemduarte (approved):
Thanks!

Issue #944, opened 2021-07-23
Improve test coverage - currently 49.5% in bio-java core


richarda23:
Test coverage is little low - covering more code would enable
 - more confident refactoring. In a code-base with contributions over a long period of time there are cases where new language features can replace existing code in perhaps a more reliable way, but without tests it's sometimes hard to be sure 
 - greater confidence in the existing functionality
 - self-documentation of methods
 - identifying obsolete or unused code which could perhaps be scheduled for removal or marked deprecated?

Like covid vaccination rates 100% is unachievable but 50% is a bit on the low side, especially for a library project. 
Making sure all public methods are invoked in test code would be a start, and it could be incremental.
I'll attach a PR adding some tests for some utility methods which identified 1 bug to get the ball rolling...

josemduarte:
Thanks @richarda23 . Totally agree, this has been an eternal pending issue for this project. More test coverage is badly needed.

One curiosity. How did you measure coverage? 

I actually setup SonarCloud for biojava sometime ago (which does coverage too), but I'm not sure if that's accessible to everyone.

richarda23:
For this, I just manually  ran EclEmma which is an Eclipse plugin (I'm not sure if it's actually maintained any longer), but  I expect most tools give similar information.

richarda23:
Latest values are 75% class, 56% method, 58% line - so some improvement. Will try to add more tests to at least touch uncovered non-trivial classes

josemduarte:
That's great news. Thanks @richarda23 . FYI I did set up Sonarqube cloud in CI. But I think test coverage reporting is somehow not working. 

Anyway if you are interested the Sonarqube reports are available here: https://sonarcloud.io/organizations/biojava/projects . Let me know if you can't access that and I'll see how to add accounts.

richarda23:
Thanks Jose. I'm tackling some classes now in biojava-core o.b.n.core.sequence package (ChromosomeSequence, ExonSequence etc) as some of these aren't covered by tests. 
I've come across quite a few issues with some of the classes that I would consider bugs, for example:
 - equals and hashcode throwing  NullPointer exceptions as some of the properties used by equals() in AbstractSequence aren't set 
 -  calculation of  intron sequences on negative strands returns unexpected results in GeneSequence
 - collections of subelements aren't encapsulated
 What would you advise? I can raise a PR with the tests I've written, with those failing ones marked @Disabled for discussion, or raise what I think are  issues.

josemduarte:
[...]

PR #945, opened 2021-07-23
Tests for StringManipulationHelper

richarda23:
- identified and fixed a bug in equalsToIgnoreEndline if number of lines was different
- identified equalsToXml as not really mplemented or documented and not used
- re-implemented join() in simpler way using tests to validate unchanged behaviour and widened to accept a Collection rather than AbstractCollection
- some updated javadoc
- work towards #944

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/util/StringManipulationHelper.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/util/StringManipulationHelper.java
@@ -29,7 +29,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.util.Collection;
-import java.util.Iterator;
 import java.util.Scanner;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


josemduarte (approved):
LGTM, thank you! Thanks for the javadocs too

biojava-core/src/main/java/org/biojava/nbio/core/util/StringManipulationHelper.java
```diff
@@ -174,14 +175,36 @@ public static boolean equalsToXml(String expected, String actual) {
...
+	 * @return The right-padded string. 
+	 * @throws IllegalArgumentException if n <= 0
+	 */
 	public static String padRight(String s, int n) {
+		validatePadding(n);
 	     return String.format("%1$-" + n + "s", s);
 	}
 
+	private static void validatePadding(int n) {
+		if (n <=0 ) {
```
aalhossary, line 227: Why can't the padding be == 0?

richarda23: If I set padding to 0, a java.util.MissingFormatWidthException is thrown from within String.format method when padRight is called, and a FormatFlagsConversionMismatchException when padLeft is called.

aalhossary: Thank you for clarification, @richarda23. I doublechecked the Javadoc and it is inhibited from their side.
Please add your name in the @author tag.

PR #948, opened 2021-07-26
Fix #946 Parse PDB KEWODS record

### LINKED ISSUE #946: parse PDB record KEYWDS
The KEYWDS record contains a set of terms relevant to the entry. Terms in the KEYWDS record provide a simple means of categorizing entries and may be used to generate index files. This record addresses some of the limitations found in the classification field of the HEADER record. It provides the opportunity to add further annotation to the entry in a concise and computer-searchable fashion.

This record is present in both **PDB** as `Keywords_PDB` and **mmCIF** as `_struct_keywords.text` and in **PDBx** as `_struct_keywords.pdbx_keywords`.

aalhossary:
They are parsed from PDB and added to Structure.
Not parsed from mmCIF yet.
Fixes #946


biojava-structure/src/test/java/org/biojava/nbio/structure/TestKeywords.java
```diff
@@ -0,0 +1,41 @@
...
+		Structure structure = pdbpars.parsePDBFile(inStream);
+		List<String> keywords = structure.getKeywords();
+		assertEquals(keywords.size(), 12);
+		assertEquals(keywords.get(11), "TRANSCRIPTION REGULATOR");
+	}
+
+	@Test
+	public void testDash() throws IOException {
+		String fileName;
+		fileName = "/pdb6elw.ent.gz";
```
josemduarte, line 30: It would be ideal to do this with a unit test that includes just a few lines of PDB-format text. Instead of adding a whole new PDB file to the resources. Would that be doable?

josemduarte (changes requested):
Thanks @aalhossary . Since mmCIF now is the main format, I think it'd be great if you could include filling the new keywords field from mmCIF and BCIF. That should be actually very straight forward. The CIF parsing work is done already, so it is only a matter of calling the `setKeywords()` with the value extracted from the appropriate mmCIF field. Could you have a look at that?

biojava-structure/src/test/java/org/biojava/nbio/structure/TestKeywords.java
aalhossary: Done

aalhossary:
Added setting keywords from `_struct_keywords.text` and fixed a wrong use of `_struct_keywords.pdbx_keywords`.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/CifStructureConsumerImpl.java
```diff
@@ -885,11 +896,26 @@ public void consumeStructConnType(StructConnType structConnType) {
...
+			for (String string : strings) {
+				keywordsList.add(string.trim());
+			}
+		}
+    	structure.setKeywords(keywordsList);
+
         StrColumn pdbxKeywords = structKeywords.getPdbxKeywords();
-        // TODO what is the correct format for these?
         if (pdbxKeywords.isDefined()) {
+        	//TODO I think no need for the Collectors.joining(", ") part.
```
JonStargaryen, line 915: Yes, you are correct. This can be simplified to:
```
        StrColumn pdbxKeywords = structKeywords.getPdbxKeywords();
        if (pdbxKeywords.isDefined()) {
            String keywords = pdbxKeywords.get(0);
            pdbHeader.setClassification(keywords);
        }
```

JonStargaryen (commented):
Thanks for the improvements @aalhossary, below are some comments on the CIF parsing part.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/CifStructureConsumerImpl.java
```diff
@@ -885,11 +896,26 @@ public void consumeStructConnType(StructConnType structConnType) {
...
             String keywords = pdbxKeywords.values().collect(Collectors.joining(", "));
-            pdbHeader.setDescription(keywords);
+            //TODO What is the right place to fill this field from?
+//        	pdbHeader.setDescription(keywords);
```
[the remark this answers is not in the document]
aalhossary: Thank you! Done.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/AbstractCifFileSupplier.java
```diff
@@ -45,6 +45,10 @@ protected CifFile getInternal(Structure structure, List<WrappedAtom> wrappedAtom
         MmCifBlockBuilder blockBuilder = CifBuilder.enterFile(StandardSchemata.MMCIF)
                 .enterBlock(structure.getPDBCode());
 
+        blockBuilder.enterStructKeywords().enterText()
```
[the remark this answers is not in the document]
aalhossary: Thank you! Don.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/CifStructureConsumerImpl.java
```diff
@@ -885,11 +896,26 @@ public void consumeStructConnType(StructConnType structConnType) {
 
     @Override
     public void consumeStructKeywords(StructKeywords structKeywords) {
+    	//There may be a simpler implementation or a better implementation using
```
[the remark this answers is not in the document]
aalhossary: Thank you! Done.

aalhossary:
Thank you @josemduarte and @JonStargaryen.
Shall we create another new issue to remove the description field altogether from the PDB Header?

josemduarte (changes requested):
Thanks @aalhossary . Actually, now that I read the discussion about the `PDBHeader.getDescription()` field. It looks like `keywords` would be better placed under `PDBHeader`. Also it seems that description was used to hold the keywords data formerly, right? How about deprecating it, keeping the data in both fields (keywords and description)?

aalhossary:
The **Header** as defined in the [PDB file format](https://www.wwpdb.org/documentation/file-format-content/format33/sect2.html#HEADER) has three specific fields (not including keywords):
- classification
- depDate
- idCode

**Keywords** may include:
- Functional classification.
- Metabolic role.
- Known biological or chemical activity.
- Structural classification.
- Other classifying terms

It would be misleading for a user to find the [KEYWODS field](https://www.wwpdb.org/documentation/file-format-content/format33/sect2.html#KEYWDS) in the header where it is not defined. Therefore I believe it's better to keep it outside the header.

===============
Yes, it seems that the `description` was used to store the keywords (from the mmCIF files only) formerly. 

Therefore, I agree with you that the best option is to deprecate description, because:
- keywords field does not _only_ include description
- A description by convention is a free text not a list of keywords.

josemduarte:
@aalhossary the `PDBHeader` class in BioJava does not intend to replicate the `HEADER` annotation in the legacy PDB format. Instead it is a way to pack into a single object annotations and extra data that are only needed in some use-cases, but not in others. For instance, if someone cares about coordinates only for a certain application, they could completely drop the `PDBHeader` by setting it to null. My point of storing keywords under `PDBHeader` is to avoid having many extra fields at the `Structure` level, that might not be used in many applications.

aalhossary:
Although BioJava travis-ci does not run currently, the latest commit of this PR was successfully run on my travis https://app.travis-ci.com/github/aalhossary/biojava (except for the deployment part of course).

biojava-structure/src/main/java/org/biojava/nbio/structure/io/PDBFileParser.java
```diff
@@ -2757,6 +2765,52 @@ private void makeCompounds(List<String> compoundList,
 
 	}
 
+	/**Parse KEYWODS record of the PDF file.<br>
```
josemduarte, line 2768: Typo: PDB file

biojava-structure/src/main/java/org/biojava/nbio/structure/PDBHeader.java
```diff
@@ -589,9 +599,16 @@ public String getTitle() {
...
+	
+	/**@deprecated will be removed later. Use {@link #getKeywords()}
+	 * @return
+	 */
 	public String getDescription() {
 		return description;
 	}
+	/**@deprecated will be removed later. Use {@link #setKeywords(List)}
+	 * @param description
+	 */
```
josemduarte, line 611: Please use the `@Deprecated` code tag as well, see https://stackoverflow.com/questions/9031832/how-to-declare-or-mark-a-java-method-as-deprecated

biojava-structure/src/main/java/org/biojava/nbio/structure/PDBHeader.java
```diff
@@ -589,9 +599,16 @@ public String getTitle() {
 	public void setTitle(String title) {
 		this.title = title;
 	}
+	
+	/**@deprecated will be removed later. Use {@link #getKeywords()}
+	 * @return
+	 */
```
josemduarte, line 605: Please use the `@Deprecated` code tag as well, see https://stackoverflow.com/questions/9031832/how-to-declare-or-mark-a-java-method-as-deprecated

josemduarte (changes requested):
Thanks again @aalhossary and sorry for the delay. See a minor issue about the deprecated tag. After that it should be ok to merge

biojava-structure/src/main/java/org/biojava/nbio/structure/io/PDBFileParser.java
aalhossary: Done.

biojava-structure/src/main/java/org/biojava/nbio/structure/PDBHeader.java
aalhossary: I didn't know that the code tag is part of the language while the javadoc is not. thanks for the info.
Done

biojava-structure/src/main/java/org/biojava/nbio/structure/PDBHeader.java
aalhossary: I didn't know that the code tag is part of the language while the javadoc is not. thanks for the info.
Done

josemduarte (approved):
Thanks!

josemduarte:

aalhossary:
I have 2 questions that need to clarify:
1) Is the first character of the current (short) PDB ID format a digit in the range [0-9] or [1-9]?
2) will the "PDB_" prefix in the future (extended) format be case sensitive or not?

PR #928, opened 2021-04-19
Update guava dependency version to 30.1.1-jre

heuermh:

PR #949, opened 2021-07-26
Update pom.xml

rsoreq:


rsoreq:
Adding Renovate

.whitesource
```diff
@@ -0,0 +1,19 @@
+{
+  "scanSettings": {
```
heuermh, line 2: What is this file for?

biojava-genome/pom.xml
```diff
@@ -75,7 +75,7 @@
 			<groupId>com.google.guava</groupId>
 			<artifactId>guava</artifactId>
 			<scope>compile</scope>
-			<version>28.1-jre</version>
+			<version>30.1.1-jre</version>
```
heuermh, line 78: 👍

biojava-protein-comparison-tool/pom.xml
```diff
@@ -58,7 +58,7 @@
 		<dependency>
 			<groupId>net.sourceforge.jmol</groupId>
 			<artifactId>jmol</artifactId>
-			<version>13.0.14</version>
+			<version>14.31.10</version>
```
heuermh, line 61: This is a major version upgrade, and a large jump at that.  Are there any build or runtime issues caused by this?

pom.xml
```diff
@@ -1,699 +1,20 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
```
heuermh, line 2: Why is this here?  I don't understand this change.

heuermh (changes requested):

biojava-genome/pom.xml
heuermh: Wait, I am confused, it already is 30.1.1-jre on `master`.  See https://github.com/biojava/biojava/pull/928

biojava-protein-comparison-tool/pom.xml
sbliven: @heuermh Unfortunately we don't have any tests for jmol functionality. I'm not even sure how that would work, other than manually running the alignment tools and/or ce-symm. However I think it's still a good idea to keep jmol recent. (I build the jmol mvn packages for at least the final release of each major version).

sbliven (changes requested):
Thanks for the PR. It's good to keep our dependencies updated. However you seem to have deleted most of pom.xml, including attribution, project info, plugins, etc. Can you say what you were trying to achieve with this change? I don't think we can accept it as is.

josemduarte:
There were no replies to the comments for a long time. I'll close

PR #950, opened 2021-07-27
Support extended pdbid

### LINKED ISSUE #930: Support extended PDB IDs
The wwPDB will run out of 4-character PDB IDs soon and will switch to an extended format: https://www.wwpdb.org/news/news?year=2021#607760112786e73a79c76f9d

The new format will be eight characters prefixed by ‘PDB’, e.g., pdb_00001abc.

Chemical compounds will also run out of 3-letter identifiers soon and will switch to 4-letter identifiers. These are incompatible with the PDB file format.

We should ensure that biojava parsers, data structures, and algorithms can handle both of these changes.

aalhossary:
I added `PDBId` class to replace the current String PdbId with conversation methods to and from short format.
I will write the documentation later, as soon as the API is approved and finalized.
Fixes #930.


biojava-structure/src/main/java/org/biojava/nbio/structure/Structure.java
```diff
@@ -788,4 +765,53 @@
...
+//	 *  <code>getStructureIdentifier().toCanonical().getPdbId()</code>
+//	 */
+//	@Deprecated
+//	String getPdbId();
+//
+//	@Deprecated
+//	void setPdbId(String pdb_id);
+
+
+	PDBId getPDBId();
```
sbliven, line 814: Needs docstring

biojava-structure/src/main/java/org/biojava/nbio/structure/Structure.java
```diff
@@ -788,4 +765,53 @@
...
+	void setPDBId(PDBId pdbId);
```
sbliven, line 815: Needs Docstring

biojava-structure/src/main/java/org/biojava/nbio/structure/Structure.java
sbliven: Docstring should mention that `getPDBId().getId()` replaces the old (differently capitallized) version.

sbliven (commented):
@aalhossary Awesome work! FYI I've started reviewing it but it's a pretty PR so it might take me a few days.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmtf/MmtfStructureReader.java
```diff
@@ -166,7 +168,11 @@ public void finalizeStructure() {
 	@Override
 	public void initStructure(int totalNumBonds, int totalNumAtoms, int totalNumGroups,
 			int totalNumChains, int totalNumModels, String modelId) {
```
sbliven, line 170: Here's somewhere where it might be cleaner to change the interface to `initStructure(..., PDBId)` since the (poorly named) modelId is intended to be a PDBId. The change would propogate to places that supply the modelId, like `StructureDataInterface`.

biojava-structure/src/test/java/org/biojava/nbio/structure/TestAtomCache.java
```diff
@@ -227,7 +228,7 @@ public void testGetScopDomain() throws IOException, StructureException {
 
 		Structure s = cache.getStructure(name);
 		assertNotNull("Failed to fetch structure from SCOP ID", s);
-		assertEquals("2gs2.A", s.getName());
+		assertEquals("2GS2.A", s.getName()); //TODO make sure that updating the case won't cause any harm
```
sbliven, line 231: Fetching lowercase structures still works, right? Maybe we should leave some tests lowercase to check this.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/CifStructureConsumerImpl.java
```diff
@@ -858,8 +860,14 @@ public void consumeStruct(Struct struct) {
         }
 
         if (struct.isDefined() && struct.getEntryId().isDefined()) {
-            pdbHeader.setIdCode(struct.getEntryId().get(0));
-            structure.setPDBCode(struct.getEntryId().get(0));
+            PDBId pdbId = null;
+			try {
+				pdbId = new PDBId(struct.getEntryId().get(0));
+			} catch (PDBIdException e) {
+				throw new IllegalArgumentException(e);
```
sbliven, line 867: Is IllegalArgumentException the best RuntimeException to wrap here?

biojava-structure/src/main/java/org/biojava/nbio/structure/SubstructureIdentifier.java
```diff
@@ -75,24 +82,27 @@
...
-		if(idRange[0].length() != 4) {
-			this.pdbId = idRange[0];
+		if(idRange[0].length() == 4) {  //Short form
+			this.pdbId = new PDBId(idRange[0]/* .toUpperCase() */);
+		} else if(idRange[0].length() == 12 && idRange[0].substring(0, 4).equalsIgnoreCase("PDB_")) { //extended form
+			this.pdbId = new PDBId(idRange[0]/* .toUpperCase() */);
+		} else {
+			this.pdbId = new PDBId(idRange[0]/* .toUpperCase() */);
 			// Changed from Exception to a warning to support files and stuff -sbliven 2015/01/22
 			logger.warn(String.format("Unrecognized PDB code %s",this.pdbId));
```
sbliven, line 105: I'm trying to remember why I changed it from an exception to a warning. I wonder if the StructureName tests break if this throws an exception for invalid PDBIds. Is this change why you had to change many of the tests to expect exceptions?

Either way, this if statement is unneeded. If we declare it to throw a PDBIdException then you can just do

```
this.pdbId = new PDBId(idRange[0]); // no if block, no warn statement.
```

Or  to preserve the old behavior, remove the `throws PDBIdExeption` from the declaration and do
```
try {
    this.pdbId = new PDBId(idRange[0]);
} catch(PDBIdException e) {
    logger.warn(...)
}

Comment (sbliven) [reply biojava-structure/src/main/java/org/biojava/nbio/structure/SubstructureIdentifier.java:105]: My personal preference would be to support structures without PDBId where possible (probably why I went with the warning back in 2015)

File: biojava-structure/src/main/java/org/biojava/nbio/structure/SubstructureIdentifier.java [hunk 7]
```diff
@@ -115,7 +125,11 @@ public SubstructureIdentifier(String pdbId, List<ResidueRange> ranges) {
 		if(ranges == null) {
```
sbliven, line 125: Should be overridden with a PDBId version

biojava-structure/src/main/java/org/biojava/nbio/structure/SubstructureIdentifier.java
```diff
@@ -115,7 +125,11 @@ public SubstructureIdentifier(String pdbId, List<ResidueRange> ranges) {
...
 			throw new NullPointerException("Null ranges list");
 		}
-		this.pdbId = pdbId;
+		try {
+			this.pdbId = new PDBId(pdbId);
+		} catch (PDBIdException e) {
+			throw new IllegalArgumentException(e);
```
sbliven, line 131: Again, should either ignore the malformed Id (perhaps set `this.pdbId = null`) or declare `throws PDBIdException`.

biojava-structure/src/main/java/org/biojava/nbio/structure/SubstructureIdentifier.java
```diff
@@ -135,14 +149,18 @@ public String toString() {
 	 */
 	@Override
 	public String getIdentifier() {
-		if (ranges.isEmpty()) return pdbId;
+		if (ranges.isEmpty()) return pdbId.getId();
 		return pdbId + "." + ResidueRange.toString(ranges);
 	}
 
 	public String getPdbId() {
```
sbliven, line 156: Should we deprecate this? Two methods that differ only by capitalization is confusing.

biojava-structure-gui/src/main/java/org/biojava/nbio/structure/gui/util/PDBUploadPanel.java
```diff
@@ -150,7 +151,8 @@ private Structure getStructure(JTextField filePath,JTextField chainId) throws St
 			throw new StructureException(e);
 		}
 
-		Structure reduced = StructureTools.getReducedStructure(s, chainId.getText());
+//		Structure reduced = StructureTools.getReducedStructure(s, chainId.getText());
+		Structure reduced = new SubstructureIdentifier(s.getPDBId().getId()+"."+ chainId.getText()).reduce(s);  //TODO double check this
```
sbliven, line 155: To avoid string parsing you can use
```
new SubstructureIdentifier(s.getPDBID(), Lists.asList(new ResidueRange(chainId.getText(), null, null))).reduce()
```

sbliven: But the concept of replacing `getReducedStructure` with `StructureIdentifier.reduce()` is correct

sbliven (changes requested):
Excellent, thanks @aalhossary! Extended ids are now [available in mmcif](https://www.wwpdb.org/news/news?year=2021#60fee1500882a5597783cc0a) so this is good timing.

I was looking for detailed information about this change but I wasn't able find much beyond the two wwpdb blogs. Do you know of any more detailed info about the change? Are there example files?

If I understand correctly, the major changes are:
- New `PDBId` class which can represent IDs as either 1ABC or PDB_00001ABC form
- Parse keywords from pdb headers

Breaking changes:
- Some deprecated and removed methods in Structure
   - notably, `String getPdbId()` becomes `PDBId getPDBId()` (note capitalization) or `String getPDBCode()`
- PDB codes are returned upper-case consistently


## PDBId class

I think that wrapping ids into a light-weight class like this is a good approach. It's better than storing them as strings and providing conversion methods, despite the small overhead of object creation.

One thing we should discuss is all the static flags controlling short/long behavior. It's nice to have this flexibility, but since it's global state it is difficult to test and can lead to unexpected bugs if the flags get changed in a different part of the code. Instead of global state, maybe any code that needs to override the defaultShorteningBehavior should be forced to call `getId(false)`?

We could also improve the memory usage by storing either the 4-byte string (if isShortCompatible) or the 8-byte extended ID. Then the `PDB_` prefix and leading zeros would only be added in `toExtendedId`. Maybe this is premature optimization, but I think loading large lists of PDBIds is a reasonable use case.

## Consistent usage

It would be good to be consistent about whether to accept String or PDBId.

I would suggest that classes that are used commonly by users should override methods to accept both Strings and PDBIds. The more "internal classes" (I'm thinking of the parser internals like `StructureDataInterface`) can just have PDBId versions. Or if we want to minimize breaking, we can define both but deprecate the String versions for a release.


## Keywords

Is the keyword parsing related to the extended ids in any way? If not it might be good to extract that to a separate PR.

Keywords should also be set from mmcif (and mmtf if they are supported).

## Parsing

This PR sets up all the infrastructure for storing extended PDBIds. However, I suspect that additional work will be  needed to ensure that the parsers are using the new IDs. For instance, I  don't see anywhere that is reading the ` _database_2.pdbx_database_accession` field in the cif parser. (This is  more a TODO than a suggestion.)

## Testing

I didn't see any tests that actually use 5+ character structures. Can we add a cif file that uses a long id and test parsing and such. I suspect we will find additional bugs (e.g. in the alignment output formats). We don't need to fix everything in this initial PR of course.


Thanks so much for building this feature. This was  way more work than I expected when I made this issue, so thanks for taking it on!

aalhossary:
Hi Spencer @sbliven,

I don't have any resources other than these 2 blogs too :). There are no details nor examples :(

While writing my edit, I focused on making the minimal changes possible to the current codebase. Therefore, I minimized throwing any new exceptions from already present code that does not do so.

I will reply your individual questions in place.

## KEYWODS records
It is actually a part of another PR. Both have some files in common. I thought by the time this PR gets reviewed, the other one would have been approved already.


## PDBId class
### Default shortening Behavior
The static flag for default shortening behavior does NOT have a setter. The static flag is mainly for easier change in the future in a later release. On the other hand, the object method getId(boolean) is public in case anybody wants it.

### Short Id to minimize memory usage
I don't think storing 4-bute short ID is a good idea for these reasons:

1. It will add a lot of overhead to compare / hashcode Ids for a small memory benefit
2. the New IDs would eventually supersede the short ID, especially as the new depositions are being added in an increasing rate, while the old ones are being obsoleted by new ones.

However, I will consider storing the `8-byte extended ID` only in a later optimization PR, after the new IDs start to emerge and we see how the format would look like.

## Consistent usage
I am planning to move away from the String gradually as possible, leaving the very frequently used classes by users have overloaded functions with deprecated String variant, but I will NOT remove them soon.

## Parsers using the new PDBId
Whatever the parser is doing, it will eventually one of the SetPdbId() [already removed altogether], setPDBId(), or setPdbCode() methods.
However, I would appreciate it if you point out any missing / additional places to take care of. I am sorry I am still nw to parsing mmCIF files.

## Testing
Before submitting my PR, I had either modified my code a lot to pass the currently available unit tests, or to modify the tests themselves, if they test for non realistic scenarios.

## The only format EXCEPTION
XXXX is the only exception of an accepted ID that doesn't follow the **`\d\p{Alnum}{3}`** format.
I added it because I found it declared in the header of **4nwr-assembly1.cif.gz** referenced from `TestURLBasedFileParsing`.
Can somebody confirm whither this file is correct or wrong and should be remediated? @josemduarte Can you help us find the answer of this question?

biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/cath/CathDomainTest.java
```diff
@@ -36,6 +36,6 @@
 	public void test() {
 		String id = "1qvrC03";
 		CathDomain domain = CathFactory.getCathDatabase().getDomainByCathId(id);
-		assertEquals("1qvr.C_332-400,C_514-540", domain.toCanonical().getIdentifier());
+		assertEquals("1QVR.C_332-400,C_514-540", domain.toCanonical().getIdentifier());
```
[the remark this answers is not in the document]
aalhossary: PDB IDs are not, while Chain IDs are

biojava-structure/src/main/java/org/biojava/nbio/structure/PDBId.java
```diff
@@ -0,0 +1,146 @@
...
+		}
+		public PDBIdException(String message, Throwable cause) {
+			super(message, cause);
+		}
+		public PDBIdException(Throwable cause) {
+			super(cause);
+		}
+	}
+
+	public static final String SHORT_PDBID_FORMRMAT		= "\\d{1}\\p{Alnum}{3}";
```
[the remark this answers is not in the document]
aalhossary: Done

biojava-structure/src/main/java/org/biojava/nbio/structure/Structure.java
aalhossary: I will add the documentation soon. We just need to catch the new release 6.0.0 first.

biojava-structure/src/main/java/org/biojava/nbio/structure/Structure.java
aalhossary: Will do soon.

biojava-structure/src/main/java/org/biojava/nbio/structure/align/client/StructureName.java
```diff
@@ -72,14 +74,15 @@
 	private static final Logger logger = LoggerFactory.getLogger(StructureName.class);
 
 	protected String name;
-	protected String pdbId;
+	protected PDBId pdbId;
 	protected String chainName;
 
-	private static final Pattern cathPattern = Pattern.compile("^(?:CATH:)?([0-9][a-z0-9]{3})(\\w)([0-9]{2})$",Pattern.CASE_INSENSITIVE);
+	//TODO Double check the modified patterns
+	private static final Pattern cathPattern = Pattern.compile("^(?:CATH:)?((?:pdb_[0-9]{4})?[0-9][a-z0-9]{3})(\\w)([0-9]{2})$",Pattern.CASE_INSENSITIVE);
```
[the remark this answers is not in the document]
aalhossary: We don't have any samples or confirmed examples.
I will add a `TODO` to check all Regular Expressions later.

biojava-structure/src/main/java/org/biojava/nbio/structure/align/pairwise/AlternativeAlignment.java
```diff
@@ -856,7 +857,11 @@ public Structure getAlignedStructure(Structure s1, Structure s2){
 		Calc.shift( s3, currentTranMatrix);
 
 		Structure newpdb = new StructureImpl();
-		newpdb.setPDBCode("Java");
+		try {
+			newpdb.setPDBId(new PDBId("XXXX"));
```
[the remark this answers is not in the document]
aalhossary: Currently, I allow ONE exception only: **XXXX**.
Do you think it is useful to have several exceptions, or even not to enforce the format at all?

biojava-structure/src/main/java/org/biojava/nbio/structure/align/util/AtomCache.java
```diff
@@ -772,7 +776,7 @@ protected void flagLoadingFinished(String name) {
 	public Structure getStructureForPdbId(String pdbId) throws IOException, StructureException {
 		if (pdbId == null)
 			return null;
-		if (pdbId.length() != 4) {
+		if (! (pdbId.length() == 4 || (pdbId.length() == 12  && pdbId.substring(0, 4).equalsIgnoreCase("pdb_")))) {
```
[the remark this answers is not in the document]
aalhossary: But `PDBIdException` already extends `StructureException`!
Anyway, it would be easier to just create an Id and throw the exception

biojava-structure/src/main/java/org/biojava/nbio/structure/io/LocalPDBDirectory.java
```diff
@@ -351,7 +354,7 @@ public Structure getStructureById(String pdbId) throws IOException {
 	 */
 	protected InputStream getInputStream(String pdbId) throws IOException{
 
-		if ( pdbId.length() != 4)
+		if (! ((pdbId.length() == 4) || (pdbId.length() == 12 && pdbId.substring(0, 4).equalsIgnoreCase("pdb_"))))
```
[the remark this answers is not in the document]
aalhossary: same as above

biojava-structure/src/main/java/org/biojava/nbio/structure/align/client/StructureName.java
```diff
@@ -341,9 +365,9 @@ private boolean initFromFile() {
 	 * @return The upper-case PDB Name, or null if not applicable
 	 * @throws StructureException Wraps errors which occur when converting to canonical form
 	 */
-	public String getPdbId() throws StructureException {
+	public PDBId getPdbId() throws StructureException {
```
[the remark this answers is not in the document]
aalhossary: OK

biojava-structure/src/main/java/org/biojava/nbio/structure/io/LocalPDBDirectory.java
```diff
@@ -351,7 +354,7 @@ public Structure getStructureById(String pdbId) throws IOException {
 	 */
 	protected InputStream getInputStream(String pdbId) throws IOException{
```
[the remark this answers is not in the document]
aalhossary: Yes. No harm to have a PDBId function, provided that we keep the String one, because it it used by the users frequently.

biojava-structure/src/main/java/org/biojava/nbio/structure/StructureIO.java
```diff
@@ -235,17 +239,17 @@ public static Structure getBiologicalAssembly(String pdbId, int biolAssemblyNr)
...
-	 * Utility method to set the location where PDB files can be found
-	 *
-	 * @param pathToPDBFiles
-	 */
-	public static void setPdbPath(String pathToPDBFiles) {
-		if (!pathToPDBFiles.endsWith(FILE_SEPARATOR))
-			pathToPDBFiles += FILE_SEPARATOR;
-	}
+//	private static final String FILE_SEPARATOR = System.getProperty("file.separator");
+//  TODO This function seems to have NO EFFECT. And it is not called anywhere.
```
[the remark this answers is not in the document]
aalhossary: OK

biojava-structure/src/main/java/org/biojava/nbio/structure/Structure.java
```diff
@@ -788,4 +765,53 @@
...
+//	/**
+//	 * Returns the PDB identifier associated with this StructureIdentifier.
+//	 * @deprecated From BioJava 4.2, use {@link #getPDBCode()} or
+//	 *  <code>getStructureIdentifier().toCanonical().getPdbId()</code>
+//	 */
+//	@Deprecated
+//	String getPdbId();
+//
+//	@Deprecated
+//	void setPdbId(String pdb_id);
```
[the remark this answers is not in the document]
aalhossary: OK

biojava-structure/src/main/java/org/biojava/nbio/structure/StructureImpl.java
```diff
@@ -1013,11 +1006,73 @@ public String getIdentifier() {
...
+//	 *@deprecated
+//	 */
+//	@Override
+//	public void setPdbId(String pdb_id) {
+//		try {
+//			this.pdbId = new PDBId(pdb_id);
+//		} catch (PDBIdException e) {
+//			throw new IllegalArgumentException(e);
+//		}
+//	}
```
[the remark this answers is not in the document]
aalhossary: OK

biojava-structure/src/main/java/org/biojava/nbio/structure/StructureTools.java
```diff
@@ -1259,66 +1259,67 @@ public static boolean isNucleotide(String groupCode3) {
...
+//					// found matching entity info. set description...
+//					newS.getPDBHeader().setDescription(
+//							"Chain " + c.getId() + " of " + s.getPDBCode()
+//							+ " " + comp.getDescription());
+//				}
+//			}
+//		}
+//
+//		return newS;
+//	}
```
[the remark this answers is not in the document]
aalhossary: OK

biojava-structure/src/test/java/org/biojava/nbio/structure/align/client/TestStructureName.java
```diff
@@ -115,72 +118,82 @@ public void testPrefixes() throws StructureException {
...
+		assertEquals(new PDBId("2EHZ"),sn.getPdbId());
 
 		// Invalid prefix
 		sn = new StructureName("XXX:2ehz");
 		assertTrue(sn.isPdbId());
 		assertTrue(sn.getSource() == PDB);
-		assertEquals("XXX:2ehz",sn.getPdbId());
+//		assertEquals("XXX:2ehz",sn.getPdbId());
+		try {
+			assertEquals(null,sn.getPdbId());
```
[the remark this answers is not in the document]
aalhossary: OK. This is easier :)

biojava-structure/src/test/java/org/biojava/nbio/structure/TestAtomCache.java
aalhossary: Fetching the lowercase structures still works, because it is converted to lower case anyway in the URL.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/CifStructureConsumerImpl.java
aalhossary: I don't know. I didn't want to propagate a checked exception in functions which I didn't explore well.
Any suggestions? @sbliven

biojava-structure/src/main/java/org/biojava/nbio/structure/SubstructureIdentifier.java
aalhossary: OK

biojava-structure-gui/src/main/java/org/biojava/nbio/structure/gui/util/PDBUploadPanel.java
aalhossary: Fine. Thank you

biojava-structure/src/test/java/org/biojava/nbio/structure/align/client/TestStructureName.java
```diff
@@ -115,72 +118,82 @@ public void testPrefixes() throws StructureException {
...
 		assertTrue(sn.getSource() == BIO);
-		assertEquals("2EHZ",sn.getPdbId());
+		assertEquals(new PDBId("2EHZ"),sn.getPdbId());
 
 		// Invalid prefix
 		sn = new StructureName("XXX:2ehz");
 		assertTrue(sn.isPdbId());
 		assertTrue(sn.getSource() == PDB);
-		assertEquals("XXX:2ehz",sn.getPdbId());
+//		assertEquals("XXX:2ehz",sn.getPdbId());
```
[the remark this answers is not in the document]
aalhossary: Because in the last line of StructureName.init(), the whole string `"XXX:2ehz"` is passed to the initFromPDB(String), not the suffix `2ehz` only. (although the function parameter is called `suffix` not `name`.
Of course `XXX:2ehz` does not match the PDBID format.
If we pass the suffix only, it will pass well, the PDBId object would be created, and there will be NO WARNING.
Any suggestions @sbliven ?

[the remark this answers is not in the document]
aalhossary: UPDATE: After thinking over, I got to a conclusion.
If it is EXPLICITLY defined to be an UNKNOWN format, it should NOT be recognized. Hence, it defaults to null.

biojava-structure/src/main/java/org/biojava/nbio/structure/SubstructureIdentifier.java
aalhossary: Done as possible. Please tell me your feedback when I submit the new commit.

biojava-structure/src/main/java/org/biojava/nbio/structure/SubstructureIdentifier.java
aalhossary: I declared it to throw PDBIdException

biojava-structure/src/main/java/org/biojava/nbio/structure/SubstructureIdentifier.java
aalhossary: Yes. Done.

biojava-structure-gui/src/main/java/org/biojava/nbio/structure/gui/util/PDBUploadPanel.java
aalhossary: OK Done

biojava-structure/src/main/java/org/biojava/nbio/structure/io/mmtf/MmtfStructureReader.java
aalhossary: This function implements an abstract function in the interface `StructureAdapterInterface` from `org.rcsb.mmtf.api`.
I can not modify the signature.

aalhossary:
I submitted a new revision (this time I performed a full system test before submission :) ).
I left the questions which need answers unresolved in @sbliven comments.

aalhossary:
@sbliven , @josemduarte This PR has been silent for 25 days. Shall I consider the API stable and start writing the documentation?

josemduarte (commented):
Thanks @aalhossary . I have a few concerns about this one, the most important of them already expressed by @sbliven : this new class to wrap the PDB identifiers should be used consistently throughout all the relevant interfaces. Doing that is quite complicated and begs for the question whether the String accessors should be deprecated or live side by side.

For that and other reasons, I'm going to vote against proceeding with this solution. I feel that having some helper methods that can do PDB id string handling could be a more flexible solution.

biojava-structure/src/main/java/org/biojava/nbio/structure/align/pairwise/AlternativeAlignment.java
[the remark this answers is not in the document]
aalhossary: I have just got your point: making a constant static PDBId object in the PDBId class.
Good idea. I will.

aalhossary:
In fact, I am not convinced with your 2 or 3 points @josemduarte for these reasons:

## Helper Method
A helper method was an option that I excluded in the first place; because what we need is not only conversion from old format to extended format and vise versa. but also
- checks of order (via `java.lang.Comparable`) among different PDBIDs.
- checks of `hashCode` and `equals` among PDBIDs of different formats to use in a `HashMap`.
- checking whether a PDBID is short or long, and hence do whatever necessary conversions.
- checking whether an extended PDBID is shortable.
- general regular expression matching.

Implementing all of these features needs several methods. If they all were implemented as static helper functions somewhere, the result would be 

- Complex code
- We reply on the user's own discretion to use them properly.
- We would either add several methods to the `Strusture` class that are actually unrelated to its main purpose, or add a _new_ separate class . In the later case, I prefer grouping all related methods in a light-weight class, which is what I did.

## The need for complicated restructuring.
I reviewed every place where `setPDBCode()` is called. I found only 4 places remaining, two are to be done in the next commit, and the other two are actually misuse of the field and should be changed.
I also reviewed every comment by Spenser Bliven and found that most if not all of them are met and solved already.

> this new class to wrap the PDB identifiers should be used consistently throughout all the relevant interfaces.

I believe my PR is quite comprehensive. @josemduarte I would appreciate it if you list out all remaining interfaces to restructure. It would then be _my own_ task to handle them.


## Should the String accessors be deprecated or live side by side?

By all means, they should coexist now and for a long-enough time. Maybe we will think about the answer of this question when we release BioJava 7.0.0 or even 10.0.0!
If you ask me about my opinion, they should **coexist (as convenience methods)**.


## Finally
I appreciate mentioning the remaining concerns you have @josemduarte.
What is the opinion of other members?
What do you think @sbliven ?

## My next tasks in this PR

1. Keeping shorter string to represent the PDB ID in memory.
2. Replacing remaining `setPDBCode` and `getPDBCode` with `setPDBId` and `getPDBId`
3. Adding documentation
4. Any other refactoring to be proposed by members.

biojava-structure/src/main/java/org/biojava/nbio/structure/align/pairwise/AlternativeAlignment.java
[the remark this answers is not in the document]
aalhossary: Done! will be added in the next commit.

josemduarte:
Sorry for the delay @aalhossary . I've read your new comments and arguments and I see your point. Let me check the PR more carefully now and submit a new review for it.

josemduarte:
BTW when you have time, could you solve the conflicts by merging master into your branch?

aalhossary:
I committed what I have till now.
I will try merging the master branch now.

aalhossary:
I have 2 questions that need to clarify:

Is the first character of the current (short) PDB ID format a digit in the range [0-9] or [1-9]?
will the "PDB_" prefix in the future (extended) format be case sensitive or not?

aalhossary:
In the regex `pdb_[a-zA-Z0-9]{8}`
I see the `pdb_` prefix in small case while the later 8 characters case insensitive. Is it intentional?

josemduarte:
Good point, I think it'd be ok to assume lower case only, i.e.

`pdb_[a-z0-9]{8}`

PR #951, opened 2021-07-31
CRC64Checksum test #944

richarda23:
Test for basic behaviour of Crc64Checksum - #944

I think there is a bug exposed by test `partialbyteRange()` and I've proposed a solution in the Crc64Checksum class.

### DIFF
--- a/biojava-core/src/test/java/org/biojava/nbio/core/util/CRC64ChecksumTest.java
+++ b/biojava-core/src/test/java/org/biojava/nbio/core/util/CRC64ChecksumTest.java
@@ -98,11 +98,6 @@ class CRC64ChecksumTest {
             ()->assertThrows(IllegalArgumentException.class,
                 ()->crc64.update(testBytes,  testBytes.length, 1))
         );
-        crc64.update(testBytes, 2, 1);
-        String partialBytesHex = crc64.toString();
-        crc64.reset();
-        crc64.update(testBytes[2]);
-        assertEquals(partialBytesHex, crc64.toString());
     }
 
 
--- a/biojava-core/src/test/java/org/biojava/nbio/core/util/CRC64ChecksumTest.java
+++ b/biojava-core/src/test/java/org/biojava/nbio/core/util/CRC64ChecksumTest.java
@@ -1,9 +1,10 @@
 package org.biojava.nbio.core.util;
 
+import static org.junit.jupiter.api.Assertions.assertAll;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.DisplayName;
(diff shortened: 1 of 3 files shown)


biojava-core/src/test/java/org/biojava/nbio/core/util/CRC64ChecksumTest.java
```diff
@@ -0,0 +1,97 @@
...
+        crc64.reset();
+        crc64.update(testBytes);
+        assertEquals(valueFromAllRange, crc64.getValue());
+    }
+
+    @Test 
+    @Disabled
+    // this doesn't work as expected
+    // and doesn't behave according to interface. The 3rd argument
+    // is treated as an index rather than a number of bytes to include
```
sbliven, line 81: It looks to me that the function is intended to take (bytes, start, end) rather than (bytes, start, length). I think this is a failure with the documentation (both non-existent javadoc and the misleadingly named parameters). It looks like it should run as expected with `crc64.update(testBytes, 2, 3);`.

sbliven (changes requested):
Thanks a lot for these tests! Improving our code coverage is great. The CRC64Checksum class seems to have been copied from expasy4j, so it

Regarding the partialbyteRange test, I'm not sure that this captures the intended behavior. Is the function used like this elsewhere? Unless there's a good reason to change it, I would prefer making this test pass now and then fixing any places that are expecting (bytes, start, length) at the caller.

biojava-core/src/test/java/org/biojava/nbio/core/util/CRC64ChecksumTest.java
richarda23: This method doesn't seem to be used anywhere internally in BioJava. The behaviour is specified in the interface    `java.util.zip.Checksum` which this class implements (javadoc reproduced below). Since this is an interface in the standard Java library  we can't change  the contract of the interface. If we assume that there is some external client code using this method,  then from their point of view fixing this  might be a breaking change, but this would be OK if it's documented and is in version 6?

```
 Updates the current checksum with the specified array of bytes.
     *
     * @param b the byte array to update the checksum with
     * @param off the start offset of the data
     * @param len the number of bytes to use for the update
```

josemduarte: I think it'd be ok introducing the breaking change for 6.0.0. For sure we should document it in the CHANGELOG

richarda23: Pushed this fix and updated the changelog.
 Also, added some input validation,  following example of `java.util.zip.CRC32` implementation

josemduarte (approved):
Thanks! LGTM

richarda23:
Hi.
this is still showing up as 'changes requested' but I think they are fixed and approved, should I do anything more?

josemduarte:
I think @sbliven would have to approve. In any case, I think his comments were addressed. Let's give it a couple of days and then merge.

PR #953, opened 2021-07-31
Force US Locale for floats.

sbliven:
#918 showed that locales which use commas for floating-point numbers
can cause problems with output formats. This commit fixes a number
of other cases.

- Force Locale.US when outputting floats to files/streams
- Usually don't force Locale.US when outputting floats to standard out &
  logging (would lead to more consistent logs but breaks localization)
- Replace some `println(String.format(` calls with Java 5 `printf`

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/util/SingleLinkageClusterer.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/util/SingleLinkageClusterer.java
@@ -69,7 +69,7 @@ public class SingleLinkageClusterer {
 			if (closestDistance==Double.MAX_VALUE) {
 				closestDistStr = String.format("%6s", "inf");
 			} else {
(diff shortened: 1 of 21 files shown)


josemduarte (approved):
LGTM thanks!

richarda23:
Just wondering how it will be remembered in future to check when new string.format invocations are added - is there a style-guide checklist or similar to aid PR reviewers to remember to check whether Locale.US is set correctly or not?
I was thinking it would be useful to have a utility method which would handle whether or not to use Locale.US. but then people have to remember to call it, so it's not much of an improvement....

josemduarte:
Good point @richarda23 . The problem is this is a case by case situation. Some times it makes sense to enforce US locale, but sometimes it may not make sense, because a user who is writing an application might want to have outputs that adhere to a certain locale.

PR #957, opened 2021-08-17
Core utils tests part2 - #944

richarda23:
not for the first time, thank goodness for `git reflog`. Here is the PR for erroneously deleted XMLHelper tests again


josemduarte (approved):
Thanks fore rescuing it :)

Issue #958, opened 2021-08-23
Travis-CI not running for negative credit balance


aalhossary:
Travis-CI does not run BioJava for some reason related to negative balance!!!
Here is the error message
**Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting.**

aalhossary:
Update: It seems that `remoteURL.openStream(); ` is working back now.
I will update the issue text and title

josemduarte:
Thanks for catching it @aalhossary . It does look like we have run out of credits on the free-plan at travis-ci.com. I don't see how or when we will get more credits. After some googling, it seems this is a common problem. 

Is the only solution dropping travis-ci and using github workflows only? Any experience on this anyone?

aalhossary:
The only experience I have is rerunning the same tests from my (personal) travis-ci account, considering that my BioJava is a clone of the BioJava/BioJava.

There is no harm in giving GitHub workflow actions a try.

sbliven:
Related to #952.

josemduarte:
I'll close this as duplicate of #952 . We'll continue the discussion there.

PR #959, opened 2021-08-25
Alignment matrix tests

richarda23:
- tests for substitution matrices
- javadoc and some reformatting
- also removed some tests that required java 11 and broke Java 8 tests, from a previous PR
-


biojava-core/src/test/java/org/biojava/nbio/core/util/CRC64ChecksumTest.java
```diff
@@ -65,16 +65,6 @@ void allbyteRange (){
...
-    @Test
-    void allRangeIsSameAsAllArray (){
-        byte [] testBytes = new byte [] {1,2,3,4,5};
-        crc64.update(testBytes, 0, testBytes.length);
-        Long valueFromAllRange = crc64.getValue();
-        crc64.reset();
-        crc64.update(testBytes);
-        assertEquals(valueFromAllRange, crc64.getValue());
-    }
-
```
josemduarte, line 77: Could you explain why this one is removed?

josemduarte (approved):
LGTM, thanks! I only have one question below.

In any case before merging we'll need to get the CI problems fixed. Travis is not running currently, see #958

biojava-core/src/test/java/org/biojava/nbio/core/util/CRC64ChecksumTest.java
richarda23: because it's a method added in Java9 so won't compile in Java 8. I recently moved from Eclipse to IntelliJ and assumed it was running tests according to Maven setup, sorry I should have just run tests from maven command line before originally committing this.

josemduarte:

PR #960, opened 2021-08-28
Multiple OBO  altids fix for  #964

richarda23:
This PR enables multiple alt_ids by storing the annotation value as a List rather than a single item.
There seems to be no restriction in Annotation as to what can be values.

But calling code would need to know somehow that getting an Annotation keyed by alt_id would return a list. 
A new test added and minor reformatting/updating of code


josemduarte (approved):
LGTM thanks! Again, let's wait for #958 to be resolved before merging.

josemduarte:

Issue #964, opened 2021-09-12
ParseGO demo application doesn't work


richarda23:
Initially this doesn't work as the URL
`http://sourceforge.net/p/song/svn/HEAD/tree/trunk/subsets/biosapiens.obo?format=raw` is redirected to https. So  changing the URL to https will fix this. 
Using a logger to show output is also broken - no output is actually printed to the console. For a simple main application, would it be more robust to use System.out.println statements to ensure some output is shown?
Will raise a PR to fix the URL.
Looking at the SF project it's not been touched for 6 years - is there a newer active repository for this?

josemduarte:
I do agree that demo applications are better off with simple System.out/err.print statements. Logging is not guaranteed to be set up properly in an end-user environment.

PR #965, opened 2021-09-12
Fix parse go demo #964

richarda23:
fix for #964


lafita (approved):
Looks good, thanks! If the file is not too large, another option is to download it to the resources folder and parse it locally. We have been trying to avoid remote dependencies in BioJava, specially in tests and demos, although many still remain.

richarda23:
it's 72kb, is that small enough ? Agree would be good to be local if possible

lafita:
Yes, I think that counts as small! Thanks @richarda23

richarda23:
- replaced url call with file
- logging doesn't output any results so replaced with system.out/err calls
- exit with error code if fails

biojava-ontology/src/main/java/demo/ParseGO.java
```diff
@@ -27,42 +27,41 @@
...
+	 * @param args
+	 */
+	public static void main(String[] args) {
 
 		try {
-			URL url = new URL(u);
 
 			OboParser parser = new OboParser();
-			InputStream inStream = url.openStream();
+			InputStream inStream = OboParser.class.getResourceAsStream("/ontology/biosapiens.obo");
```
lafita, line 51: Have you committed the file? I can't see it in the commits

josemduarte (approved):
LGTM thanks!

Issue #966, opened 2021-09-12
Main methods in classes are untested and some no longer work


richarda23:
Many classes have main methods. E.g. here is a listing of those in core project:
```
grep -r 'public static void main' . 
./src/main/java/org/biojava/nbio/core/util/UncompressInputStream.java:	public static void main(String[] args) throws Exception {
./src/main/java/org/biojava/nbio/core/util/FileDownloadUtils.java:	public static void main(String[] args) {
./src/main/java/org/biojava/nbio/core/sequence/location/InsdcParser.java:	public static void main(String[] args){
./src/main/java/org/biojava/nbio/core/sequence/loader/GenbankProxySequenceReader.java:	public static void main(String[] args) throws Throwable {
./src/main/java/org/biojava/nbio/core/sequence/loader/UniprotProxySequenceReader.java:	public static void main(String[] args) {
./src/main/java/org/biojava/nbio/core/sequence/DNASequence.java:	public static void main(String[] args) throws Exception {
./src/main/java/org/biojava/nbio/core/sequence/io/FastaReaderHelper.java:	public static void main(String[] args) throws Exception {
./src/main/java/org/biojava/nbio/core/sequence/io/GenbankReaderHelper.java:	public static void main(String[] args) throws Exception {
./src/main/java/org/biojava/nbio/core/sequence/io/GenericFastaHeaderParser.java:	public static void main(String[] args) {
./src/main/java/org/biojava/nbio/core/sequence/io/FastaGeneWriter.java:	public static void main(String[] args) {
./src/main/java/org/biojava/nbio/core/sequence/io/GenbankWriter.java:	 * public static void main(String[] args) { try { FileInputStream is = new
./src/main/java/org/biojava/nbio/core/sequence/io/FastaWriter.java:	public static void main(String[] args) {
./src/main/java/org/biojava/nbio/core/sequence/io/FastaReader.java:	public static void main(String[] args) {
./src/main/java/org/biojava/nbio/core/sequence/io/CasePreservingProteinSequenceCreator.java:	public static void main(String[] args) throws Exception {
./src/main/java/org/biojava/nbio/core/sequence/ProteinSequence.java:	public static void main(String[] args) throws Exception {
./src/main/java/demo/ParseFastaFileDemo.java:	public static void main(String[] args) throws Exception {
./src/main/java/demo/DemoSixFrameTranslation.java:	public static void main(String[] args){
```
[...]

PR #967, opened 2021-09-12
Run mains #966

richarda23:
Suggestion for automated way to test main methods for breakage. 
Each main method would have a line in 'invocations.txt' with a FQDN of the class with the main method and any arguments. 
These get executed in turn and failures reported

We could add additional columns for expected output if we wanted more sophisticated testing other than avoidance of complete failure. 
Currently there is a broken main method (FastaWriter) that squashes an IO exception to a non-existent file and does not return a non-zero exit code on failure - so as well as this PR we'd need to make the main methods more 'well-behaved' and not squash failures.


josemduarte:
Thanks @richarda23 . As I commented on #966 , I think it'd be best to try to get rid of main methods completely. The ones that are one-off developer tests should be removed (both `ProteinSequence` and `FastaWriter` look like that) and the ones that are good tests should be moved to the tests package (perhaps `UncompressInputStream` qualifies as one of those). 

That would avoid the need of having custom solutions to test them, like the one proposed here.

Issue #943, opened 2021-07-21
NPE when parsing certain PDB structures with "params.setCreateAtomBonds(true)"


cleberecht:
Hi,
thank you a lot providing and updating BioJava continuously.

I encountered a problem with the legacy pdb file parsing in the latest `6.0.0-alpha4` version of BioJava. When creating creating bonds an NPE is thrown.

Minimal (not)working example:
``` java
PDBFileReader reader = new PDBFileReader();
FileParsingParameters params = new FileParsingParameters();
params.setCreateAtomBonds(true);
reader.setFileParsingParameters(params);
reader.getStructureById("1v9i");
```
throws
```
java.lang.NullPointerException: Cannot invoke "org.biojava.nbio.structure.Chain.getGroupByPDB(org.biojava.nbio.structure.ResidueNumber)" because "chain" is null

	at org.biojava.nbio.structure.io.BondMaker.getAtomFromRecord(BondMaker.java:475)
	at org.biojava.nbio.structure.io.BondMaker.formLinkRecordBond(BondMaker.java:336)
	at org.biojava.nbio.structure.io.PDBFileParser.formBonds(PDBFileParser.java:2775)
	at org.biojava.nbio.structure.io.PDBFileParser.parsePDBFile(PDBFileParser.java:2700)
	at org.biojava.nbio.structure.io.PDBFileParser.parsePDBFile(PDBFileParser.java:2557)
	at org.biojava.nbio.structure.io.PDBFileReader.getStructure(PDBFileReader.java:140)
	at org.biojava.nbio.structure.io.LocalPDBDirectory.getStructureById(LocalPDBDirectory.java:334)
	at ...
```
I think there is a problem with the legacy chain id format where `authId:C` is requested but only `asymId:A` and `asymId:B` are present.

All the Best
Christoph

josemduarte:
Thanks for reporting @cleberecht . I'll see if I can work on a fix. As a general advice, PDB format is legacy and with time support for it will be degrading to the most basic parsing only. If you are dealing with deposited PDB files, please use mmCIF or BCIF.

Issue #929, opened 2021-04-27
BondMaker - Could not find atoms specified in SSBOND record


zedomel:
Hi, 

I'm getting the follow warning when reading a PDB file:
<pre>
[main] WARN  org.biojava.nbio.structure.io.BondMaker - Could not find atoms specified in SSBOND record: [SSBOND:
Atom 1:
	Chain: I
	Residue #: 24
	Ins. Code: 
Atom 2:
	Chain: I
	Residue #: 56
	Ins. Code: 
]
</pre>

Looking at the code it seems that `PDBFileParser::triggerEndFileChecks` is being called before `PDBFileParser::formBonds`. The method `triggerEndFileChecks` calls `assignAsymIds` which changes the ID's of the chains and, then the method `BondMaker::formDisulfideBonds` can't find the chains of the SSBonds, since it has changed by `assignAsymIds`.

I will try to do a pull request to fix it, but before I need to know what is the best approach to fix it.

thanks.

josemduarte:
Thanks for the report @zedomel . First of all one important point: I think this feature works well when reading from mmCIF files. If you are dealing with PDB-deposited files, then I would strongly recommend to use mmCIF. It has been the main format for many years now, whilst the PDB format is considered legacy. Also most software packages for structure solution will output in mmCIF format.

In any case, regarding the bug you report, I think the best strategy would be to write a small unit test that demonstrates the issue. Then it should be pretty straight forward to fix. The problem sounds like a confusion with asym_ids and author chain ids. In PDB files there's only 1 identifier per chain (the author chain id).

erikedlund:
I'm running into a variant of this bug loading structure 1A4W from a .pdb with `FileParsingParams.setCreateAtomBonds(true)`

It results in a NullPointerException preventing parsing. Looks like the SSBondImpl gets an author chain ID, while the BondMaker class is looking for an asym_id in `getAtomFromRecord()`


> java.lang.NullPointerException
> 	at org.biojava.nbio.structure.io.BondMaker.getAtomFromRecord(BondMaker.java:475)
> 	at org.biojava.nbio.structure.io.BondMaker.formDisulfideBond(BondMaker.java:298)
> 	at org.biojava.nbio.structure.io.BondMaker.formDisulfideBonds(BondMaker.java:292)
> 	at org.biojava.nbio.structure.io.PDBFileParser.formBonds(PDBFileParser.java:2778)
[...]

PR #968, opened 2021-09-22
Fix for #929

josemduarte:
Fix and integration test for #929 . Related to recent fix to #943


JonStargaryen (commented):
There are still warnings about link records that can't be resolved, e.g.:
`LINK        NA    NA H 541                 O   HOH H 497     1555   1555  2.21 `

I think the issue is that these reference 'water chains' and the current checks only cover polymer and non-polymer chains.

No warnings when I add something like this:
```java
Chain waterChain1 = structure.getWaterChainByPDB(linkRecord.getChainID1());
Chain waterChain2 = structure.getWaterChainByPDB(linkRecord.getChainID2());

[...]

if (waterChain1!=null && linkRecord.getResName1().equals("HOH")) allChainIds1.add(waterChain1.getId());
if (waterChain2!=null && linkRecord.getResName2().equals("HOH")) allChainIds2.add(waterChain2.getId());
```

josemduarte:
Thanks for checking and for the fix suggestion! Indeed, with that fix, the water bonds are made and warnings are gone. I've pushed the commit.

JonStargaryen (approved):
Thanks, LGTM now!

Issue #969, opened 2021-09-25
SCOP files < 2.08 not available for download


aalhossary:
the web folder [https://scop.berkeley.edu/downloads/parse/](https://scop.berkeley.edu/downloads/parse/) contains files of version 2.08 only.

All previous files are no longer available.

jmchandonia:
Sorry, that was an error on my part in releasing SCOPe 2.08.  All the downloadable files for previous releases are now available again.  We're still working on generating ASTRAL sequences and subsets for 2.08, so those will be available soon.

josemduarte:
Thanks for the info @jmchandonia  

In any case, changing the default to 2.0.8 ( #972) is still a good idea, thanks @aalhossary .

PR #972, opened 2021-09-25
Fix 969 Update SCOP version to 2.08

aalhossary:
All SCOP installation file downloads fail because all files before 2.08
no longer exist (error 404).

1- All downloads are corrected by upgrading the LATEST_VERSION to 2.08.
2- testGetStructureForChainlessDomains is ignored because chainless
domains are present only in SCOP < 1.73


aalhossary:
Fix for #969

lafita (approved):
Looks good, thanks!

josemduarte (approved):
Thanks for catching and fixing! LGTM

josemduarte:
I'll merge this. I've tested locally and it passes all tests. 

I'll also remove an obsolete scop test that was using some old http endpoint that is now gone. I'll push that directly to master.

aalhossary:
I restored one previously ignored test.

PR #974, opened 2021-09-30
Update ciftools dependency and fix build/tests

JonStargaryen:
- ciftools 3.0.0 (brings AlphaFold/model support to the table)
- ~propose to ignore `openchart` dependency as it doesn't seem to be used - somebody should confirm this, I've never used any of the forester stuff~
- ~with that exclusion, a dependency on jackson-databind was needed~
- ~`Acetylation_site_dataset.gz` reports 403 in tests and on console, but can be downloaded manually via the browser, added this as 1 MB test resource to circumvent test issues~

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,7 +45,7 @@
 		<log4j.version>2.14.0</log4j.version>
 		<junit-jupiter.version>5.7.2</junit-jupiter.version>
 		<ciftools.artifact>ciftools-java-jdk8</ciftools.artifact>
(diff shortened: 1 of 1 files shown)


JonStargaryen:
Removed other changes, as the `openchart` problem is solved and `Acetylation_site_dataset.gz` requires more work (#975). Now only updates the ciftools version.

josemduarte (approved):
Thanks, LGTM

Issue #993, opened 2021-11-07
Support of D-AminoAcids


aalhossary:
D-AminoAcids (https://proteopedia.org/wiki/index.php/Amino_Acids) are optical isomers or enantiomers (mirror images) of naturally occuring L-AminoAcids. They have the same structure but with opposite chirality.
When I parse structures that contain D-AminoAcids, they are parser as hetatoms.

Here are some of my notes:

- I tried adding the D-AminoAcids to StructureTools lookup table, but that was not enough.
- `HetatomImpl.isAminoAcid()` returns `false` for cases like **DAL** (D-Alanine), because it tests whether `PolymerType.PROTEIN_ONLY.contains(pt)`.
- Although PolymerType includes types for both `peptide("polypeptide(L)")` and `dpeptide("polypeptide(D)")`, `PolymerType.PROTEIN_ONLY` contains the `peptide` type only.
- Also I see a lot of redundancy between `StructureTools` and `ChemCompTools`. Is there a need to keep both of them?, or we can simply deprecate the former for the sake of the later.

I think a restructuring task can be done in conjunction with supporting longer ChecmComp names expected in 2022.
We can schedule refactoring this feature to the new version of BioJava. However, we need to start by supporting D-AminoAcids soon.

aalhossary:
Here is another question for community discussion:
When an amino acid (regardless to being L or D) is the substrate (say in an oligopeptide), is it considered both polymeric and ligand or only one of them?

josemduarte:
I have some reservations about making d-aminoacids behave just as normal aminoacids in biojava. When they appear as part of polymers in the PDB, they appear mostly as short peptides, not as full proteins. Also there are a few cases in the PDB where there are engineered proteins composed of only D-aminoacids (e.g. in racemic mixture crystals) but in any case very rare and non-natural.

I can see how in some applications it might make sense to treat them as aminoacids, but I'd say that in most applications you would want them treated as something different from standard aminoacids. For instance, adding them to `PolymerType.PROTEIN_ONLY` seems dangerous to me for the general case.

aalhossary:
I agree to most of your points @josemduarte.
D-AminoAcids are not as frequent as L-AminoAcids (at least yet). However, there are aminoacids at the end of the day.
They have the natural right to possess getN(), getC(), getO() as all other free AminoAcids citizens (sorry for the silly joke :) ).
[...]

PR #994, opened 2021-11-07
(incremental) Support for D-AminoAcids

aalhossary:
Support of D-AminoAcids (to be recognized as AminoAcids instead of HetAtomGroups).

Other modoficatiosn may follow to support bigger ChemComp names.

Related to #993


JonStargaryen:
Brief question: Does this fix a particular problem or is the motivation solely completeness?

At least according to the docs it seems as treating D-amino acids as hetatms is the desired behavior:
https://github.com/biojava/biojava/blob/ebb6eb0151594aaab9b4f1576920cc1605e8dc4e/biojava-structure/src/main/java/org/biojava/nbio/structure/GroupType.java#L45

aalhossary:
> Does this fix a particular problem or is the motivation solely completeness?

It fixes a particular problem. I am searching the PDB for different types of cross links (e.g. isopeptide) which are present in D as well as L aminoacids. Treating D-AminoAcids as just HetAtom group limits me to a great extent.

=======

Regarding completeness, technically they are not different from L-AminoAcids and different bonds can form between them and L-AminoAcids.
Add to this that nucleotide- analogues and D-AminoAcids are increasingly being deposited in PDB nowadays, in contrast to when BioJava 1.7 was designed.

=======

I am willing to take the responsibility of this initiative, but we need to discuss any major design change thoroughly first.

josemduarte (commented):
See my comments in #993

aalhossary:
Let's keep this pull request for the implementation discussions, and keep the design discussions to the #993 .

PR #996, opened 2021-11-11
Change JAXB implementation for better java version compatibility

sbliven:
There was an issue with running CeSymm on java 15+ (rcsb/symmetry#104). I suspect that this issue effected all projects that include biojava with the maven-shade-plugin. I was able to trace the problem back to the old com.sun.xml.bind implementations. Switching these to the glassfish JAXB implementation fixed the problem.

- Downgrade jgrapht to maintain java 8 support
- Switch JAXB to glassfish implementation for better java 15+ support

### DIFF
--- a/biojava-aa-prop/pom.xml
+++ b/biojava-aa-prop/pom.xml
@@ -20,15 +20,15 @@
 	<build>
 		<plugins>
 			<!-- Excluding demo package is required for avoiding namespace clashes (demo package is in all modules) for signing the jar. See issue #387 -->
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-jar-plugin</artifactId>
-                <configuration>
-                    <excludes>
-                        <exclude>demo/**</exclude>
-                    </excludes>
-                </configuration>
-            </plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-jar-plugin</artifactId>
+				<configuration>
+					<excludes>
+						<exclude>demo/**</exclude>
+					</excludes>
+				</configuration>
+			</plugin>
 
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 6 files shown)


biojava-structure/pom.xml
```diff
@@ -62,48 +62,41 @@
 
 		<dependency>
 			<groupId>org.jgrapht</groupId>
-  			<artifactId>jgrapht-core</artifactId>
-  			<version>1.5.1</version>
+			<artifactId>jgrapht-core</artifactId>
+			<!-- Pin! Java 8 requires 1.3.0 -->
+			<version>1.3.0</version>
```
josemduarte, line 67: How about 1.4.0? It seems that would be the last java-8 compatible. That's what their changelog says: https://github.com/jgrapht/jgrapht/blob/master/HISTORY.md. I've double checked by downloading the jar and looking at a class file.

biojava-aa-prop/pom.xml
```diff
@@ -80,28 +80,37 @@
...
 		</dependency>
+
+		<dependency>
+			<groupId>javax.xml.bind</groupId>
+			<artifactId>jaxb-api</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.glassfish.jaxb</groupId>
+			<artifactId>jaxb-runtime</artifactId>
+		</dependency>
```
josemduarte, line 114: It looks like biojava-aa-prop did not have an xml dependency before. I think this is not needed here

biojava-core/pom.xml
```diff
@@ -43,35 +43,43 @@
...
+			<artifactId>log4j-core</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>javax.xml.bind</groupId>
+			<artifactId>jaxb-api</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.glassfish.jaxb</groupId>
+			<artifactId>jaxb-runtime</artifactId>
+		</dependency>
```
josemduarte, line 83: It looks like biojava-core did not have an xml dependency before. I think this is not needed here

biojava-modfinder/pom.xml
```diff
@@ -1,80 +1,88 @@
...
+			<artifactId>log4j-core</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>javax.xml.bind</groupId>
+			<artifactId>jaxb-api</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.glassfish.jaxb</groupId>
+			<artifactId>jaxb-runtime</artifactId>
+		</dependency>
```
josemduarte, line 63: It looks like biojava-modfinder did not have an xml dependency before. I think this is not needed here

josemduarte (changes requested):
Thanks! good catch about jgrapht dependency. I think we'll need a new bugfix release soon to keep the java-8 compatibility.

See comments below

biojava-structure/pom.xml
sbliven: Cood catch. I was looking at `Build-Jdk: 11.0.6` in the manifest, when of course I should have been checking `Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"`. I will update to 1.4.0

biojava-aa-prop/pom.xml
sbliven: I got test failures without it ('No JAXP implementation found'). I think it was serendipitously included before since we were using the com.sun namespace, but that the glassfish implementation needs to be specifically included. I have included it on all modules that import javax.xml.bind.

josemduarte (approved):
LGTM thanks!

PR #997, opened 2021-11-16
Remove mains #966

richarda23:
converting main methods to junit tests or removing altogether from biojava-core
each commit covers one removsl


biojava-core/src/main/java/org/biojava/nbio/core/sequence/GeneSequence.java
```diff
@@ -259,6 +259,7 @@ public ExonSequence removeExon(String accession) {
 	 * @return exon sequence
 	 */
 	public ExonSequence addExon(AccessionID accession, int begin, int end) throws Exception {
+		// TODO should be IllegalArgumentException, is more specific
```
heuermh, line 262: I agree, go ahead and make this change

biojava-core/src/main/java/demo/UncompressFile.java
```diff
@@ -0,0 +1,42 @@
+package demo;
```
heuermh, line 1: This new file (and others in this pull request) need the license header.  For example
https://github.com/biojava/biojava/blob/master/biojava-aa-prop/src/main/java/demo/BioJavaAADemo.java#L1

heuermh (changes requested):
Thanks!  Love to see a pull request that deletes 2x more lines than it adds.

heuermh (approved):

richarda23:
changing the method signature of public ExonSequence addExon(AccessionID accession, int begin, int end) will be a breaking API change as it no longer declares a checked exception

heuermh:
Ah well, no reason to break compatibility here -- we could keep the method signature as `throws Exception` and just change the thrown exception to `IllegalArgumentException`, which extends `Exception` via `RuntimeException`.

richarda23:
hm, it's a bit non-standard to declare runtime exceptions in method signatures, they are normally documented in javadoc: https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html. Just thinking about it some more, calling code will already be handling the checked exception , so won't need to change.  I think only subclasses overriding this method may have to change their signature, which will be a much smaller set of cases.

biojava-core/src/main/java/org/biojava/nbio/core/sequence/GeneSequence.java
```diff
@@ -257,10 +257,11 @@ public ExonSequence removeExon(String accession) {
 	 * @param begin
 	 * @param end
 	 * @return exon sequence
+	 * @throws IllegalArgumentException if accessionID is already added.
 	 */
-	public ExonSequence addExon(AccessionID accession, int begin, int end) throws Exception {
+	public ExonSequence addExon(AccessionID accession, int begin, int end) {
 		if (exonSequenceHashMap.containsKey(accession.getID())) {
-			throw new Exception("Duplicate accesion id " + accession.getID());
+			throw new IllegalArgumentException("Duplicate accesion id " + accession.getID());
```
josemduarte, line 264: Thanks for this change. This is a lot cleaner now. I think generally `throws Exception` signatures should be avoided

josemduarte (approved):
Looks great to me, thanks @richarda23 

Regarding the change of signature in `GeneSequence.addExon`, I think we can allow it even though 6.0.0 is out already, because a) it is a very minor, b) the 6.0.0 release was sort of invalid because it breaks java-8 run time compatibility, see #996 . We need a 6.0.1 soon to fix that

heuermh:
Thank you, @richarda23 @josemduarte

PR #1001, opened 2021-11-27
Parsing resolution values in PDB (REMARK 2 and REMARK 3 [new format]) and mmCif

### LINKED ISSUE #1000: Resolution lost in cryo-EM
Several entries have the resolution value set to **99** when obtained from PDB and 0 when obtained from mmCif.

Each was cryo-EM with resolutions in the range 3.35 - 3.8 Å.

4 examples are 
(**3j8h**, **3jaj** (available only in mmCIF).
**6hu9**, **7o3e** (available as PDB files).

By experimenting, I found that 
- The PDB entries have the resolution reported in **REMAK 2** (not parsed at all) and in **REMARK 3** (using a format not recognized by the `PDBFileParser`). --> they are left as the default resolution value (**99**).
- The mmCif entries have the `.` value in the `_refine.ls_d_res_high` field, which indicates that the value [_may be used to identify that there is no appropriate value for the item or that a value has been intentionally omitted_](https://mmcif.wwpdb.org/docs/tutorials/mechanics/pdbx-mmcif-syntax.html#:~:text=A%20period%20(.)%20may%20be%20used%20to%20identify%20that%20there%20is%20no%20appropriate%20value%20for%20the%20item%20or%20that%20a%20value%20has%20been%20intentionally%20omitted.) (I don't know why omitted) --> when the period (`.`) is parsed into `float`, it is wrongly parsed as (**0.0**)

aalhossary:
Apparently it affects only cryo-EM entries.

Partially Fixes #1000 
It fixes the PDBFileParser.
Before fixing the mmCifParser we need first to check whether the files should be remediated or not.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/PDBHeader.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/PDBHeader.java
@@ -580,6 +580,16 @@ public class PDBHeader implements PDBRecord {
 		this.crystallographicInfo = crystallographicInfo;
 	}
 
+	/**
+	 * Returns the resolution (or effective resolution) of the experiment. This is
+	 * related to <code>_refine.ls_d_res_high</code> (DIFFRACTION) or
+	 * <code>_em_3d_reconstruction.resolution</code> (ELECTRON MICROSCOPY) for mmCif
+	 * format, or to <code>REMARK 2</code> or <code>REMARK 3</code> for PDB legacy
+	 * format. If more than one value is available (in rare cases), the last one is
+	 * reported. If no value is available, it defaults to 99.
+	 * 
+	 * @return The reported experiment resolution, 99 if no value is available.
+	 */
 	public float getResolution() {
 		return resolution;
 	}
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/PDBHeader.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/PDBHeader.java
@@ -586,9 +586,11 @@ public class PDBHeader implements PDBRecord {
 	 * <code>_em_3d_reconstruction.resolution</code> (ELECTRON MICROSCOPY) for mmCif
 	 * format, or to <code>REMARK 2</code> or <code>REMARK 3</code> for PDB legacy
 	 * format. If more than one value is available (in rare cases), the last one is
-	 * reported. If no value is available, it defaults to 99.
+	 * reported. If no value is available, it defaults to
+	 * {@link #DEFAULT_RESOLUTION} ({@value #DEFAULT_RESOLUTION}).
 	 * 
-	 * @return The reported experiment resolution, 99 if no value is available.
+	 * @return The reported experiment resolution, {@link #DEFAULT_RESOLUTION}
+	 *         ({@value #DEFAULT_RESOLUTION}) if no value is available.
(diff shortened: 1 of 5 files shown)


aalhossary:
I parsed the mmCif data from `_em_3d_reconstruction.resolution` and skipped the `_refine.ls_d_res_high` if it is **not defined**.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/CifStructureConsumerImpl.java
```diff
@@ -831,6 +843,10 @@ public void consumePdbxStructOperList(PdbxStructOperList pdbxStructOperList) {
     public void consumeRefine(Refine refine) {
         for (int rowIndex = 0; rowIndex < refine.getRowCount(); rowIndex++) {
             // RESOLUTION
+        	ValueKind valueKind = refine.getLsDResHigh().getValueKind(rowIndex);
+        	if (ValueKind.NOT_PRESENT.equals(valueKind)) {
```
JonStargaryen, line 847: `if (!ValueKind.PRESENT.equals(valueKind)) {` might be slightly safer here. Possible `ValueKind` values are `PRESENT`, `NOT_PRESENT`, and `UNKNOWN`. The latter 2 would resolve to 0.0 for floats.

JonStargaryen (approved):
Thanks for the changes, LGTM.
1 small remark below.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/CifStructureConsumerImpl.java
aalhossary: Thanks. Done.

aalhossary:
I fixed that. However, there is a new question in #1000 that needs answer before we close that issue and this PR.

aalhossary:
Is it fine now? Shall I merge?

biojava-structure/src/main/java/org/biojava/nbio/structure/PDBHeader.java
```diff
@@ -580,6 +580,16 @@ public void setCrystallographicInfo(PDBCrystallographicInfo crystallographicInfo
 		this.crystallographicInfo = crystallographicInfo;
 	}
 
+	/**
+	 * Returns the resolution (or effective resolution) of the experiment. This is
+	 * related to <code>_refine.ls_d_res_high</code> (DIFFRACTION) or
+	 * <code>_em_3d_reconstruction.resolution</code> (ELECTRON MICROSCOPY) for mmCif
+	 * format, or to <code>REMARK 2</code> or <code>REMARK 3</code> for PDB legacy
+	 * format. If more than one value is available (in rare cases), the last one is
+	 * reported. If no value is available, it defaults to 99.
```
josemduarte, line 589: Nice docs, thanks! It'd be better to point to the variable with `{@value DEFAULT_RESOLUTION}`

josemduarte (approved):
Just one more minor comment. After that it's ok to merge.

biojava-structure/src/main/java/org/biojava/nbio/structure/PDBHeader.java
aalhossary: Done.

PR #1002, opened 2021-11-27
Miscellaneous

aalhossary:


biojava-structure/src/main/java/org/biojava/nbio/structure/HetatomImpl.java
```diff
@@ -168,7 +168,7 @@ public void addAtom(Atom atom){
 				String altLocStr = "";
 				char altLoc = atom.getAltLoc();
 				if (altLoc != ' ') altLocStr = "(alt loc '" + altLoc + "')";
-				logger.warn("An atom with name " + atom.getName() + " " + altLocStr + " is already present in group: " + this.toString() + ". The atom with serial " + atom.getPDBserial() + " will be ignored in look-ups.");
+				logger.warn("An atom with name " + atom.getName() + " " + altLocStr + " is already present in group: " + this.toString() + ". The atom with serial " + existingAtom.getPDBserial() + " will be ignored in look-ups.");
```
josemduarte, line 171: Good catch, thanks!

biojava-core/src/test/java/org/biojava/nbio/core/sequence/io/FastaGeneWriterTest.java
```diff
@@ -58,7 +58,7 @@ void basicGeneWriterTest() throws Exception {
         fastaWriter.process();
 
         String output = new String(os.toByteArray(), "UTF-8");
-        String [] lines =  output.split("\n");
+        String [] lines =  output.split("\r?\n");
```
josemduarte, line 61: How about using `\R`?, see https://www.baeldung.com/java-string-split-by-newline

josemduarte (changes requested):
Thanks!

josemduarte (approved):
Thanks! LGTM

aalhossary:
Thank you for the tip. I didn't know about \R before.

Do you have any idea why it failed twice?

josemduarte:
It looks like there are problems with the SCOP servers. I'll try it again tomorrow.

aalhossary:
I triggered a new build which was successful this time. Will merge now.

Issue #103, opened 2014-03-18
Unit tests need to make sure they reset the state of singletons


josemduarte:
I've come across this very subtle issue recently. I have found that running all tests in one bundle with "mvn install" can have a very different behaviour than running tests individually. Maven by default uses one single JVM to run the tests. Static variables like those used in singletons (e.g. AtomCache) are not necessarily reinitialise at the beginning of each test. Thus the tests have a different outcome when running in a bundle or running separately.

One workaround is always using this setting in the pom.xml:

```
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-surefire-plugin</artifactId>
        <version>2.16</version>
        <configuration>
            <reuseForks>false</reuseForks>
            <forkCount>1</forkCount>
        </configuration>
</plugin>
```

which makes sure that each test is run in a new process (see http://stackoverflow.com/questions/3365628/junit-tests-pass-in-eclipse-but-fail-in-maven-surefire)

For instance if one uses that, then 2 tests fail: `TestBond` and `SiftsChainToUniprotMappingTest` (easily reproducible by just running the tests individually). 

For `TestBond` the issue seems to be a PDB parsing bug in entry 145D. Apparently when running tests in one bundle the flag `setCreateAtomBonds` was not set as expected and then the bug was not seen. 

For `SiftsChainToUniprotMappingTest` the issue seems to be that the sifts file was not downloaded if run individually but it was if run in a bundle. Spencer proposed a fix for that, see this commit https://github.com/biojava/biojava/commit/d6b64f72634c15d53696b1b0ac0b611939067d5c

shivangshah:
Hi .. I am not sure if this bug is already fixed or not but I could reproduce this bug in surefire 2.18.1. Basically we have some singleton objects which should be newly generated per test case but apparently they are not and because of that when a single test case is run it fails, while if we run it using mvn test for the whole project, all test cases pass. The only workaround for now is to nullify the singleton at every @AfterClass and let the new test case reinitialize it .. Anyone got any insights on this?

PR #1017, opened 2022-01-13
Improve MAVEN build Performance

xiayingfeng:
Maven allows you to run the compiler as a separate process by setting `<fork>true</fork>`. This feature can lead to much less garbage collection and make Maven build faster. This project has more than 1000 source files. We can consider enabling this feature.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -176,6 +176,7 @@
 					<configuration>
 						<source>${jdk.version}</source>
 						<target>${jdk.version}</target>
+						<fork>true</fork>
 					</configuration>
 				</plugin>
 				<plugin>


aalhossary:
According to the conversation [here](https://stackoverflow.com/questions/41270364/what-does-forktrue-fork-means-in-pom-xml-while-using-maven), it leads to non repeatable compilation operation.

josemduarte:
I'm reluctant to switch this on because we've had many problems in the past with reproducibilty of tests, mostly because of other issues, like abuse of singletons. See a related conversation: #103 

In any case at best this would provide a minor gain in test performance.

PR #1027, opened 2022-04-03
Update guava dependency version to 31.1-jre

heuermh:


josemduarte (approved):
Thanks @heuermh . Note that the tests are failing due to an unrelated issue with an RCSB REST endpoint. When that is solved we can force re-run and check everything is ok.

heuermh:
Thank you @josemduarte

PR #1031, opened 2022-06-20
Update URL of Assembly CIF File in TestURLBasedFileParsing

JonStargaryen:
- per-assembly files moved to new location 
- https://www.rcsb.org/news/feature/62559153c8eabd0c4864f208


josemduarte (approved):
Good catch, thanks!

Issue #886, opened 2020-08-21
String Formatter error when key or value of Qualifier has character "%"


chenzhiw:
feature.addQualifier("note", new Qualifier("note", "coding sequ%_ece", true));
GenbankWriterHelper.writeNucleotideSequence(......)

```
Exception in thread "main" java.util.UnknownFormatConversionException: Conversion = '_'
        at java.util.Formatter.checkText(Formatter.java:2579)
        at java.util.Formatter.parse(Formatter.java:2555)
        at java.util.Formatter.format(Formatter.java:2501)
        at java.io.PrintWriter.format(PrintWriter.java:905)
        at org.biojava.nbio.core.sequence.io.GenbankWriter.process(GenbankWriter.java:95)
        at org.biojava.nbio.core.sequence.io.GenbankWriterHelper.writeNucleotideSequence(GenbankWriterHelper.java:135)
        at GenBank.main(GenBank.java:41)
```

enilapb:
Have you tested doubling **%%** to scape **%** ?

sbliven:
I think it probably makes sense for BioJava to handle the escaping rather than forcing the user to do it.

Actually, I don't understand why format is getting called in [GenbankWriter.java:95](https://github.com/biojava/biojava/blob/b9519e4c568e8423dcfb9b5fdb06688935360dd7/biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenbankWriter.java#L95) as all. Does calling [PrintWriter.format(s)](https://docs.oracle.com/javase/7/docs/api/java/io/PrintWriter.html#format(java.lang.String,%20java.lang.Object...)) have any effect without additional arguments? It seems like all the formatting (and escaping) should be done within GenericGenbankHeaderFormat.

emckee2006:
IIRC, it converts %n to the platform default line separator. There are a
few flags that don't require arguments.

sbliven:
@emckee2006 Is there any reason why those couldn't be moved inside the HeaderFormat? It would still require escaping some sequences, but at least it only has to happen once.

jamesmorris:
Hello, I have recently come across this issue myself. Is anybody currently working a fix for this problem?

If not I am happy to have a go at proposing a solution

josemduarte:
Thanks @jamesmorris , I don't think anyone is working on it. We're always happy with pull requests :)

emckee2006:
That might not be a bad idea...

PR #1034, opened 2022-07-06
#886 String Formatter error when key or value of Qualifier has character "%"

jamesmorris:
Proposed fix and test for issue #886


josemduarte (approved):
LGTM thanks!

josemduarte:
I'll merge

Issue #942, opened 2021-07-12
GenBankWriter does not use information in the original header


jamesmorris:
Hello,

I have recently been working on adding an accession ID to a few hundred GenBank files using biojava.

I have been using the following approach:
1. First reading the GenBank format files into a DNASequence object using GenbankReaderHelper.readGenbankDNASequence()
2. Then adding the accession ID with the setAccession(new AccessionID("new_accession")) method on the DNASequence object
3. Finally I use GenbankWriterHelper.writeNucleotideSequence() to create an updated GenBank file which includes the new accession

This works for inserting the new accession ID however information is lost in the locus line. As rather than using information for the original file a new line is created using default settings.

For example if I update a GenBank file that contains the following original locus line:
**LOCUS       test_locus_name   9291 BP DS-DNA   CIRCULAR  SYN       13-JUL-1994**

The GenBank file that gets written by  the writeNucleotideSequence() method will look like this:
**LOCUS       new_accession           9291 bp    DNA              circular 12-Jul-2021**

We therefore loose the following:
- 	The locus name
- 	The double stranded prefix on DNA
- 	The domain
- 	The original date

I would argue that there should be another way to write a GenBank file from a DNASequence that could use an original header so no information is lost through the processes of reading and writing the same file.

I would be interested to know what you think about this?

Many thanks,
James

richarda23:
Hi, 
Looking at Genbank spec and example sample record, there is only 1 field to hold the locus name: https://www.ncbi.nlm.nih.gov/Sitemap/samplerecord.html 
which currently takes the accessionId when writing. So there doesn't seem a way to store the original and new one at the same time? Or are you just expecting the 'ACCESSION' row to change?

This is the spec from the above link, so I guess what is correct behaviour is a little undefined.

`However, the 10 characters in the locus name are no longer sufficient to represent the amount of information originally intended to be contained in the locus name. The only rule now applied in assigning a locus name is that it must be unique. For example, for GenBank records that have 6-character accessions (e.g., U12345), the locus name is usually the first letter of the genus and species names, followed by the accession number. For 8-character character accessions (e.g., AF123456), the locus name is just the accession number.`

Looking at the GenbankParser, it ignores fields 4 and 5 of the input document LOCUS row and instead generates them from other information gleaned from the sequence - so it does lose the original MoleculeType and Division information? 

There is only 1 field that modification date can be stored. You could argue that you've modified the Genbank record, so the modificationDate should reflect that? 

I suppose a solution might be
[...]

PR #1042, opened 2022-10-18
GenBankWriter does not use information in the original header #942

jamesmorris:
Hello, 
I have been working on the issue that a GenBank file read in through the parser and then written back out will result in loss of information from the original file. While looking into this I have also fixed a few other issues with the parser and writer for Genbank files. The highlights of this work includes:
- added a new writer helper method that will retain the information in the original locus line
- fixed writing out db_xref feature qualifiers which was not working
- fixed writing out the accession version and GI ID
- allow locus ids that contain spaces
- allow the locus id to be empty
- fixed writing out feature locations containing joins and partial locations
- fixed reading and writing locations on multiple lines
- Maintained quotes around feature qualifier values in files that are written out
- Added a test and test file for feature locations containing joins
- Added a test to check for differences between an original file and one that had been written out by the writer

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericGenbankHeaderFormat.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericGenbankHeaderFormat.java
@@ -20,6 +20,7 @@
  */
 package org.biojava.nbio.core.sequence.io;
 
+import org.biojava.nbio.core.sequence.AccessionID;
 import org.biojava.nbio.core.sequence.compound.AminoAcidCompoundSet;
 import org.biojava.nbio.core.sequence.compound.DNACompoundSet;
 import org.biojava.nbio.core.sequence.compound.RNACompoundSet;
@@ -38,6 +39,7 @@ public class GenericGenbankHeaderFormat<S extends AbstractSequence<C>, C extends
 	private static final int HEADER_WIDTH = 12;
 	private static final String lineSep = "%n";
 	private String seqType = null;
+	private Boolean useOriginalHeader = false;
 
 	public GenericGenbankHeaderFormat() {
 		seqType = null;
@@ -45,6 +47,10 @@ public class GenericGenbankHeaderFormat<S extends AbstractSequence<C>, C extends
 
 	public GenericGenbankHeaderFormat(String seqType) {
 		this.seqType = seqType;
+	}
+	
+	public GenericGenbankHeaderFormat(Boolean useOriginalHeader) {
+		this.useOriginalHeader = useOriginalHeader;
 	}
 
 	/**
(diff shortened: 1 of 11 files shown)


biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericGenbankHeaderFormat.java
```diff
@@ -222,6 +228,22 @@ private String _write_the_first_line(S sequence) {
 		 */
 	}
 
+	/**
+	 * Write the original LOCUS line.
+	 *
+	 * @param sequence
+	 */
+	private String _write_original_first_line(S sequence) {
```
josemduarte, line 236: Nitpick: could you keep the method name as per usual camel case convention?

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericGenbankHeaderFormat.java
```diff
@@ -38,6 +39,7 @@
 	private static final int HEADER_WIDTH = 12;
 	private static final String lineSep = "%n";
 	private String seqType = null;
+	private Boolean useOriginalHeader = false;
```
josemduarte, line 42: Any specific reason not to use the primitive `boolean`?

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericGenbankHeaderParser.java
```diff
@@ -37,15 +37,57 @@
...
 public class GenericGenbankHeaderParser<S extends AbstractSequence<C>, C extends Compound> implements SequenceHeaderParserInterface<S,C> {
 
+	@SuppressWarnings("unused")
+
+	// Brief description of sequence; includes information such as source organism,
+	// gene name/protein name, or some description of the sequence's function (if
+	// the sequence is non-coding). If the sequence has a coding region (CDS),
+	// description may be followed by a completeness qualifier, such as "complete
+	// cds".
+	private String description;
```
josemduarte, line 47: Nice docs! Could you make them into javadocs? i.e. surround comments by `/**   */` . The same for others below

josemduarte (changes requested):
Very nice, thank you! A few minor comments

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericGenbankHeaderFormat.java
jamesmorris: Sure happy to change, I would normally but I was just following the convention already used in the class
e.g. _write_single_line, _write_multi_line etc

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericGenbankHeaderFormat.java
jamesmorris: No, I have now fixed this

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericGenbankHeaderParser.java
jamesmorris: Sure this is now done

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericGenbankHeaderFormat.java
josemduarte: Oh I didn't realise that was the style already for that file. Well I guess we can live with that a bit longer...

emckee2006:
It was initially ported from biopython in a very rudimentary fashion, so
the function names pretty much match the equivalent biopython ones....

josemduarte (approved):
Thanks! LGTM

PR #1044, opened 2022-10-26
Switching to new DNS name for wwpdb http file server

josemduarte:
As per news announcement here: https://www.wwpdb.org/news/news?year=2022#6303be3b707ecd4f63b3d3d9

Also updated the changelog in preparation for next release


JonStargaryen (approved):
LGTM, thanks!

Issue #1051, opened 2023-01-10
Sonar S4973 "==" and "!=" should not be used when "equals" is overridden


jlerbsc:
My company develops software that performs automatic remediation of java code. This software detects defects identified for example by Sonar and automatically applies corrective measures. Maybe you would be interested that we experiment our product on your code? For example, we could submit PRs?

When we try the rule S4973 "==" and "!=" should not be used when "equals" is overridden
https://rules.sonarsource.com/java/RSPEC-1698

Below are the changes made by our software to correct this rule.

    biojava-genome/src/main/java/org/biojava/nbio/genome/homology/GFF3FromUniprotBlastHits.java
    @@ -140,7 +140,7 @@ public class GFF3FromUniprotBlastHits {
                                                }
                                                proteinIndex = proteinIndex + seq.length();
    -                                               if (startIndex != null && endIndex != null && startIndex != endIndex) {
    +                                               if (startIndex != null && endIndex != null && !startIndex.equals(endIndex)) {
                                                        CDSSequence cdsSequence = cdsSequenceList.get(i);
                                                        String hitLabel = "";
                                                        if (transcriptSequence.getStrand() == Strand.POSITIVE) {

    biojava-genome/src/main/java/org/biojava/nbio/genome/homology/GFF3FromUniprotBlastHits.java
    @@ -140,7 +140,7 @@ public class GFF3FromUniprotBlastHits {
                                                }
                                                proteinIndex = proteinIndex + seq.length();
    -                                               if (startIndex != null && endIndex != null && startIndex != endIndex) {
    +                                               if (startIndex != null && endIndex != null && !startIndex.equals(endIndex)) {
                                                        CDSSequence cdsSequence = cdsSequenceList.get(i);
                                                        String hitLabel = "";
                                                        if (transcriptSequence.getStrand() == Strand.POSITIVE) {

    biojava-structure/src/main/java/org/biojava/nbio/structure/ecod/EcodInstallation.java
    @@ -373,7 +373,7 @@ public class EcodInstallation implements EcodDatabase {
                                return false;
                        // Re-download old copies of "latest"
    -                       if(updateFrequency != null && requestedVersion == DEFAULT_VERSION ) {
    +                       if(updateFrequency != null && requestedVersion.equals(DEFAULT_VERSION) ) {
                                long mod = f.lastModified();
                                // Time of last update
                                Date lastUpdate = new Date();

[...]

PR #1052, opened 2023-01-13
Sonar S4973 '==' and '!=' should not be used when 'equals' is overridden

jlerbsc:
These modifications were made by an automatic java code remediation software. They are related to the rule S4973 "==" and "!=" should not be used when "equals" is overridden.

https://rules.sonarsource.com/java/RSPEC-1698

**General rule for comparing object**

It is equivalent to use the equality == operator and the equals method to compare two objects if the equals method inherited from Object has not been overridden. In this case both checks compare the **object references**.
But as soon as equals is overridden, two objects not having the same reference but having the same value can be equal.

**Why should this rule also be applied to primitive types?**

Because in the case of classes wrapping primitive types, the operation of the equality test depends on the implementation of the JVM and the optimizations that have been made.

Usage of the == and != operators for comparing the values of fully memoized boxed primitive types is permitted. 
But usage of the == and != operators for comparing the values of boxed primitive types that are not fully memoized is permitted only when the range of values represented is guaranteed to be within the ranges specified by the JLS to be fully memoized.
And usage of the == and != operators for comparing the values of boxed primitive types is not allowed in all other cases.

JLS explanation is here https://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.1.7

### DIFF
--- a/biojava-genome/src/main/java/org/biojava/nbio/genome/homology/GFF3FromUniprotBlastHits.java
+++ b/biojava-genome/src/main/java/org/biojava/nbio/genome/homology/GFF3FromUniprotBlastHits.java
@@ -140,7 +140,7 @@ public class GFF3FromUniprotBlastHits {
 						}
 
 						proteinIndex = proteinIndex + seq.length();
-						if (startIndex != null && endIndex != null && startIndex != endIndex) {
+						if (startIndex != null && endIndex != null && !startIndex.equals(endIndex)) {
 							CDSSequence cdsSequence = cdsSequenceList.get(i);
 							String hitLabel = "";
 							if (transcriptSequence.getStrand() == Strand.POSITIVE) {
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/ecod/EcodInstallation.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/ecod/EcodInstallation.java
@@ -373,7 +373,7 @@ public class EcodInstallation implements EcodDatabase {
 				return false;
 
 			// Re-download old copies of "latest"
-			if(updateFrequency != null && requestedVersion == DEFAULT_VERSION ) {
+			if(updateFrequency != null && requestedVersion.equals(DEFAULT_VERSION) ) {
 				long mod = f.lastModified();
 				// Time of last update
 				Date lastUpdate = new Date();
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/io/BondMaker.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/io/BondMaker.java
@@ -215,7 +215,7 @@ public class BondMaker {
 			for (Atom a2:a2s) {
 				if (a1.getAltLoc() != null && a2.getAltLoc()!=null &&
 						a1.getAltLoc()!=' ' && a2.getAltLoc()!=' ' &&
-						a1.getAltLoc() != a2.getAltLoc()) {
+						!a1.getAltLoc().equals(a2.getAltLoc())) {
 					logger.debug("Skipping bond between atoms with differently named alt locs {} (altLoc '{}') -- {} (altLoc '{}')",
 							a1.toString(), a1.getAltLoc(), a2.toString(), a2.getAltLoc());
 					continue;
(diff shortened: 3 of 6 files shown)


aalhossary:
In addition to the case of primitive values, I mentioned 3 concerns regarding the cases when one of the operands is `null` and the other is not [here](https://github.com/biojava/biojava/issues/1051#issuecomment-1377605662).

jlerbsc:
> In addition to the case of primitive values, I mentioned 3 concerns regarding the cases when one of the operands is `null` and the other is not [here](https://github.com/biojava/biojava/issues/1051#issuecomment-1377605662).

Checking that an object is not null before invoking a method on that object is a preventative or protective measure. Your software evolves and what is true today may not be true tomorrow so this check is a protective measure.

aalhossary:
Writing good software is like writing poetry: There are both science and art in it. Please spend some time to understand my concern(s) before replying, instead of blindly accepting any machine-proposed modifications.

Let's take the last proposed change as an example:

In the original code
```Java
if (seqClusterId.get(i) != seqClusterId.get(j)) {
    return false;
}
```
The `if` condition will test whether `seqClusterId.get(i)` (the 1st expression) `!=` (is not equal to) `seqClusterId.get(j)` (the 2nd expression).
That is `true` in these 3 cases

- 1st expression is null and 2nd expression is not null
- 1st expression is not null and 2nd expression is null
- 1st expression is not null and 2nd expression is not null and their values and not equal


Well,
What about the proposed update?
```Java
if (seqClusterId.get(i) != null && !seqClusterId.get(i).equals(seqClusterId.get(j))) {
    return false;
}
```

It will evaluate to `true` in these 2 cases only:
- 1st expression is not null and 2nd expression is null
- 1st expression is not null and 2nd expression is not null and their values and not equal

but it will evaluate to `false` in the case
- 1st expression is null and 2nd expression is not null

which is a change of behavior.

jlerbsc:
You are right.

jlerbsc:
Thank you for your relevant analysis of our proposal. We are going to take your comments into consideration and adapt the way our solution works to avoid applying this code transformation rule when equality tests are performed on boxed primitive types or strings.

tmrpavan (approved):

aalhossary:
@tmrpavan Your account looks very fresh (joined GitHub 2 weeks ago). May you please introduce yourself?
@josemduarte Is approving pull requests available to anyone, or only to people with certain privileges (e.g. write access)?

tmrpavan:
Hello @aalhossary, 

You are right I am new to open source project. I am Java Backed Developer with around 3 years of experience.  I was going through the PR. I felt the changes are valid. So I approved. 

I think approve option is available for all.

tmrpavan:
Hello @aalhossary, 

I would like to contribute to the project. Please let me know if any issue ore features that i can work on it.

tmrpavan:
image

tmrpavan:
Why Avoid Comparing Integers in Java using == Operator? Example

Read more: https://javarevisited.blogspot.com/2010/10/what-is-problem-while-using-in.html#ixzz7yTZDXO94https://javarevisited.blogspot.com/2010/10/what-is-problem-while-using-in.html#axzz7yTYv7igN

aalhossary:
Hi @tmrpavan!
Welcome. We always welcome new committers.

as you found that the changes are valid, but there was an ongoing discussion about it, you should have joined the discussion and add a comment instead of rushing into doing the actions.
I advise you to read the discussion on THE CURRENT page (especially [this](https://github.com/biojava/biojava/pull/1052#issuecomment-1385124659) comment).

tmrpavan:
@aalhossary Please see this PR https://github.com/biojava/biojava/pull/1060

We can use Objects.equals(seqClusterId.get(i), seqClusterId.get(j) to full fill your needs.

Please see the below image and article to know why we can not user == or != or objects.   

https://javarevisited.blogspot.com/2010/10/what-is-problem-while-using-in.html#ixzz7yTZDXO94https://javarevisited.blogspot.com/2010/10/what-is-problem-while-using-in.html#axzz7yTYv7igN

image

josemduarte:
Closing in favour of #1060

PR #1057, opened 2023-04-03
POM Cleanup: Add Maven descriptors & remove mention of Travis CI

JonStargaryen:
- need Maven descriptors for `pom.properties` files when using BioJava as dependency
- this ensures that they are added to the artifacts


josemduarte (approved):
LGTM thanks.

There's an unrelated test failure. It seems due to an obsoleted PDB id. I'll try to push a fix to master, so that then this can be tested.

josemduarte:
Ok I committed a fix: could you merge master in, so that tests run?

josemduarte:
Actually there were a couple more test issues (after recent remediation of some PDB entries). Now I think I fixed them all. Please try again.

JonStargaryen:
Thanks for updating the tests. Let's see how this goes.

PR #1059, opened 2023-04-07
Additional refactoring to maintain cohesion and solid principles.

xcalibur91:
Hello,
I have refactored some code in the project.

Refactor : Rename Variable
Description : The variables didn’t explain the exact use. The fetched value was a fraction and not percentage. The Aliphatic index is converted into percentage at the end. The formulation was also broken down for readability.

Refactor : Extract Class
Description: Separated file network operations and basic read write operations to different utility class to have single principle property. 

Refactor : Move method
Description : The method convertDNAtoProteinSequence in ProteinMappingTools.java was very specific to DNASequence.java class and hence the code was refactored to move the method to maintain cohesion.

Refactor : Pull up method
Description : The GeneSequence class along with other classes qualify for Broken Hierarchy Design smell. Upon further analysis, it can be seen that the getLength() method can be pulled up the hierarchy. That is what this refactoring does.
A new class DNAExtractSequence is introduced to handle this situation which extends the original DNASequence.

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/util/FileOperationUtils.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/util/FileOperationUtils.java
@@ -1,0 +1,151 @@
+package org.biojava.nbio.core.util;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.nio.channels.FileChannel;
+import java.nio.file.*;
+import java.nio.file.attribute.BasicFileAttributes;
+
+/**
+ * @author Sreejith Nair
+ */
+public class FileOperationUtils {
+    /**
+     * @param src
+     * @param dst
+     * @throws IOException
+     */
+    @SuppressWarnings("resource")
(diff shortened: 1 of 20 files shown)


biojava-core/src/main/java/org/biojava/nbio/core/util/FileOperationUtils.java
```diff
@@ -0,0 +1,151 @@
...
+ * @author Sreejith Nair
+ */
+public class FileOperationUtils {
+    /**
+     * @param src
+     * @param dst
+     * @throws IOException
+     */
+    @SuppressWarnings("resource")
+    public static void copy(File src, File dst) throws IOException {
```
josemduarte, line 21: Since this PR is all about cleaning up, how about taking this opportunity to remove this method and replace it by the one provided in the java library (see TODO comment in original javadocs above)

josemduarte (changes requested):
I think the changes look good. But it'd be nice to clean the file copy function as commented below.

One general comment: please try to create one PR per concern. This one seems to have 3 independent concerns in it.

josemduarte:
I'll close for lack of activity.

PR #1060, opened 2023-04-16
Sonar S4973 '==' and '!=' should not be used when 'equals' is overrid…

tmrpavan:
Avoid Comparing Integers in Java using == Operator

image

### DIFF
--- a/biojava-genome/src/main/java/org/biojava/nbio/genome/homology/GFF3FromUniprotBlastHits.java
+++ b/biojava-genome/src/main/java/org/biojava/nbio/genome/homology/GFF3FromUniprotBlastHits.java
@@ -140,7 +140,7 @@ public class GFF3FromUniprotBlastHits {
 						}
 
 						proteinIndex = proteinIndex + seq.length();
-						if (startIndex != null && endIndex != null && startIndex != endIndex) {
+						if (startIndex != null && endIndex != null && startIndex.equals(endIndex)) {
 							CDSSequence cdsSequence = cdsSequenceList.get(i);
 							String hitLabel = "";
 							if (transcriptSequence.getStrand() == Strand.POSITIVE) {
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/io/BondMaker.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/io/BondMaker.java
@@ -214,8 +214,8 @@ public class BondMaker {
 		for (Atom a1:a1s) {
 			for (Atom a2:a2s) {
 				if (a1.getAltLoc() != null && a2.getAltLoc()!=null &&
-						a1.getAltLoc()!=' ' && a2.getAltLoc()!=' ' &&
-						a1.getAltLoc() != a2.getAltLoc()) {
+						!a1.getAltLoc().equals(' ') && !a2.getAltLoc().equals(' ') &&
+						a1.getAltLoc().equals(a2.getAltLoc())) {
 					logger.debug("Skipping bond between atoms with differently named alt locs {} (altLoc '{}') -- {} (altLoc '{}')",
 							a1.toString(), a1.getAltLoc(), a2.toString(), a2.getAltLoc());
 					continue;
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/symmetry/core/RotationSolver.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/symmetry/core/RotationSolver.java
@@ -21,12 +21,7 @@
 
 package org.biojava.nbio.structure.symmetry.core;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 
 import javax.vecmath.AxisAngle4d;
 import javax.vecmath.Matrix4d;
@@ -90,7 +85,7 @@ public class RotationSolver implements QuatSymmetrySolver {
 		// isSpherical check added 08-04-11
 		if (maxSymOps % 60 == 0 && isSpherical) {
 			maxSymOps = 60;
-		 }
+		}
 
 		AxisAngle4d sphereAngle = new AxisAngle4d();
 		Matrix4d transformation = new Matrix4d();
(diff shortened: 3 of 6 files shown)


tmrpavan:
@aalhossary Can you please review this PR

josemduarte:
I think auto-unboxing takes care of this in most cases. For instance if I rewrite your first test as:
```java
Integer m = 1;
Integer n = 1;
System.out.println(m==n);
```
The output is `true`

josemduarte:
Interestingly this doesn't work:
```java
Integer m = new Integer(1);
Integer n = new Integer(1);
System.out.println(m==n);
```
but this works:
```java
Integer m = Integer.valueOf(1);
Integer n = Integer.valueOf(1);
System.out.println(m==n);
```

aalhossary:
@josemduarte That is because of the internal implementation of `valueOf()`
It first tries to find a cached/cachable value of the parameter in a static HashTable of `Integer`. If found, it returns that staticallic cached object.
`new Integer` on the other hand, creates a **new object** in all cases.

Therefore, checking the equality of two autoboxed wrappers using `==` is safe within a range (as far as I remember in the range [-127, 128]).


However, the situation in this pull request (and the previous one) is actually about "not equal".

tmrpavan:
comparison rather than equality comparison. When using the "not equal" operator (!=) to compare two autoboxed Integer values, the result is not guaranteed to be correct, even within the range you mentioned.

This is because the "not equal" operator first unboxes the Integer values into primitive int values and then compares those values. If the two Integer values being compared happen to be in the cached range, they will be the same object reference and thus have the same primitive value. However, if they are not in the cached range, they will be different object references even if they have the same primitive value, and the "not equal" comparison will return false when it should have returned true.

To avoid this issue, it's recommended to use the equals() method for comparing two autoboxed Integer values instead of the "not equal" operator, as it compares the values of the objects rather than their references.

tmrpavan:
@aalhossary See the implementation on Objects.equals();

It first compares the objects with == operator. 
1-> If bother are in cached range if both are same the first expression return true.
2-> If any one of the object is not in cached range if first expression return is flase for some case that case only it is going to check the values.

I hope using equals method safe.

public static boolean equals(Object a, Object b) {
        return (a == b) || (a != null && a.equals(b));
    }
I
equals ion Integer class
public boolean equals(Object obj) {
        if (obj instanceof Integer) {
            return value == ((Integer)obj).intValue();
        }
        return false;
    }

tmrpavan:
@aalhossary
Is there any side effects using Objects.equals(). As per you previous post -> https://github.com/biojava/biojava/pull/1052#issuecomment-1385124659 it gives expected results for all the scenarios.

aalhossary:
@tmrpavan using `Objects.equals()` instead of `Integer.equals()` (and instead of `==` of course) does the right job.

aalhossary:
@josemduarte Now both this PR and #1052 address the same issue (i.e. duplicates).
What shall we do? close the old one and review the new one, or close the new one and ask the old one OP for a modifications?

tmrpavan:
@aalhossary Any update on this

josemduarte:
> Now both this PR and https://github.com/biojava/biojava/pull/1052 address the same issue (i.e. duplicates).
What shall we do? close the old one and review the new one, or close the new one and ask the old one OP for modifications?

Since #1052 has gone quiet (specially from the PR creator side). Let's close that one in favour of this. Still we'll want to review this PR of course.

biojava-genome/src/main/java/org/biojava/nbio/genome/homology/GFF3FromUniprotBlastHits.java
```diff
@@ -140,7 +140,7 @@ PairwiseSequenceAlignerType.LOCAL, new SimpleGapPenalty(),
 						}
 
 						proteinIndex = proteinIndex + seq.length();
-						if (startIndex != null && endIndex != null && startIndex != endIndex) {
+						if (startIndex != null && endIndex != null && startIndex.equals(endIndex)) {
```
aalhossary, line 143: Do you mean **not** equals()?

biojava-structure/src/main/java/org/biojava/nbio/structure/io/BondMaker.java
```diff
@@ -214,8 +214,8 @@ private void formBondAltlocAware(Group g1, String name1, Group g2, String name2,
 		for (Atom a1:a1s) {
 			for (Atom a2:a2s) {
 				if (a1.getAltLoc() != null && a2.getAltLoc()!=null &&
-						a1.getAltLoc()!=' ' && a2.getAltLoc()!=' ' &&
-						a1.getAltLoc() != a2.getAltLoc()) {
+						!a1.getAltLoc().equals(' ') && !a2.getAltLoc().equals(' ') &&
+						a1.getAltLoc().equals(a2.getAltLoc())) {
```
aalhossary, line 218: **not** a1.getAltLoc().equals(a2.getAltLoc())

biojava-structure/src/main/java/org/biojava/nbio/structure/symmetry/core/RotationSolver.java
```diff
@@ -21,12 +21,7 @@
 
 package org.biojava.nbio.structure.symmetry.core;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
aalhossary, line 24: This is dangerous. Unless the rule has changed, the norm is to import the exact files/classes you need only.

biojava-structure/src/main/java/org/biojava/nbio/structure/symmetry/core/SystematicSolver.java
```diff
@@ -31,10 +31,7 @@
 import javax.vecmath.Quat4d;
 import javax.vecmath.Vector3d;
 
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
aalhossary, line 34: Same comment.

aalhossary (changes requested):
Please see my comments on my modifications

tmrpavan:
I am working on this

tmrpavan:
@aalhossary Modified as per the comments

aalhossary (approved):
LGTM. Thank you!

PR #1063, opened 2023-05-24
Some more categories written out to CIF 

josemduarte:
Adding `_entity` and `_entity_poly` to CIF output, so that some more metadata is available for consumers.

Also: 

- now adding seqres groups to `Chain`s when no entity info is present in input file
- upgrading to latest ciftools-java

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/AbstractCifFileSupplier.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/AbstractCifFileSupplier.java
@@ -1,12 +1,6 @@
 package org.biojava.nbio.structure.io.cif;
 
-import org.biojava.nbio.structure.Atom;
-import org.biojava.nbio.structure.Chain;
-import org.biojava.nbio.structure.Element;
-import org.biojava.nbio.structure.EntityType;
-import org.biojava.nbio.structure.Group;
-import org.biojava.nbio.structure.GroupType;
-import org.biojava.nbio.structure.Structure;
+import org.biojava.nbio.structure.*;
 import org.biojava.nbio.structure.xtal.CrystalCell;
 import org.biojava.nbio.structure.xtal.SpaceGroup;
 import org.rcsb.cif.CifBuilder;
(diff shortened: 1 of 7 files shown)


biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/AbstractCifFileSupplier.java
```diff
@@ -92,6 +90,49 @@ protected CifFile getInternal(Structure structure, List<WrappedAtom> wrappedAtom
...
+            String[] entityDescriptions = new String[entityInfos.size()];
+
+            for (int i=0; i<entityInfos.size(); i++) {
+                EntityInfo e = entityInfos.get(i);
+                entityIds[i] = Integer.toString(e.getMolId());
+                entityTypes[i] = e.getType().getEntityType();
+                entityDescriptions[i] = e.getDescription() == null? "?" : e.getDescription();
+            }
+
+            String[] polyEntityIds = entityInfos.stream().filter(e -> e.getType() == EntityType.POLYMER).map(e -> Integer.toString(e.getMolId())).collect(Collectors.toList()).toArray(new String[]{});
```
JonStargaryen, line 106: Streams have a `toArray` method as well. `.collect(Collectors.toList()).toArray(new String[]{})` can be simplified to `.toArray(String[]::new)`, effectively skipping the creation of a `List`.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/AbstractCifFileSupplier.java
```diff
@@ -92,6 +90,49 @@ protected CifFile getInternal(Structure structure, List<WrappedAtom> wrappedAtom
...
+            String[] polyEntitySeqs = entityInfos.stream().filter(e -> e.getType() == EntityType.POLYMER).map(e -> e.getChains().get(0).getSeqResSequence()).collect(Collectors.toList()).toArray(new String[]{});
```
JonStargaryen, line 107: Same as above.

JonStargaryen (changes requested):
Thanks, code looks good and should solve our problem.
There are 2 stream expressions that can be simplified.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/AbstractCifFileSupplier.java
josemduarte: Nice one, thanks. Added a commit. 

And shame to IntelliJ for not suggesting this!

JonStargaryen (approved):
LGTM.

PR #1065, opened 2023-06-21
Second fix for sonar rule 's1130' throws declarations should not be superfluous

jlerbsc:
This fix is specific to constructor declarations.


josemduarte (approved):
LGTM thanks

Issue #1066, opened 2023-06-25
Suggestion: Sonar S2129 Constructors should not be used to instantiate "String", "BigInteger", "BigDecimal" and primitive-wrapper classes


jlerbsc:
While analyzing your project using our Automatic bug fixing software, i noticed that sonar rule S2129 was sometimes violated. If you wish, I can offer you a correction in the form of a PR.

https://sonarcloud.io/organizations/default/rules?open=java%3AS2129&q=S2129
"Calling constructors for String, BigInteger, BigDecimal and the objects used to wrap primitives is less efficient and less clear than relying on autoboxing or valueOf.
Consider simplifying when possible for more efficient and cleaner code."

Below are examples of the changes made by our software to correct this rule.

    diff --git a/biojava-alignment/src/main/java/org/biojava/nbio/alignment/io/StockholmStructure.java b/biojava-alignment/src/main/java/org/biojava/nbio/alignment/io /StockholmStructure.java
    index 97261433b..22783b43c 100644
    --- a/biojava-alignment/src/main/java/org/biojava/nbio/alignment/io/StockholmStructure.java
    +++ b/biojava-alignment/src/main/java/org/biojava/nbio/alignment/io/StockholmStructure.java
    @@ -256.9 +256.8 @@ public class StockholmStructure {
                         }
                         String[] seqDetails = splitSeqName(sequencename);
                         seq.setDescription(seqDetails[0]);
    - seq.setBioBegin((seqDetails[1] == null || seqDetails[1].trim().equals("") ? null: new Integer(
    - seqDetails[1])));
    - seq.setBioEnd((seqDetails[2] == null || seqDetails[2].trim().equals("") ? null: new Integer(seqDetails[2])));
    + seq.setBioBegin((seqDetails[1] == null || seqDetails[1].trim().equals("") ? null : Integer.valueOf(seqDetails[1])));
    + seq.setBioEnd((seqDetails[2] == null || seqDetails[2].trim().equals("") ? null: Integer.valueOf(seqDetails[2])));

                         seqs.add(seq);
                 }

josemduarte:
That seems reasonable to me, thanks. Happy to review a PR.

jlerbsc:
The corresponding PR is https://github.com/biojava/biojava/pull/1067

PR #1067, opened 2023-06-27
Fix Sonar S2129 Constructors should not be used to instantiate 'String', …

jlerbsc:
…'BigInteger', 'BigDecimal' and primitive-wrapper classes (Issue #1066)

This PR fixe all or part of the sonar violation S2129 whose sonar recommendation is to consider simplifying when possible for more efficient and cleaner code.


josemduarte (approved):
LGTM thanks!

PR #1070, opened 2023-06-29
Fixing genbank export new line issue

### LINKED ISSUE #1069: GenBank export producing spurious %n
Hi,

I'm having some problems here trying to write genomic data to Genbank format with a
`GenbankWriter<DNASequence, NucleotideCompound>` instance; instead of proper
line breaks in the Genbank header and list of features, I end up with `%n` instead of
`\n` as proper line ends. 

This is caused by https://github.com/biojava/biojava/blob/master/biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericGenbankHeaderFormat.java#L40 and https://github.com/biojava/biojava/blob/master/biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericInsdcHeaderFormat.java#L52

I was able to circumvent this by copying all required methods (since they are private) and replacing
the `lineSep` with `\n`, but this obviously isn't the right way to go.

Biojava version is 7.0.2.

Any ideas?

josemduarte:
Should fix #1069

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericGenbankHeaderFormat.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericGenbankHeaderFormat.java
@@ -37,7 +37,7 @@ public class GenericGenbankHeaderFormat<S extends AbstractSequence<C>, C extends
 		extends GenericInsdcHeaderFormat<S, C> implements
 		GenbankHeaderFormatInterface<S, C> {
 	private static final int HEADER_WIDTH = 12;
-	private static final String lineSep = "%n";
+	private static final String lineSep = System.lineSeparator();
 	private String seqType = null;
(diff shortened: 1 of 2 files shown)


aalhossary:
Thank you @josemduarte for the fix. But I believe 2 files should be updated not only one:

- [ ] - GenericGenbankHeaderFormat.java (L40)
- [ ] - GenericInsdcHeaderFormat.java (L52)

aalhossary (changes requested):
Please include a fix to [GenericInsdcHeaderFormat.java#L52](https://github.com/biojava/biojava/blob/master/biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/GenericInsdcHeaderFormat.java#L52) as well.

josemduarte:
Good catch thanks! Now pushed

aalhossary (approved):
Thank you. LGTM

PR #1078, opened 2024-01-16
Update guava dependency version to 33.0.0-jre

heuermh:


heuermh:
Any way to kick the JDK 11 action?  It seems stuck

josemduarte:
> Any way to kick the JDK 11 action? It seems stuck

Good question, I've been wondering about that myself for some time. 

Checking our workflows (https://github.com/biojava/biojava/blob/a3d943783b1d6e1958331b64cd33e3c6009d028a/.github/workflows/pull_request.yml#L22) it doesn't seem that we have 11 anymore. So it seems that there is some issue with github updating its state after this kind of workflow change (??)

heuermh:
Ah, that is strange!

PR #1077, opened 2024-01-08
Changed the SCOP Installation URLs

hilbertglm:
I was going through the tutorial and was displaying a jmol panel.  When I selected Show SCOP Domains for the color on the selection, I received a `FileNotFoundException`.   A review of the Berkeley site shows that they changed from http to https for the scheme.   They also appear to have changed the path portion of the URL in version 2 from 'scop' to 'scope.'

All of the unit tests worked before and after this change.  I tried to create a unit test to show the failure, fix it, and run the unit test to show that it was fixed, but I couldn't quite follow the path to get to the ScopInstallation code from the Swing application.

To recreate my original error, the following code can be used (derived from the tutorial).

```
         Structure structure = StructureIO.getStructure("4FRW");
         System.out.println(StructureTools.getNrAtoms(structure));
         StructureAlignmentJmol jmolPanel = new StructureAlignmentJmol();

         jmolPanel.setStructure(structure);

         // send some commands to Jmol
         jmolPanel.evalString("select * ; color chain;");
         jmolPanel.evalString("select *; spacefill off; wireframe off; cartoon on;  ");
         jmolPanel.evalString("select ligands; cartoon off; wireframe 0.3; spacefill 0.5; color cpk;");
```

PR #1080, opened 2024-01-18
Added Java Streaming I/O for FASTA Files

hilbertglm:
Added a class to use Java streams to process FASTA files.


biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/FastaStreamer.java
```diff
@@ -0,0 +1,179 @@
...
+	/**
+	 * Create a stream of protein sequences from the contents of the path
+	 * @return the stream
+	 */
+	public Stream<ProteinSequence> stream() {
+		InputStreamProvider provider = new InputStreamProvider();
+		InputStream input;
+		try {
+			input = provider.getInputStream(getPath().toFile());
+		} catch (IOException exception) {
```
josemduarte, line 99: Wouldn't it be better to throw the exception so that it can be handled at a higher level?

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/FastaStreamer.java
```diff
@@ -0,0 +1,179 @@
...
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/**
+ * Read from a FASTA file (or gzipped FASTA file) and create a Java stream of {@link ProteinSequence} objects
+ * for use in a functional programming paradigm.
+ *
+ * @author Gary Murphy
+ * @since 7.0.3
```
josemduarte, line 30: I think we should bump up to 7.1.0 since this introduces a (very nice) new feature

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/FastaStreamer.java
```diff
@@ -0,0 +1,179 @@
...
+						iterator = chunk.entrySet().iterator();
+					}
+					if (iterator.hasNext()) {
+						Map.Entry<String, ProteinSequence> entry = iterator.next();
+						return createSequence(entry.getKey(), entry.getValue());
+					}
+					closed = true;
+					reader.close();
+				} catch (IOException exception) {
+					throw new RuntimeException(String.format("I/O error reading the FASTA file from '%s'", getPath()));
```
josemduarte, line 143: Same comment as above

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/FastaStreamer.java
josemduarte: Or if `stream()` doesn't permit a throws in the signature, then I'd advice to use an `UncheckedIOException` instead of `RuntimeException`

biojava-structure/src/main/java/org/biojava/nbio/structure/scop/ScopInstallation.java
```diff
@@ -85,8 +85,8 @@ public class ScopInstallation implements LocalScopDatabase {
 	public static final String comFileName = "dir.com.scop.txt_";
 
 	// Download locations
-	public static final String SCOP_DOWNLOAD = "http://scop.berkeley.edu/downloads/parse/";
-	public static final String SCOP_DOWNLOAD_ALTERNATE = "http://scop.berkeley.edu/downloads/parse/";
```
josemduarte, line 89: It looks like these changes were from #1077  . I've merged that one now. Could you merge master in to have a cleaner diff?

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/FastaStreamer.java
```diff
@@ -0,0 +1,179 @@
...
+	/**
+	 * Create the sequence with the information from the header.  This implementation return the sequence as-is, but
+	 * this is an opportunity for the implementer to build specific information into the user collection space
+	 * of the sequence
+	 *
+	 * @param header the original header
+	 * @param sequence the protein sequence
+	 * @return the sequence
+	 */
+	protected ProteinSequence createSequence(String header, ProteinSequence sequence) {
```
josemduarte, line 160: The `header` parameter is not used, is that intentional?

josemduarte (changes requested):
Very nice feature, thanks! A couple of comments below. Apologies for the slow response.

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/FastaStreamer.java
hilbertglm: The exceptions would break stream chaining, as the contrived example below shows.  I changed to `UncheckedIOException`.  I originally had the exception being thrown, so it is allowed by Java, but it caused me problems when I used it in my code, so I get where you are coming from.

```java
		String file = this.getClass().getResource("PF00104_small.fasta.gz").getFile();
		String file2 = this.getClass().getResource("PF00105_small.fasta.gz").getFile();
		List<Path> paths = List.of(Paths.get(file), Paths.get(file2));
		List<ProteinSequence> sequences = paths
				.stream()
				.flatMap(path -> FastaStreamer.from(path).stream()) // <-- Not allowed if exception
				.collect(Collectors.toList());
```

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/FastaStreamer.java
hilbertglm: I changed to 7.1.0-SNAPSHOT

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/FastaStreamer.java
hilbertglm: Changed to `UncheckedIOException`

biojava-structure/src/main/java/org/biojava/nbio/structure/scop/ScopInstallation.java
hilbertglm: Dang. I thought I was.  This is my first time dealing with a forked repo.  I guess I will use the webapp instead of CLI.  (Sigh)

biojava-core/src/main/java/org/biojava/nbio/core/sequence/io/FastaStreamer.java
hilbertglm: Good catch.  The was in there before I found `getOriginalHeader`.

hilbertglm:
I am not sure how to fix the diff issue.  I did the `Sync fork` on the GitHub UI.  Do I need to create another pull request?  Sorry... I am new to opensource contributions.

josemduarte:
Simply merge the latest master into your branch: you do that locally in your computer, then push to your fork. Then github will automagically pick it up and you will see an updated PR.

josemduarte (changes requested):
Thanks for the changes. One final request from me: could you add a new `7.1.0 - future release` section in the changelog and describe this new feature briefly?

heuermh:
This looks pretty cool -- may I ask is it primarily about new syntax/language features, or are there some performance considerations?  If the caller is simply going to `collect()` at the end, it likely doesn't matter.

josemduarte (approved):
Thanks again! I'll try to create a new release in the next days

hilbertglm:
@heuermh It is more to do with liking in the streams programming style.  I do all of my alignment testing, homologies, etc. using streams-based processing.  I haven't done any benchmarking, but I would expect it to be pretty close to imperative loops.

Issue #1081, opened 2024-02-23
Quality: Sonar S1132 Strings literals should be placed on the left side when checking for equality


jlerbsc:
We noticed that you have 495 violations of the Sonar rule RSPEC-1132 "Strings literals should be placed on the left side when checking for equality". It is preferable to place string literals on the left-hand side of an equals() or equalsIgnoreCase() method call. This prevents null pointer exceptions from being raised, as a string literal can never be null by definition.

https://rules.sonarsource.com/java/RSPEC-1132/

You can find another reference to this problem here https://en.wikibooks.org/wiki/Java_Programming/Preventing_NullPointerException (section "Comparing string variable with a string literal")

The next version of our software, which will be deployed no later than 26/02/2024, will automatically correct this issue and free of charge.  You can find links and documentation here https://www.indepth.fr/

If you prefer, we can also submit a PR, but using our software is very simple.

Below is an example of the problem identified and the solution provided by Indepth.

    org\biojava\nbio\core\sequence\io\ABITrace.java
    @@ -192,5  +192,5 @@
	 */
	public int[] getTrace (String base) throws CompoundNotFoundException {
    -		if (base.equals("A")) {
    +		if ("A".equals(base)) {
			return A;
    -		} else if (base.equals("C")) {
    +		} else if ("C".equals(base)) {

PR #1082, opened 2024-02-26
fix: issue #1081 Sonar S1132 Strings literals should be placed on the left side when checking for equality

jlerbsc:
We are aware that you may be perplexed about the ability of the Indepth solution to improve the quality of the project's code in depth without adding bugs. That's why, to reassure you, we suggest this PR.


josemduarte (approved):
Thanks! very nice usage of your tool. This should further improve our codesmell sonarqube scores (which now are already getting an A. Our "bugs" score is terrible on the other hand).

PR #1083, opened 2024-03-02
More lenient mmCIF parsing

josemduarte:
[ESMAtlas](https://esmatlas.com/) files miss some of the atom_site fields, which was breaking their parsing. This makes parsing more lenient, not requiring these fields if not present.


heuermh (approved):

JonStargaryen (approved):
LGTM, thanks.

PR #1084, opened 2024-03-04
Replacing home-made copy by Files.copy

josemduarte:
And other minor fixes.

Trying also to test if github workflows is working after changes (1b5fa3c503d5d071f66f0a59ebbe396ebb2f6181)


josemduarte:
Still problems with workflows, I'll close and reopen

Issue #789, opened 2018-08-07
Logger/string formatting slows down some calculations


biocryst:
Calculating secondary structure of 4hhb 10000 times takes 589 seconds on my machine. If the following two lines are commented out, it takes 34 seconds:

https://github.com/biojava/biojava/blob/8a7b651afc1bd75bf70f03d309250058a45b982d/biojava-structure/src/main/java/org/biojava/nbio/structure/secstruc/SecStrucCalc.java#L851-L856

https://github.com/biojava/biojava/blob/8a7b651afc1bd75bf70f03d309250058a45b982d/biojava-structure/src/main/java/org/biojava/nbio/structure/secstruc/SecStrucCalc.java#L869-L870

The problem is likely to be in the String.format method. Using SLF4J parameterized messages instead ("{}") is much faster, but does not support float formatting to number of decimal places (see https://stackoverflow.com/questions/22720865/formatting-floating-point-numbers-in-slf4j, thanks @josemduarte ).

My suggestion is to go through the logging messages which use string formatting and, if feasible, change them to parameterized messages. 

The performance is noticeably affected only if the logging statements are somewhere deep within a loop.

josemduarte:
Good catch! Indeed this shows how careful use of the loggers is very important. 

So reminder for everyone: please always use string substitution:
```java
logger.debug("The value of var is {}", var);
```
in favor of concatenation:
```java
logger.debug("The value of var is " + var);
```

Reference: https://www.slf4j.org/faq.html#logging_performance

sbliven:
Should we search systematically for string concatenations or String.format calls in logging statements?

lee-mcfaul:
Sonarqube is a pretty good tool for finding such things as well as a plethora of other coding bugs, smells and other faux-pas. It's a little depressing running it on a codebase for the first time, but overall it's a pretty good tool to keep code clean and use best practices.

 It could be set up similar to the Travis build where code bellow a certain quality is blocked from merging. 

https://www.sonarqube.org/

josemduarte:
Yes, Sonarqube is indeed a great tool. We are actually trying it right now for other projects. I'm going to give SonarCloud a try in BioJava as soon as I can.

josemduarte:
Closing. I'm trying to get sonarqube working in CI #924

PR #1086, opened 2024-03-05
Fix sonar issue 2629 Logging arguments should not require evaluation

jlerbsc:
Some method calls can effectively be "no-ops", meaning that the invoked method does nothing, based on the application’s configuration (eg: debug logs in production). However, even if the method effectively does nothing, its arguments may still need to evaluated before the method is called. This fix improves performance, especially in cases where concatenation is used at low logging levels.


josemduarte (approved):
Very nice indeed, thank you! And this completely solves issue #789

PR #1088, opened 2024-03-29
Fix: refactor implementation and design smells

cassandrabeaton:
This is a fix for several code smells discovered in the code.
1. Complex method in TranscriptSequence (Fix: Extracted helper method and removed duplication)
2. Complex conditional in AtomPositionMap (Fix: Decomposed conditionals)
3. Long statement in GroupContactSet (Fix: Added explaining variables)
4. Insufficient modularization in SubunitCluster (Fix: Extracted a new class)
5. Feature envy in StructureInterface (Fix: moved method to Pair class)
6. Cyclically-dependent modularization in PdbPairXMLConverter (Fix: changed dependency to unidirectional)

Tests are passing for all fixes. Test class TestSubunitClusterMerge was created for the new extracted class mentioned in #4.


biojava-structure/src/main/java/org/biojava/nbio/structure/contact/Pair.java
```diff
@@ -97,4 +97,14 @@ public boolean equals(Object obj) {
 	public String toString() {
 		return "<" + first.toString() + ", " + second.toString() + ">";
 	}
+
+	/**
+	 * Tells whether the interface corresponds to one mediated by crystallographic symmetry,
+	 * i.e. it is between symmetry-related molecules (with same chain identifier)
+	 * @return
+	 */
+	public boolean isSymRelated() {
```
josemduarte, line 106: The Pair class is general and intended for any kind of pair (note the generic type `T`). This would break that generality.

josemduarte (changes requested):
Thanks. Please see some comments below

PR #1091, opened 2024-04-18
Fix sonar issue S2293 Replace the type specification in this constructor call with the diamond operator ('<>')

jlerbsc:
Java uses angular brackets (< and >) to provide a specific type (the "type argument") to a generic type. For instance, List is a generic type, so a list containing strings can be declared with List<String>.

Indepth, our Java code remediation solution, has detected and corrected 1103 violations of this type in the current version of the Biojava project. Manually, this correction would have taken more than 2 days.

We suggest that you correct this to improve the quality and maintainability of the code.

### DIFF
--- a/biojava-aa-prop/src/main/java/org/biojava/nbio/aaproperties/CommandPrompt.java
+++ b/biojava-aa-prop/src/main/java/org/biojava/nbio/aaproperties/CommandPrompt.java
@@ -95,8 +95,8 @@ public class CommandPrompt {
 		/*
 		 * Parse input arguments
 		 */
-		List<Character> propertyList = new ArrayList<Character>();
-		List<Character> specificList = new ArrayList<Character>();
+		List<Character> propertyList = new ArrayList<>();
+		List<Character> specificList = new ArrayList<>();
 		String inputLocation = null;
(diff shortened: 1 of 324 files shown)


josemduarte:
> I'm not sure that these tests (in class EcodInstallationTest) fail because of this PR.

Yes this is due to an external resource (ECOD) having downtime for the last week or so. I have notified them and I'm hoping for it to come back soon.

> I can't check this because the tests fail on my Windows machine with the error
[ERROR] Errors:
[ERROR] FastaStreamerTest.iterate:43 " InvalidPath Illegal char <:> at index 2: /D:/de...
[ERROR] FastaStreamerTest.stream:21 " InvalidPath Illegal char <:> at index 2: /D:/dev...

That sounds like a bug in that test. Thanks for reporting. Can't test it myself directly because I don't have windows at hand. Could you create an issue?

jlerbsc:
> Yes this is due to an external resource (ECOD) having downtime for the last week or so. I have notified them and I'm hoping for it to come back soon.

Should I do anything to restart the build?

josemduarte:
Ok tests are passing now that ECOD is back

Issue #1092, opened 2024-04-18
Biojava core tests fail on Windows


jlerbsc:
Tests fail with the following errors

    [ERROR] Errors:
    [ERROR] FastaStreamerTest.iterate:43 " InvalidPath Illegal char <:> at index 2: /D:/de...
    [ERROR] FastaStreamerTest.stream:21 " InvalidPath Illegal char <:> at index 2: /D:/dev...

josemduarte:
@hilbertglm I think this came in your recent PR: #1080 . Could you have a look?

josemduarte:
In order to prevent this in the future, I've added windows to the OS list to test in github workflows for Pull Requests. See 4fcb3b1b

PR #1093, opened 2024-04-24
Fix FastaStreamerTest tests failing on windows due to URL handling

cometbroom:
Issue #1092 :
When getting a resource with the, getFile() method. It does not contain the scheme of the resource you're trying to access. That is needed by Paths.get() to create a path appropriate for Windows file system. 
The url returned by getFile() would be "C:/Users/etc" while it needs to be "file:/C:/Users/etc"

### DIFF
--- a/biojava-aa-prop/pom.xml
+++ b/biojava-aa-prop/pom.xml
@@ -2,7 +2,7 @@
 	<parent>
 		<artifactId>biojava</artifactId>
 		<groupId>org.biojava</groupId>
-		<version>7.1.2-SNAPSHOT</version>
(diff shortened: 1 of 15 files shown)


josemduarte (approved):
Thanks, LGTM. Did you manage to test this in windows?

One thing only: could you revert the changes to the pom files? Those shouldn't be needed. We'll increment versions at next release automatically.

cometbroom:
@josemduarte Yes all tests passed successfully on windows. 

Also reverted changes to pom files. My bad, missed that part.

PR #1094, opened 2024-04-25
Upgrading to slf4j2

josemduarte:
As subject says


JonStargaryen (approved):
LGTM, thanks.

Because it's nearby in the code: Do we have a plan for the MMTF deprecation? The codec is still useful once mmtf.rcsb.org stops serving data but could also get rid of it.

josemduarte:
> Because it's nearby in the code: Do we have a plan for the MMTF deprecation? The codec is still useful once mmtf.rcsb.org stops serving data but could also get rid of it.

Yes the codec is still useful, so in principle it is a good idea to keep for a while longer. I would not want it to stay too long though, because there's not much maintenance on the mmtf codec java libs.

PR #1095, opened 2024-05-15
Fix issue s1319 Declarations should use Java collection interfaces on  biojava-core

jlerbsc:
This PR is an experiment to fix the rule Sonar s1319 "Declarations should use Java collection interfaces such as 'List' rather than specific implementation classes such as 'LinkedList'".

It's a bit complicated on the Biojava project because the Sonar rule raises a few false positives and the project makes extensive use of interface implementations and sometimes methods that are implementation-specific. But after a few manual adjustments to standardise the modifications in the dependent modules, the tests pass with the exception of the TestCrystallographicMetadata.test1zna:92 test in the biojava-integrationtest module. I'm not convinced that the error stems from these modifications, but I'll let you be the judge.

Below is a link to the sonar page describing the problem https://rules.sonarsource.com/java/tag/bad-practice/RSPEC-1319/.

Indepth has fixed 102 issues that would have taken about 2 days to correct manually.

### DIFF
--- a/biojava-alignment/src/test/java/org/biojava/nbio/alignment/TestDNAAlignment.java
+++ b/biojava-alignment/src/test/java/org/biojava/nbio/alignment/TestDNAAlignment.java
@@ -43,6 +43,7 @@ import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 
 public class TestDNAAlignment {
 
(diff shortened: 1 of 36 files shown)


jlerbsc:
There sometimes seem to be problems with the BerkeleyScopInstallationTest.

josemduarte (approved):
LGTM thank you!

I've triggered the tests task that failed. But 3 of them have succeeded so I guess the `TestCrystallographicMetadata.test1zna` is working fine.

jlerbsc:
Does this mean that you are going to merge this PR? If so, I'll try to continue resolving this issue.

josemduarte:
Yes I'd be happy to merge this. Will you do one PR per module then?

jlerbsc:
I don't know yet... Above all, I'm going to try to have a stable situation with each PR. It's likely that this will lead me to do one RP per module.

jlerbsc:
I suggest you do one PR per module. What do you think?

PR #1099, opened 2024-07-22
Removing functionality that uses mmtf.rcsb.org, now deprecated

josemduarte:
This removes any functionality that uses files from the mmtf.rcsb.org API, now deprecated by RCSB PDB.

It is still possible to read/write MMTF files from elsewhere.

Also: demoted logging level when content-length size can't be retrieved (happens for bcif retrieval from models.rcsb.org)


JonStargaryen (approved):
LGTM, thanks for the cleanup.
Could also consider serving content-lengths for models.rcsb.org. Seems like a bug that that's missing.

PR #1101, opened 2024-08-30
Fix CE-Symm bug with short repeats

sbliven:
This fixes rcsb/symmetry#118.

It switches the determination of the symmetry group and rotation axes to use the CESymm alignment directly. Before it was recomputed using QuatSymmDetector, which failed with short repeats.

A new CE-Symm version will be released after the next biojava release.


heuermh:
Thank you, @sbliven!

PR #1103, opened 2025-01-06
More lenient CIF parsing

josemduarte:
mmCIF files generated by packages like Gemmi don't have all the fields that archived PDB files do. This makes the parser more lenient to some missing fields. See https://github.com/rcsb/rcsb-api-tools/issues/8


sbittrich (approved):
Thanks for investigating, LGTM.

PR #1105, opened 2025-02-13
Update guava dependency version to 33.4.0-jre

heuermh:


heuermh:
Thank you, @josemduarte

PR #1106, opened 2025-02-21
More on lenient PDBx/mmCIF parsing

josemduarte:
Following #1103 . This covers some fields that are missing in [PDB-IHM](https://pdb-ihm.org/) entries.


sbittrich (approved):
LGTM, thanks.

PR #1107, opened 2025-02-21
ASA calculation should be done per model

josemduarte:
With this the default `AsaCalculator` constructor will use model 1 only. Also introduced a new constructor to choose the model.


biojava-structure/src/main/java/org/biojava/nbio/structure/asa/AsaCalculator.java
```diff
@@ -135,7 +165,7 @@ static class IndexAndDistance {
 	 * NACCESS' -h option
 	 */
 	public AsaCalculator(Structure structure, double probe, int nSpherePoints, int nThreads, boolean hetAtoms) {
-		this.atoms = StructureTools.getAllNonHAtomArray(structure, hetAtoms);
+		this.atoms = StructureTools.getAllNonHAtomArray(structure, hetAtoms, 1);
```
sbittrich, line 168: Delegating to the new constructor makes this easier to maintain.

```suggestion
		this(structure, probe, nSpherePoints, nThreads, hetAtoms, 1);
```

sbittrich (approved):
Looks good. Suggest to remove the duplicated code though.

PR #1109, opened 2025-03-18
CIF parsing: entity type should be case insensitive

josemduarte:
Some PDB-IHM files use upper case, unlike PDB files that use lower case.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/EntityType.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/EntityType.java
@@ -106,7 +106,7 @@ public enum EntityType {
 			return null;
 
 		for(EntityType et : EntityType.values()) {
-			if (entityType.toLowerCase().equals(et.entityType)) {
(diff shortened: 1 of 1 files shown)


biojava-structure/src/main/java/org/biojava/nbio/structure/EntityType.java
```diff
@@ -101,19 +98,15 @@ private void setEntityType(String entityType) {
...
+	public static EntityType entityTypeFromString(String entityType) {
+		if (entityType == null)
 			return null;
 
-		for(EntityType et : EntityType.values())
-		{
-			if(entityType.equals(et.entityType))
-			{
+		for(EntityType et : EntityType.values()) {
+			if (entityType.toLowerCase().equals(et.entityType)) {
```
sbittrich, line 109:
```suggestion
			if (entityType.equalsIgnoreCase(et.entityType)) {
```

sbittrich (approved):
LGTM, thanks.

I would go with `equalsIgnoreCase` directly. Is the change in visiblity for the constructor intended?

PR #1110, opened 2025-03-18
Upgrade to latest ciftools-java version

josemduarte:
Should fix performance for some edge cases, see https://github.com/rcsb/ciftools-java/issues/13


sbittrich (approved):
LGTM, thanks.

PR #1125, opened 2026-05-27
Issue 1038 - Test was refactored by using junit5 features

arkanar611:
Issue 1038 - Test was refactored by using junit5 features. 
The next step should be converting other tests in the module to JUnit5


josemduarte:
In principle this looks good, thank you. However we currently have an issue with tests (due to a failing external resource used during testing) that we'd like to fix before going with this junit5 refactoring.

josemduarte (approved):
LGTM thanks. Tried the tests locally and they worked.

PR #1126, opened 2026-05-27
Biojava genome migration to junit5

arkanar611:
All tests in the biojava-genome module were migrated to junit5

### DIFF
--- a/biojava-genome/pom.xml
+++ b/biojava-genome/pom.xml
@@ -82,22 +82,6 @@
 			<version>7.2.7-SNAPSHOT</version>
 			<scope>compile</scope>
 		</dependency>
-		<dependency>
-			<groupId>junit-addons</groupId>
-			<artifactId>junit-addons</artifactId>
-			<version>1.4</version>
-			<scope>test</scope>
-			<exclusions>
-				<exclusion>
-					<groupId>xerces</groupId>
-					<artifactId>xmlParserAPIs</artifactId>
-				</exclusion>
-				<exclusion>
-					<groupId>xerces</groupId>
-					<artifactId>xercesImpl</artifactId>
-				</exclusion>
-			</exclusions>
-		</dependency>
 		<!-- logging dependencies (managed by parent pom, don't set versions or scopes here) -->
 		<dependency>
         	<groupId>org.slf4j</groupId>
--- a/biojava-genome/pom.xml
+++ b/biojava-genome/pom.xml
@@ -68,6 +68,18 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-engine</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-params</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.junit.vintage</groupId>
+			<artifactId>junit-vintage-engine</artifactId>
+		</dependency>
+		<dependency>
 			<groupId>org.biojava</groupId>
 			<artifactId>biojava-core</artifactId>
 			<version>7.2.6-SNAPSHOT</version>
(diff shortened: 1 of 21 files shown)


josemduarte:
In principle this looks good, thank you. However we currently have an issue with tests (due to a failing external resource used during testing) that we'd like to fix before going with this junit5 refactoring.

biojava-genome/src/test/java/org/biojava/nbio/genome/TestChromosomeMappingTools.java
josemduarte: Unused now, should be removed

biojava-genome/src/test/java/org/biojava/nbio/genome/io/fastq/FastqTest.java
```diff
@@ -21,25 +21,26 @@
 package org.biojava.nbio.genome.io.fastq;
 
 import org.junit.Assert;
```
josemduarte: Import from junit4. Could you move the import and corresponding usages below. AI tells me to use `Assertions.assertThrows(X.class, () -> ...)`

biojava-genome/src/test/java/org/biojava/nbio/genome/GeneFeatureHelperTest.java
josemduarte: Could you remove this import and then replace the usages below. AI recommends:
`FileAssert.assertEquals() -> assertIterableEquals(Files.readAllLines(), Files.readAllLines(), "the text")`

After that you can get rid of the old dependency in pom.xml

biojava-genome/pom.xml
josemduarte: You can remove this dependency once the comment above is solved (see my other comment). For instance, this is important because this dep brings in the old junit4 transitively

josemduarte (changes requested):
Thank you, see some comments below.

josemduarte:
Also, please could you merge master into your branch. Then tests should be able to run cleanly.

arkanar611:
@josemduarte Thank you for reviewing my MR. I've removed junit-addons from dependencies and merged master branch changes but there are still some issues with biojava-protein-disorder tests https://github.com/biojava/biojava/actions/runs/33166795706/job/98834028902?pr=1126

biojava-genome/src/test/java/org/biojava/nbio/genome/GeneFeatureHelperTest.java
```diff
@@ -93,15 +93,17 @@ public void testLoadFastaAddGeneFeaturesFromUpperCaseExonFastaFile() throws Exce
...
 
 		File fastaSequenceFile = new File("src/test/resources/volvox_all.fna");
 		File gffFile = Files.createTempFile("volvox_length","gff3").toFile();
 		gffFile.deleteOnExit();
 		GeneFeatureHelper.outputFastaSequenceLengthGFF3(fastaSequenceFile, gffFile);
-		FileAssert.assertEquals("volvox_length.gff3 and volvox_length_output.gff3 are not equal", gffFile,
-				new File("src/test/resources/volvox_length_reference.gff3"));
+		Assertions.assertEquals(
+				Files.readString(new File("src/test/resources/volvox_length_reference.gff3").toPath()),
+				Files.readString(gffFile.toPath()),
```
josemduarte, line 105: AI tells me that these 2 should be Files.readAllLines instead. Or otherwise tests will fail in windows (due to the hard-coded `\n` in https://github.com/biojava/biojava/blob/d6a960419d3690e472a6367aaec65952f4514a23/biojava-genome/src/main/java/org/biojava/nbio/genome/parsers/gff/GFF3Writer.java#L46 )

josemduarte (changes requested):
Thank you for the changes, one more comment below.

Note that the failures were due to a UniProt server being down.

biojava-genome/src/test/java/org/biojava/nbio/genome/GeneFeatureHelperTest.java
josemduarte: Ok never mind: AI was wrong. It worked in CI. Though we are still seeing flakiness in UniProt servers

josemduarte (approved):
LGTM thanks

Issue #980, opened 2021-10-11
Detect incomplete or corrupted downloaded files


aalhossary:
- It is quite often when a file is being downloaded that it downloads incomplete or corrupted; which could lead to unexpected potential problems.
 
- I propose to create one (_or two_) files beside the target file **target.ext**, named **target.ext.size** (and **target.ext.sha**).
- When we need to parse a file, we currently check its presence (e.g. in `ScopInstallation.ensureClaInstalled()` for SCPR classification DB file). In my proposal, we do not only check for its presence, but also for its size +/- its hash code.
 

1. **HashCode**: Some web folders publish the HashCode along with the download file itself. If not, we ignore the hash code :(.
2. **Size**: We can issue a `HEADER` request before the `GET` or `POST` call that downloads the file itself, extract the size, and save it.
 
- The four functions to **store/check size/hash** can be implemented centrally in a separate location (most probably in `BioJava-core`) as public static methods and called only at the respective use places.
 
- This way, we _hopefully_ will not modify in the current code a lot: just adding one (or two) lines to call `storeSize()`, and `storeHash()` at the file download location and `checkSize()` +/- `checkHash()` method from within the `ensureXXXInstalled()` functions.

Sounak123:
> Sample code in case someone wants to work on it.
> 
> ```java
> URL url;
> HttpURLConnection httpConnection = null;
> try {
> 	url = new URL("https://scop.berkeley.edu/downloads/parse/dir.des.scope.2.07-stable.txt");
> 	URLConnection connection = url.openConnection();
> 	if (connection instanceof HttpURLConnection) {
> 		httpConnection = (HttpURLConnection) connection;
> 	}
> 	System.out.println("Content-Length: " + connection.getContentLengthLong());
> //			if (httpConnection != null) {
> //				System.out.println(httpConnection.getResponseCode() + " " + httpConnection.getResponseMessage());
> //				httpConnection.disconnect();
> //			}
> } catch (MalformedURLException e) {
> 	e.printStackTrace();
> } catch (IOException e) {
> 	e.printStackTrace();
> }
> ```

Hi, Could you please provide me a brief overview on where the change is required? I would like to contribute!

aalhossary:
Hi @Sounak123, sorry for my late reply.
We always welcome new contributors.

You will need to create 2 abstract methods somewhere in biojava-core

- `createValidationFiles(URLConnection resource, File localDestination, URLConnection sourceHassh)`
[...]

Issue #979, opened 2021-10-10
BerkeleyScopInstallationTest.traverseHierarchy() needs update


aalhossary:
Version 2.08 fails this test.
I commented out line 50 (`ScopFactory.LATEST_VERSION,`) to work around it locally. But we need to fix it please.

aalhossary:
@sbliven Can you fix this please?

aalhossary:
Thank you for your reply @josemduarte. 
After double checking, I found now that my dir.cla.scop.txt_2.08 file was incompletely downloaded (20 MB instead of 33 MB).
When I deleted and redownloaded it, everything worked fine.

Sorry for the false alarm. I will close the issue now.

PR #1133, opened 2026-08-15
Fix FileDownloadUtils validation and implement checksum verification

### LINKED ISSUE #1138: CathInstallation downloads over http, which now redirects to https, and the redirect body is cached as classification data
Sub-issue of #1135, cause 1 of 3.

## The bug

`CathInstallation.java:57`:

```java
public static final String CATH_DOWNLOAD_URL = "http://download.cathdb.info/cath/releases/";
```

That host now redirects:

```
$ curl -sI http://download.cathdb.info/cath/releases/
HTTP/1.1 301 Moved Permanently
Location: https://download.cathdb.info/cath/releases/

$ curl -sI https://download.cathdb.info/cath/releases/
HTTP/1.1 200 OK
```

`HttpURLConnection` follows redirects within a protocol but deliberately will not follow one that switches `http` to `https`. `downloadFileFromRemote` read the response with a bare `remoteURL.openStream()` and no status check, so the body of the 301 was written into the local cache as though it were classification data. Parsing then produced no domains, and the first symptom appeared much later and far from the cause:

```
CathDomainTest.test:40  NullPointer
  Cannot invoke "org.biojava.nbio.structure.cath.CathDomain.toCanonical()" because "domain" is null
```

## Why the redirect matters more than the URL

Switching to `https` alone turns the test green, and the upstream change is out of our hands. Accepting a redirect body as data is not. Any download that reads a stream without checking the status can cache an error page or a redirect body that looks like a valid file forever after — and because it is cached, it keeps failing after the service recovers.

## Fix

Fixed by commit `ed36d3a` in #1133, which changes the URL to `https` **and** routes the download through the status-checking `FileDownloadUtils.downloadFileWithValidation`. #1133 closes the same hole in `DownloadChemCompProvider`, which is not broken today only because the RCSB ligand endpoint still serves plain http.

Verified in CI: before that commit all five matrix jobs failed on `CathDomainTest`; after it, none do (run `31916362199`).

```bash
mvn verify -pl biojava-integrationtest -Dtest=CathDomainTest
```

aalhossary:
The download-validation helpers added in 7.0.0 (#979, #980) have several gaps that only surface once a caller passes a real hash URL, or downloads from a server that can return 404. This fixes all of them and implements hash verification rather than leaving it stubbed.

### Correctness fixes

- `createValidationFiles(URL, ...)` passed a literal `Hash.UNKNOWN` to its `URLConnection` overload instead of the caller's argument, so it could never write a hash file and threw `IllegalArgumentException` for any caller that supplied a
 `hashURL`.
- Neither `downloadFile` nor `createValidationFiles` checked the HTTP status, so a 404 error page was written into the cache as though it were the requested file. Because the `.size` sidecar was then taken from that same error response, `validateFile` subsequently declared it **valid**. A new `HttpStatusException` lets callers tell "the resource is not there" apart from a transport failure, which matters for anything that tries several mirrors in turn.
- `downloadFile` used `FileChannel.transferFrom(rbc, 0, Long.MAX_VALUE)`, which is not guaranteed to drain a socket-backed channel and could silently truncate a download. Replaced with `Files.copy`, which loops to end of stream.
- `downloadFile` leaked its temporary file on every failure path.
- `validateFile` threw `NullPointerException` for a file with no parent directory, and again if `listFiles()` returned null; an empty `.size` file raised an unchecked `NoSuchElementException` that escaped the surrounding catch.
- `validateFile` checked only the first hash sidecar it found, ignoring the rest.

### New functionality

- `validateFile` now really verifies MD5, SHA-1 and SHA-256. Sidecars are written as bare lowercase hex and parsed tolerantly, so a file downloaded verbatim from a server in coreutils or BSD layout is also understood. A sidecar that cannot be parsed is skipped with a warning rather than failing an otherwise good download.
- `ETagPolicy` lets an `ETag` that is a bare hex digest be recorded as a checksum without a second request. **`files.wwpdb.org` and `files.rcsb.org` return the content MD5 as the `ETag`** — verified byte-exact — so every download from the wwPDB archive now gets a real integrity check for free. The `<mtime>-<size>` ETags used by the EBI servers contain a dash and can never be misread as a digest; a test pins that.
- `downloadFileWithValidation` downloads and validates over a single connection. The previous pattern opened one connection to read `Content-Length` and another to fetch the bytes, so the recorded size described a different response than the one written; if the resource changed in between, the cache entry was left permanently failing validation.
`LocalPDBDirectory` uses the new single-connection download, and its two-character directory hash is promoted to a reusable `getMiddleHash(String)`.

### Behaviour change

The existing four-argument `createValidationFiles` overloads now default to `ETagPolicy.USE_IF_HEX_DIGEST`, so callers start recording checksums where the server offers one.
Actually the work is targeted at 7.3.0 rather than a patch release (check next PR). Therefore This small behavioural change should be acceptable.

### Consumers fixed here

Fixes #1138 — `CathInstallation` downloaded over `http`, which now 301s to `https`; `HttpURLConnection` will not follow a redirect that changes protocol, so the body of the 301 was cached as classification data. Commit `ed36d3a` switches the URL and routes the download through the status-checking path added here. Commits `5ec48ac` and `5d07241` close the same hole in `DownloadChemCompProvider`, which is not broken today only because the RCSB ligand endpoint still serves plain http.

### Testing

37 tests in `FileDownloadUtilsTest`, including the live wwPDB ETag-to-MD5 round trip and a 404 that must leave neither a cached file nor a `.size` sidecar behind. Full `biojava-core` suite: 522 tests, no failures.

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/util/FileDownloadUtils.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/util/FileDownloadUtils.java
@@ -21,9 +21,9 @@
  */
 package org.biojava.nbio.core.util;
 
+import java.io.BufferedInputStream;
 import java.io.File;
 import java.io.FileNotFoundException;
-import java.io.FileOutputStream;
 import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStream;
@@ -32,11 +32,15 @@ import java.net.HttpURLConnection;
 import java.net.SocketTimeoutException;
 import java.net.URL;
 import java.net.URLConnection;
-import java.nio.channels.Channels;
-import java.nio.channels.ReadableByteChannel;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.*;
 import java.nio.file.attribute.BasicFileAttributes;
+import java.security.DigestInputStream;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.Scanner;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -47,8 +51,43 @@ public class FileDownloadUtils {
 	private static final String HASH_EXT = ".hash";
 	private static final Logger logger = LoggerFactory.getLogger(FileDownloadUtils.class);
 
+	/** Buffer used when streaming a file through a {@link MessageDigest}. */
+	private static final int DIGEST_BUFFER_SIZE = 64 * 1024;
+
+	/** A bare hex digest, optionally followed by whitespace and a file name (the
+	 * layout written by <code>md5sum</code>, <code>sha1sum</code> and friends). */
+	private static final Pattern BARE_HEX_HASH = Pattern.compile("^([0-9a-fA-F]{32,128})(?:[\\s*].*)?$");
+
+	/** The BSD layout, e.g. <code>MD5 (file.txt) = d41d8cd9...</code>. */
+	private static final Pattern BSD_HASH = Pattern.compile("^\\w+\\s*\\(.*\\)\\s*=\\s*([0-9a-fA-F]{32,128})$");
+
 	public enum Hash{
 		MD5, SHA1, SHA256, UNKNOWN
+	}
+
+	/**
+	 * What to do with the <code>ETag</code> response header when no explicit hash
+	 * URL is available.
+	 * <p>
+	 * Some archives &mdash; notably <code>files.wwpdb.org</code> and
+	 * <code>files.rcsb.org</code> &mdash; return the MD5 digest of the content as
+	 * the bare <code>ETag</code> value, which lets us record a real checksum
+	 * without a second request. Others (the EBI servers, for instance) return a
+	 * <code>&lt;modification-time&gt;-&lt;size&gt;</code> form instead; because that
+	 * always contains a <code>-</code>, it can never be mistaken for a hex digest.
+	 *
+	 * @author Amr ALHOSSARY
+	 * @since 7.3.0
+	 */
+	public enum ETagPolicy {
+		/** Never look at the <code>ETag</code> header. */
+		IGNORE,
+		/** Record the <code>ETag</code> as a checksum when it is a bare hex digest
+		 * of a length matching one of the supported algorithms. */
+		USE_IF_HEX_DIGEST,
+		/** As {@link #USE_IF_HEX_DIGEST}, but log a warning when the header is
+		 * missing or is not a usable digest. */
+		REQUIRE
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


aalhossary:
**Reviewer note:** This PR should (when approved) should be merged before #1134 not after it; because this PR is a prerequisite and dependency of the other PR.

aalhossary:
Added a commit here that fixes **cause 1 of the 3** listed in #1135: `ed36d3a` — *Fix CATH downloads: use https, and check the response before caching it*.

It belongs in this PR rather than a separate one, because it is the first real consumer of the hardening added here.

### What it fixes

`download.cathdb.info` now answers plain http with a 301 to https. `HttpURLConnection` follows redirects within a protocol but deliberately will not follow one that switches http to https, so our download never reached the real file. `downloadFileFromRemote` read the response with a bare `openStream()` and no status check, so the body of the 301 was written into the cache as though it were classification data. Parsing produced no domains, and the first symptom appeared much later and far from the cause:

```
CathDomainTest.test:40 NullPointer
  Cannot invoke "CathDomain.toCanonical()" because "domain" is null
```

The URL change alone turns the test green. The reason the commit also swaps the hand-rolled copy loop for `downloadFileWithValidation` is that the URL was only the trigger — accepting a redirect body as data is the actual defect, and without the status check the next service that changes a redirect reproduces the same silent corruption and the same puzzling NPE.

### Scope, to be clear

| # | Cause in #1135 | Status |
|---|---|---|
| **1** | **CATH downloader follows a stale `http://` URL that now 301s (#1138)** | **fixed by `ed36d3a`** |
| 2 | ECOD changed its distribution format (#1139) | not addressed here; fixed by #1141 |
| 3 | `Master Build` fails on an expired/absent `SONAR_TOKEN` (#1140) | not addressed; needs someone with the SonarCloud credentials |

So this does not turn CI green on its own — it removes one of the two test failures that currently fail every pull request, and fixes a genuine runtime bug that affects anyone calling `CathInstallation` today.

### Verification

`CathDomainTest` fails on master and passes here, in about 11 seconds:

```
mvn verify -pl biojava-integrationtest -Dtest=CathDomainTest
```



---

*Edited to point at the sub-issues, and to correct row 2: the ECOD failure was not a judgement call about download size but a format change upstream, now fixed by #1141.*

aalhossary:
Two more commits, applying the same guard to the chemical component download:

| Commit | |
|---|---|
| `5ec48ac` | Reject redirects and error responses in the chem comp download |
| `5d07241` | Test that a redirect or error is never cached as a chem comp definition |

### Why this one needed it too

`downloadChemCompRecord` read the response body without ever checking the status. A 4xx or 5xx already failed safely, because `getInputStream()` throws for those — which is why the existing `testWeDontCacheGarbage` (#703) passes and never revealed anything. A redirect is different: when the JDK declines to follow a 3xx, and it always declines when the redirect changes http to https, `getInputStream()` returns the body of the redirect instead. That body is short but not empty, so the existing "did we read any lines" check accepted it, gzipped it, and stored it under the component's name. Every later lookup then read that back and failed to parse it, far from the request that caused it.

That is precisely how CATH broke in `ed36d3a`, and `files.rcsb.org` would do the same to us the day it stops serving plain http. Chemical components are fetched constantly and cached indefinitely, so a poisoned entry there is worth more than the two lines it takes to prevent.

### What was deliberately not changed

The download re-compresses the response as it writes, so the local `.cif.gz` is not a byte copy of the remote `.cif`. `downloadFileWithValidation` writes bytes verbatim and cannot stand in for it without changing the on-disk format, so this commit adds only the status check.

### On the test

It serves its responses from a local `HttpServer` and needs no network. Pointing a test at a third party that happens to redirect today would make it fail the day they stop — the coupling that has kept this build red since December. Three cases: a 301 that changes protocol, a 503, and a 200 to confirm the guard rejects bad responses rather than refusing to download at all.

```
mvn test -pl biojava-structure -Dtest='TestChemCompRedirectNotCached,TestDownloadChemCompProvider'
```

8 tests, green, offline.

### Where this PR now stands

`FileDownloadUtils` hardening, plus its first two consumers: CATH (`ed36d3a`, which fixes cause 1 of 3 in #1135) and chem comps (`5ec48ac`, `5d07241`, pre-emptive — nothing is broken there today).

aalhossary:
**Reviewer note:** 
I propose handing this bunch of PRs in this order: 

1. #1133 (this PR)
2. #1141 (depends on it)
3. #1134 It depends on the current PR.


These 2 are up to your decision and can be done in any order
- #1137
- #1142

josemduarte (approved):
LGTM thanks

I'll go ahead and merge, since this contain the important fix to CATH url that is making one of the integration tests to fail.

PR #1134, opened 2026-08-15
Fetch, cache and display electron density and cryo-EM maps

### LINKED ISSUE #947: Fetch, cache, load, and view electron density maps
The validation files are provided from **PDBe**, in [ftp.ebi.ac.uk/pub/databases/pdb/validation_reports/](http://ftp.ebi.ac.uk/pub/databases/pdb/validation_reports/).

We need to establish a 1ry mechanism to fetch and cache them as we do with atomfiles.
Then to load them into Jmol when needed.

aalhossary:
Closes #947.

This PR is about a new feature, therefore, I recommend releasing it as BioJava **7.3.0** rather than **6.2.7**.
It adds `org.biojava.nbio.structure.io.density`, which downloads and caches density maps the way `LocalPDBDirectory` caches coordinate files, and a `JmolPanel` method that contours the result.

```java
DensityMapCache cache = new DensityMapCache();
DensityMapResult map = cache.getDensityMap(new PdbId("1cbs"), DensityMapKind.TWO_FO_FC);
jmolPanel.loadDensityMap(map);
```

### Why a chain of sources rather than one

`edmaps.rcsb.org` shut down in October 2024. What RCSB documents in its place —the map coefficients published with the wwPDB validation reports — are structure factors, not a sampled grid, and cannot be displayed without a Fourier transform.
Several other services do serve grids, and they differ enormously in size for the same entry, so rather than picking one, sources are tried in order until one answers:

| | order |
|---|---|
| X-ray | RCSB density server → PDBe CCP4 → PDBe density server → wwPDB coefficients *(off by default)* |
| cryo-EM | RCSB density server → PDBe density server → EMDB primary map |

Measured sizes for the same data:

| entry | full archive | density server |
|---|---|---|
| 1cbs (X-ray) | 2.1 MB (two CCP4 files) | 210 kB at detail 0 |
| EMD-0262 (6hu9) | 116 MB | 0.5–3.7 MB |

wwPDB coefficients are supported for completeness but disabled by default, and `DensityFileFormat.isJmolLoadable()` lets a viewer refuse them rather than silently drawing nothing.

### Design notes

- A source that has nothing for an entry is skipped; **any other transport failure aborts the chain**. A network outage must never be reported as "this entry has no density". When every source is exhausted, `NoDensityMapException` carries the reason from each one, so the UI can explain rather than just fail.
- A density server response contains both the 2Fo-Fc and Fo-Fc blocks, so the two kinds share one cache entry instead of downloading the identical file twice.
- Cryo-EM entries are resolved through EMDB's search API, with RCSB as a fallback. That lookup also yields the depositors' recommended contour level, which is how an EM map should be contoured; it is attached to the result whichever source supplied the voxels. The experimental method is never inferred from resolution, which BioJava parses incorrectly for some cryo-EM entries (#1000).
- `Ccp4Header` checks for the `MAP ` stamp at byte 208, so a server answering with an error page and HTTP 200 produces a clean cache miss rather than a corrupt entry.
- A `.meta` sidecar fully describes each cached result, so `LOCAL_ONLY` is served without opening a connection.

### Three Jmol behaviours established by experiment

Each of these changed the implementation, and none is documented:

1. **Option order in `isosurface` is load-bearing.** With `mesh nofill` before the file name, Jmol accepts the command, reports no error, and draws nothing at all.
2. **A negative sigma does not contour at a negative level.** Jmol reserves negative sigma internally, so `sigma -3.0` silently contours at the default level — the intended red lobe of a difference map came out identical to the blue one. Difference maps are drawn as a single signed surface instead.
3. **Selecting the Fo-Fc block of a cached BinaryCIF file needs the marker in the file *name*.** `BCifDensityReader` picks the block by testing whether the file name contains `&diff=1`, which normally arrives in the URL query string. A cached local file has no query string; appending the marker to the file URL fails, and the `#diff=1` form the reader's own to-do comment suggests is ignored. Embedding it in the name works, so the cache exposes the difference map under a companion name, hard-linked to the same bytes.

Measured proof of (3): the plain file contours at cutoff 0.356 over a −1.31 to 3.78 range; the companion name contours at 0.374 over −0.69 to 0.85. Different data, so the marker genuinely selects the other block.

### Testing

43 unit tests, entirely offline — the chain tests use stub providers, so "404 falls through" and "anything else aborts" are pinned exactly. 6 integration tests against the live services costing a couple of megabytes in total; the cryo-EM path is covered without downloading the 116 MB map by setting the size limit so the guard fires after resolution but before transfer.

### Jmol version

No change: BioJava stays on `net.sourceforge.jmol:jmol:14.31.10`, which is the newest on Maven Central (October 2020). Every behaviour relied on here — `MrcBinaryReader`, `BCifDensityReader`, the `&diff=1` selection, gzip sniffing — is identical in 14.31.10 and current Jmol, so downstream projects overriding the dependency with a newer build are unaffected.


aalhossary:
**Worth discussing: how small is small enough?**

This PR deliberately prefers the smallest adequate representation, and I would like a second opinion on where the line should sit.

The density servers return a downsampled grid, and the saving is not marginal. For 1cbs, detail 0 is 210 kB against 2.1 MB for the two full CCP4 files — and one request instead of two, since a single response carries both map kinds. For the cryo-EM map behind 6hu9 it is 0.5–3.7 MB against 116 MB. The current defaults are detail 3 for RCSB and detail 6 for PDBe, matching what each service's own viewer asks for.

Open questions:

1. **Is a downsampled grid adequate for what people actually use BioJava density for?** 
For looking at a ligand pocket in a viewer I am confident it is. For anything quantitative — real-space correlation, occupancy refinement, automated density fit scoring — it may well not be, and someone doing that would want the full grid and would not necessarily notice they had been handed a coarser one. The result object does report its source, so the information is available, but "available if you check" is weaker than a sensible default.

3. **Should the default detail level be lower?**
Detail 0 for 1cbs is a tenth the size of detail 3 and still perfectly readable around a ligand. I kept the services' own defaults rather than choosing for them, but I am not attached to that.

5. **Should the size ceiling be the primary mechanism instead of the ordering?**
    Right now the 256 MiB ceiling rarely fires: EMD-0262 is 116 MB and would sail under it. What actually avoids the 116 MB download is putting the density servers first. An alternative design would always prefer the full map and fall back only when it is genuinely huge, which is more predictable but much more expensive by default.

6. **Should a quantitative caller get a different default than a viewer?**
There is already `allowNonRenderableFormats` distinguishing the two audiences; a `preferFullResolution` flag would be a small addition if the distinction is real.

Happy to change any of this — the ordering is one list and a couple of setters.

aalhossary:
**Reviewer note:** This PR (when approved) should be merged after #1133; because the other PR is included, a prerequisite, and a dependency of this PR.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/density/WwpdbMapCoefficientsProvider.java
```diff
@@ -0,0 +1,147 @@
...
+ * @author Amr ALHOSSARY
+ * @since 7.3.0
+ */
+public class WwpdbMapCoefficientsProvider extends AbstractDensityMapProvider {
+
+	/** Default base URL, the wwPDB validation report archive. */
+	public static final String DEFAULT_SERVER_URL = "https://files.wwpdb.org/pub/pdb/validation_reports/";
+
+	/** An RCSB mirror serving byte-identical files. */
+	public static final String RCSB_MIRROR_URL = "https://files.rcsb.org/pub/pdb/validation_reports/";
```
josemduarte: Both wwpdb and rcsb.org servers support short style URLs, e.g. https://files.rcsb.org/validation/download/3hbx_validation_2fo-fc_map_coef.cif.gz . That's important because in 2027 the PDB will transition to the new beta archive with extended PDB ids and a new layout of files and dirs. If you use short style URLs here, then this will automatically work with the new archive from July next year. If you use the full paths with middle hash etc then the paths will need to be fixed next year.

See https://www.rcsb.org/news/feature/6a4fc5f40fba48755e86064b

josemduarte (commented):
Thank you! I skimmed through it and it looks good. See only one comment about the PDB archives layout.

josemduarte:
And one thing to consider: since the tests are new, write them in junit5 (jupiter). Probably an easy task for AI.

aalhossary:
## Comment drafted with the help of Claude

Thanks — changed, and you were right that it reaches past the one file.

## Here are some decisions I took, for your review

### **The endpoint.** 
Coefficients now go through `https://files.wwpdb.org/validation/download/1cbs_validation_2fo-fc_map_coef.cif.gz`. All six host × spelling combinations (wwpdb / rcsb / files-beta, short / extended) return 200 and serve the same bytes, and it holds across entries — `3alb`, `1m3q`, `2atk`, `3hbx`, `8ug2` all 200, while `6hu9` (cryo-EM) and `1a03` (NMR) 404, which is the right answer and is what raises `NoDensityMapException`. This was the only provider that ever built a path; the density servers, PDBe and EMDB are addressed by identifier already, so nothing else here is exposed.

### **Kept `files`, not `files-beta`**
Even though beta serves the same endpoint with newer content. The beta page says it replaces the main archive on 21 July 2027, "after [which] the beta archive URL will be supported with redirection for 3 years" — so beta is the name that would need revisiting twice, while `files.wwpdb.org` simply becomes the new archive. Nothing is lost by preferring it: the two are byte-identical today across the nine entries I compared. Beta is kept as a constant for the opposite reason — it *is* the post-2027 content, so it is how I checked the endpoint's semantics survive the cutover.

### **The spelling is left to `PdbId`.**
As `urlId()` delegates to `PdbId.getId(true)`, which returns the short form where an entry has one and the extended form otherwise, I'm relying on that deliberately.

### **`{mid}` stays; `{extid}` is new.** 
EBI publishes no name-resolving endpoint — three plausible flat forms all 404 — so mirrors still need directory paths, and the divided path survives as `DIVIDED_*_TEMPLATE`. That exposed a real gap: the 2027 layout needs the *extended* ID in two positions regardless of the entry's own spelling, and `{pdbid_lc}` cannot produce it. Hence `{extid}`, which makes the new archive a configuration line rather than a patch:
```
entries/{mid}/{extid}/validation_reports/{extid}_validation_2fo-fc_map_coef.cif.gz
```
kept as `ENTRIES_*_TEMPLATE`, tested against the beta tree's real paths.

### **Cache layout**
Since it is the one thing we can't revise after release: I considered mirroring the per-entry tree and decided against it. A cache cannot become a mirror — the archive publishes structure factors and coefficients but never grids, so density is fetched whatever else is rsynced. And putting our files inside a tree whose contract is "an exact copy of upstream" misbehaves under the documented `rsync --delete`: excluded files are protected, so the usual symptom is rsync failing to remove the directory and reporting it every sync rather than silent loss, but it becomes real deletion under `--delete-excluded`. For a whole-archive sweep it is also ~195k directories holding one or two small files each, against ~1,300 today. If you would rather converge anyway, deciding later is cheap: every cached path is computed in `DensityCacheLayout` and nowhere else. That reasoning is now in its javadoc.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/density/WwpdbMapCoefficientsProvider.java
aalhossary: I replied in a separate comment.

aalhossary:
Following up on the switch to `/validation/download/`: it works, but it costs the checksum, and the integration test caught it.

The divided archive path returns the content MD5 as its `ETag`; the flat endpoint returns neither an `ETag` nor a `Content-Length`:

```
files.wwpdb.org/pub/pdb/validation_reports/cb/1cbs/…   Content-Length: 219757   ETag: "f99fb9d9…5745cf"
files.wwpdb.org/validation/download/…                  (neither)
files.rcsb.org/validation/download/…                   (neither)
```

**That is not an argument for reverting.** Checking all four mirrors, both archives:

| | Content-Length | ETag |
|---|---|---|
| current wwPDB / RCSB, static `/pub` | yes | content MD5 |
| current wwPDB / RCSB, `/download/` shortlinks | no | no |
| **beta** wwPDB / RCSB, **any path** | no | no |
| beta PDBe / PDBj, per-entry | yes | size-mtime |

So the content-MD5 disappears at the July 2027 transition whatever URL form we choose — it is a property of the static file paths, not of the divided layout. Pinning the divided path would buy about ten months and then need redoing.

`d5a8f1fb9` therefore asserts the digest **when the server offered one** and reports when it did not, rather than requiring it:

```
No MD5 recorded for https://files.wwpdb.org/validation/download/1cbs_validation_2fo-fc_map_coef.cif.gz
  - the server offered no usable ETag. Size validation still applies.
```

Size validation is unaffected — the `.size` sidecar is written from the bytes actually read, so it exists either way, and the corruption check at the end of that test still passes. What is lost is detecting a same-length corruption. All six integration tests pass.

If the headers come back, the assertion starts running again with no code change.

**Reported upstream.** The gap is wider than map coefficients: `files.rcsb.org/download/1cbs.cif.gz` — the coordinate shortlink, in production today — also returns no `ETag`, `Content-Length`, `Last-Modified` or `Accept-Ranges`, and neither does any beta path on those two hosts, while PDBe and PDBj beta mirrors return all four. That means no integrity check, no `If-Modified-Since`, and no resumable downloads, on precisely the endpoints the wwPDB news of 15 August recommends for automated workflows. I have written to wwPDB and RCSB about it, with @josemduarte copied.

josemduarte (approved):
LGTM thanks

Issue #1135, opened 2026-08-15
CI red on all PRs since 2025-12: three independent causes (tracking)


aalhossary:
## Summary

Every `PR Build` since 2025-12-19 has failed, on all five matrix jobs, regardless of what the pull request changes. A one-line Dependabot version bump (#1132) fails identically to a large feature branch. Nobody can currently get a green check, and a genuine regression would be indistinguishable from the existing noise.

There are three independent causes, now tracked separately as sub-issues. Two of them are bugs in our own code that affect anyone using `CathInstallation` or `EcodInstallation` at runtime, not only CI.

| | Cause | Sub-issue | Fixed by |
|---|---|---|---|
| 1 | `CathInstallation` downloads over `http`, which now 301s to `https`, and the redirect body is cached as classification data | #1138 | #1133 (`ed36d3a`) |
| 2 | ECOD changed its distribution format; neither the version header nor the column layout is recognised | #1139 | #1141 |
| 3 | `Master Build` fails on an expired or absent `SONAR_TOKEN` | #1140 | needs credentials, no PR possible |

---

## Correction to the original diagnosis

**This issue originally recorded cause 2 as "`EcodInstallationTest.testVersion` downloads 657 MB from a slow host", and described it as a judgement call rather than a bug. That was wrong.**

The download completes. The file is read. The assertion then fails because ECOD changed its version comment from `#ECOD version develop291` to `# Version: v295`, so `getVersion()` falls back to the literal string `"latest"` — exactly what the test asserts against. The column layout changed at the same time, from 15 columns to 23 and then 25, which means `getAllDomains()` on the current release returns an **empty list**, silently, for every user of the library.

The 657 MB is real and is still worth addressing — it is pulled once per matrix job, five times per pull request — but it is a cost, not the failure, and making the download faster would not have fixed anything. Full detail in #1139.

---

## Evidence

| | |
|---|---|
| Last green `PR Build` | 2025-12-14, run `20214548025` |
| Last green `Master Build` | 2025-12-10, run `20115022310` |
| `PR Build` runs since | every one failed |
| Same failures on a one-line dependency bump | #1132 |
| After the CATH fix, ECOD only | run `31916362199` |
| With integration tests excluded, all five green | run `31916395882` |

Logs older than GitHub's 90-day retention have expired, so the *cause* is only confirmable for recent runs. The failure streak itself is visible in run metadata back to 2025-12-19; whether the earliest ones share this cause cannot now be checked.

## Reproducing

```bash
[...]

PR #1136, opened 2026-08-15
Fix CATH downloads: use https, and check the response before caching it

aalhossary:
Fixes cause **1 of 3** in #1135.

`download.cathdb.info` now answers plain http with a 301 to https. `HttpURLConnection` follows redirects within a protocol but deliberately will not follow one that switches http to https, so our download never reached the real file. `downloadFileFromRemote` read the response with a bare `openStream()` and no status check, which meant the body of the 301 was written into the local cache as though it were classification data. Parsing it produced no domains, and the first sign of trouble was a `NullPointerException` much later, far from the cause:

```
CathDomainTest.test:40 NullPointer
  Cannot invoke "CathDomain.toCanonical()" because "domain" is null
```

This has been failing on **every** pull request since the redirect appeared — a one-line dependency bump fails exactly as a feature branch does.

### Changes

1. `CATH_DOWNLOAD_URL` uses `https`, which fixes the immediate breakage.
2. The hand-rolled copy loop is replaced by `FileDownloadUtils.downloadFileWithValidation`, so a non-2xx response throws instead of being cached, and a redirect that changes protocol is logged explicitly. That is what turns this class of failure from a mysterious NPE into an error naming the URL and the status.

The shared download also records the byte count it actually wrote, so CATH files now get size validation they never had. `download.cathdb.info` sends no ETag, so there is no checksum to record, but the size check alone would have caught a truncated file.

`CathDomainTest` passes again, in about 11 seconds.

### Note on the commit list

**This branch is stacked on #1133**, which is where `downloadFileWithValidation` comes from. GitHub therefore shows #1133's three commits here as well. The CATH change itself is the single commit `7244f21` — reviewing that one commit is enough to judge this PR.

If you would rather not couple the two, the `https` one-liner alone turns the test green. I would not recommend stopping there: without the status check, the next time a service changes a redirect we get the same silent corruption and the same puzzling NPE, and it will take just as long to work out why.

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/util/FileDownloadUtils.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/util/FileDownloadUtils.java
@@ -21,9 +21,9 @@
  */
 package org.biojava.nbio.core.util;
 
+import java.io.BufferedInputStream;
 import java.io.File;
 import java.io.FileNotFoundException;
-import java.io.FileOutputStream;
 import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStream;
@@ -32,11 +32,15 @@ import java.net.HttpURLConnection;
 import java.net.SocketTimeoutException;
 import java.net.URL;
 import java.net.URLConnection;
-import java.nio.channels.Channels;
-import java.nio.channels.ReadableByteChannel;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.*;
 import java.nio.file.attribute.BasicFileAttributes;
+import java.security.DigestInputStream;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.Scanner;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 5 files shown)


aalhossary:
To be explicit about scope, since #1135 lists three separate causes and this PR only addresses one:

| # | Cause | Status |
|---|---|---|
| **1** | **CATH downloader follows a stale `http://` URL that now 301s to `https`** | **fixed here** |
| 2 | `EcodInstallationTest.testVersion` downloads 657 MB from a slow server | not addressed — needs a decision, see #1135 |
| 3 | `Master Build` fails on an expired/absent `SONAR_TOKEN` | not addressed — needs someone with the SonarCloud credentials |

So merging this does not turn CI green on its own. It removes one of the two test failures that currently fail every pull request, and it fixes a real runtime bug: anyone calling `CathInstallation` today silently caches the body of a 301 instead of the classification data.

Cause 2 is the other test failure, and it is a judgement call rather than a bug — the download genuinely works, it is just 657 MB from a slow host, pulled once per matrix job. I have deliberately not picked an option for you.

Cause 3 only affects `Master Build`, which does not run on pull requests, so it is invisible here but keeps master red.

Verification for this one: `CathDomainTest` fails on master and passes on this branch in about 11 seconds.

```
mvn verify -pl biojava-integrationtest -Dtest=CathDomainTest
```

aalhossary:
Closing: this belongs in #1133 rather than as a separate pull request, since it builds on the download hardening there. The CATH commit has been moved onto that branch.

Issue #910, opened 2020-12-11
PDBStatusTest might cause flakiness due to network calls


Shilpa-Rani:
The following functions in [PDBStatus](https://github.com/biojava/biojava/blob/master/biojava-structure/src/main/java/org/biojava/nbio/structure/PDBStatus.java) are called from [PDBStatusTest.java:](https://github.com/biojava/biojava/blob/master/biojava-structure/src/test/java/org/biojava/nbio/structure/PDBStatusTest.java)
1.public static Status **getStatus**(String pdbId) throws IOException
2.public static Status[] **getStatus**(String[] pdbIds) throws IOException
3.public static String getCurrent(String oldPdbId) throws IOException

But these function makes network calls, due to latency and network errors, these tests might fail sometimes, introducing flakiness. So, a better way to circumvent this issue is via mocking.

**Example**: testGetStatus() calls getStatus() in PDBStatus and a call is made to the following URL:
https://data.rcsb.org/rest/v1/holdings/status/1HHB
Assert.assertEquals(Status.REMOVED, PDBStatus.getStatus("1HHB"));

Considering the data returned from the call, the resulting JSON node can be marked as:

```
String jsonString = "{\"rcsb_repository_holdings_combined\":{\"id_code_replaced_by_latest\":\"4HHB\",\"status\":\"REMOVED\",\"status_code\":\"OBS\"},\"rcsb_repository_holdings_combined_entry_container_identifiers\":{\"entry_id\":\"1HHB\",\"rcsb_id\":\"1HHB\",\"update_id\":\"2020_48\"},\"rcsb_id\":\"1HHB\"}";
ObjectMapper mapper = new ObjectMapper();
JsonNode **actualObj** = mapper.readTree(jsonString);
```

The getStatus() function is:
```
	/**
	 * Get the status of a PDB id.
	 *
	 * @param pdbId the id
	 * @return The status.
	 */
	public static Status getStatus(String pdbId) throws IOException {
		URL url = new URL(String.format(STATUS_ENDPOINT, DEFAULT_RCSB_DATA_API_SERVER, pdbId.toUpperCase()));
		ObjectMapper objectMapper = new ObjectMapper();
		JsonNode node = objectMapper.readValue(url.openStream(), JsonNode.class);
		return parseStatusRecord(node);
	}
```

We can use power mockito dependency in POM.xml

```
		<!-- https://mvnrepository.com/artifact/org.mockito/mockito-all -->
		<dependency>
[...]

PR #1137, opened 2026-08-16
Run integration tests nightly rather than on every pull request

aalhossary:
Addresses the structural half of #1135.

The integration tests reach CATH, ECOD, RCSB, EBI and others. That is what makes them worth having — they are how we learn that an upstream service changed a URL, a format or a redirect. It is also what makes them a poor gate on pull requests, because any of those services having a bad day blocks every contributor, and a real regression then cannot be distinguished from the surrounding noise.

That is not hypothetical. Every `PR Build` since 2025-12-19 has failed, on all five matrix jobs, whatever the pull request contained — a one-line dependency bump (#1132) fails exactly as a feature branch does. Two integration tests account for it: `CathDomainTest`, because `download.cathdb.info` began redirecting http to https (#1138), and `EcodInstallationTest.testVersion`, because ECOD redesigned its distribution format (#1139).

`Master Build` already excludes the module with `-pl '!biojava-integrationtest'`. This applies the same exclusion to pull requests and adds a scheduled workflow that runs the full suite nightly, so the coverage is kept but is no longer in anybody's way. An upstream break still gets caught, within a day, by a run whose failure means what it says.

### Both failures are now fixed, which changes the argument for this PR

#1133 fixes CATH and #1141 fixes ECOD, so with those merged `PR Build` should be green **with** the integration tests still running. The emergency case for this PR is therefore gone, and I would rather say so than let it merge on momentum.

What is left is the stronger argument, and it is the one worth weighing:

- **Both breaks were upstream changes, and both took months to notice.** ECOD changed its file format at v294.1. The signal we got was `EcodInstallationTest.testVersion:285 Values should be different. Actual: latest`, on somebody else's unrelated pull request, competing for attention with a second failure and with eight months of identical red checks. It named neither ECOD nor the format. A nightly run would have reported it the following morning, alone, against a tree nobody had just changed.
- **Meanwhile the cost is paid by everyone.** `ecod.latest.domains.txt` is 657 MB, pulled once per matrix job, five times per pull request. That is not why the test failed — see #1139 — but it is real, and it buys coverage that a one-second offline test now provides better (#1141 adds fixtures pinning every ECOD file layout the project has ever read).

So the trade is not "keep the build green" any more. It is: should a pull request's check depend on five third-party services being up and fast, when the thing those services actually tell us arrives more usefully as a nightly failure?

### The nightly job

- **One JDK (21).** The pull request build already covers 11, 17 and 21; the point of this run is the network paths, not the language level.
- **90 minute timeout.** These downloads are large and GitHub's six hour default is not a useful ceiling if something hangs.
- **Surefire reports kept 30 days on failure.** GitHub expires run logs after 90 days, and those reports carry the stack traces identifying which service misbehaved — exactly what was missing when diagnosing the current breakage, where the logs for the older failures had already expired.
- **`workflow_dispatch`**, so it can be run by hand to confirm a service is back.

### What this gives up

Plainly: a pull request that breaks an integration test is not caught until that night. For changes touching the download or parsing paths, that is a real regression window. If that trade is unacceptable — and now that both tests are fixed it is a much more open question than when I opened this — the coherent response is to close this PR. I would rather that than have it merged by default.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,54 @@
+name: Nightly Integration Tests
+
+# The integration tests reach out to CATH, ECOD, RCSB, EBI, UniProt and others.
+# That makes them valuable - they are how we find out that an upstream service
+# has changed a URL, a format or a redirect - but it also makes them unsuitable
+# as a gate on pull requests, because an outage anywhere blocks every
+# contributor. Running them on a schedule keeps the coverage while decoupling it
+# from people's ability to merge.
+on:
+  schedule:
+    # 03:17 UTC daily. Off the hour deliberately: scheduled jobs that ask for
+    # exactly midnight queue behind everybody else's.
+    - cron: '17 3 * * *'
+  # Also runnable by hand, e.g. to confirm an upstream service is back.
+  workflow_dispatch:
+
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
+jobs:
+  integrationtest:
+    runs-on: ubuntu-latest
+    # These tests download large files from servers we do not control; the
+    # default 6 hour limit is far more than they need and far more than we want
+    # to spend if one of them hangs.
+    timeout-minutes: 90
+    strategy:
+      matrix:
+        # One JDK only. The point of this run is to exercise the network paths,
+        # not the language level, which the pull request build already covers
+        # across 11, 17 and 21.
+        java: [21]
+      fail-fast: false
+    name: Integration tests, JDK ${{ matrix.java }}
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'oracle'
+          java-version: ${{ matrix.java }}
+      - name: Build and run integration tests
+        run: mvn verify --no-transfer-progress
+      - name: Upload surefire reports
+        # Kept on failure so an upstream break can be diagnosed after the fact:
+        # GitHub expires run logs after 90 days, and these reports carry the
+        # stack traces that say which service misbehaved.
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: surefire-reports
+          path: '**/target/surefire-reports/**'
+          retention-days: 30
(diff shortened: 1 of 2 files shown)


aalhossary:
### On reversing a deliberate choice

Worth putting the history on the table, since this PR undoes something that was decided on purpose rather than by accident.

#910 raised network-dependent tests as a flakiness risk in 2020. Closing it in 2023, @josemduarte wrote:

> It is true that it is not ideal that there's a dependency on network in some tests. We should strive to have all tests that depend on network in integration-test module.
>
> Note: since transitioning to github workflows some time ago, we run integration-tests only in pull requests but not in master

So integration tests were deliberately placed on pull requests, and deliberately kept off master. This PR takes them off the only place they currently run. That deserves an argument rather than a shrug.

The argument is that the premise has stopped holding. Running them on pull requests is sound for as long as a red check carries information. Since 2025-12-19 it has not: **every** `PR Build` has failed, whatever the pull request contained — #1132, a one-line dependency bump, fails exactly as a feature branch does. Eight months of a check that is red regardless of input is not a degraded signal, it is no signal, and it costs something: nobody can distinguish a genuine regression from the background, so the rational response is to stop reading it, which is worse than not having it.

#1022 is also relevant. It tried to solve this from the other end, with a JUnit rule that skips tests when a network exception occurs. It stayed open for nine months and was closed unmerged. That is not a criticism of the approach — it suggests making network tests reliable in place is harder than changing when they run.

**What this keeps** is the coverage @josemduarte asked for. The tests still run, nightly, against the same services, and an upstream change is still caught within a day — by a run whose failure names the service and is not competing for attention with unrelated pull requests. #1135 exists because that signal did eventually get through; it just took eight months and a manual investigation rather than one nightly failure.

**What this gives up**, plainly: a pull request that breaks an integration test is not caught until that night. For changes touching the download or parsing paths, that is a real regression window. If that trade is unacceptable, the coherent alternative is to fix the two tests instead and close this — and that alternative has since been carried out: #1133 fixes CATH and #1141 fixes ECOD, so the build can be green with the integration tests still running. I would rather you reject this deliberately than merge it by default.



---

*Edited: ECOD turned out to be a bug rather than something needing a decision — the distribution format changed (#1139), fixed by #1141. Worth noting that it is also the best evidence for this PR: a nightly run would have named it the next morning instead of it hiding for months inside an unrelated red check.*

josemduarte (approved):
LGTM thanks. Let's give this a try. It is something that we were aiming to do but we never got to it. Integration tests are causing too much pain for the rest of the project and are being a blocker rather than a help.

PR #1141, opened 2026-08-17
Support the new ECOD distribution format

### LINKED ISSUE #1139: ECOD changed its distribution format: neither the version header nor the column layout is recognised
After submitting the issues/PRs yesterday, I spent overnight digging the exact reason for the exception on the "version test" until I found it and so fixed it. I asked my AI agent to continue the effort.

Sub-issue of #1135, cause 2 of 3 — and the earlier diagnosis there was wrong. See the last section.

## What changed upstream

ECOD redesigned its distribution at v294.1. Two things in the file break us.

### 1. The version comment

| Release | Header line |
|---|---|
| up to develop292 | `#ECOD version develop291` |
| v294.1 | `# Version: v294.1` |
| v295 (current `latest`) | `# Version: v295` |

`EcodInstallation.EcodParser` matches `^\s*#.*ECOD\s*version\s+(\S+).*`, which the new form does not satisfy. `parsedVersion` stays null and `getVersion()` falls back to the string that was *requested* — the literal `"latest"`. That is precisely what the test asserts against:

```
EcodInstallationTest.testVersion:285  Values should be different. Actual: latest
```

### 2. The column layout

| Format | Columns | Column header row |
|---|---|---|
| v1.4 (develop124) / v1.5 (develop204) | 15 | `#uid …` |
| v1.6 (develop291) | 16 — `f_id` renamed `t_id`, `unp_acc` inserted at column 9 | `#uid …` |
| v294.1 | 23 | `uid …` — **no longer commented out** |
| v295 | 25 — adds `ligand_comp_ids`, `ligand_pdbnum` | `uid …` |

The parser accepts only 13, 14 or 15 fields, so every data line in v294.1 and v295 is rejected. Other things that changed with it:

- `manual_rep` now holds `True`/`False` rather than `MANUAL_REP`/`AUTO_NONREP`;
- `assembly_id` and `domain_id_short` are declared but **empty on all 2,945,500 rows**;
- the ligand list moved from the last column to `ligand_comp_ids` (column 24);
- `f_name` is empty rather than `F_UNCLASSIFIED` for unclassified domains, while `f_id` now carries a fourth level (`1.1.1.3`);
- `uid` restarts from 0;
- **1,573,776 of the 2,945,500 rows (53%) are domains classified from AlphaFold models**, with `ecod_domain_id` values like `P44140_F1_nD2` and an empty `pdb` column. `EcodDomain` is keyed by `PdbId` and cannot represent them.

`ecod.latest.domains.txt` itself still resolves, so no URL change is needed. The new `/ecod/af2_pdb/distribution` index links into the same `/ecod/distributions/` directory — it is a second index over one file set, not a second host.

## Impact beyond CI

This is not only a test failure. Anyone calling `EcodFactory.getEcodDatabase()` today — which defaults to `latest` — gets an **empty domain list**, silently:

```java
EcodDatabase ecod = EcodFactory.getEcodDatabase();
ecod.getVersion();          // "latest", not "v295"
ecod.getAllDomains();       // empty
ecod.getDomainsForPdb("1cbs"); // null
```

The parser logs "Unexpected number of fields", capped at 10 warnings, and then hands back nothing. No exception is thrown.

## Correcting the earlier diagnosis

#1135 recorded this as *"`EcodInstallationTest.testVersion` downloads 657 MB from a slow host"* and described it as a judgement call rather than a bug. **That was wrong, and I would rather say so plainly than quietly move on.** The download completes, the file is read, and the assertion then fails on the version string. The 657 MB is a genuine CI cost — pulled once per matrix job, five times per pull request — but it is not why the test fails, and no amount of making the download faster would have fixed it. That cost is an argument for #1137, not a bug in itself.

The reason a wholesale format change surfaced as a single one-line assertion failure is that `EcodInstallationTest` pins `VERSION = "develop204"` everywhere else. Only `testVersion` touches `latest`.

## Reproducing

```bash
mvn verify -pl biojava-integrationtest -Dtest='EcodInstallationTest#testVersion'
```

Or without the download:

```bash
curl -r 0-400 http://prodata.swmed.edu/ecod/distributions/ecod.latest.domains.txt
```

## Fix

A pull request follows. It reads files that declare a column header **by column name rather than by position**, which covers all four layouts at once and is the only approach that will not need revisiting the next time a column moves; keeps a positional fallback for pre-develop101 files that have no header; matches both version comment forms; and adds an offline unit test pinning every layout, so the next format change is caught in milliseconds rather than by a 657 MB download that only one test exercises.

aalhossary:
Fixes #1139.

ECOD redesigned its distribution at v294.1. Neither the version comment nor the column layout is recognised any more, so `EcodInstallationTest.testVersion` fails and — less visibly — anyone calling `EcodFactory.getEcodDatabase()` today gets an empty domain list with no exception thrown.

## The version comment

| Release | Header line |
|---|---|
| up to develop292 | `#ECOD version develop291` |
| v294.1 onwards | `# Version: v295` |

`parsedVersion` stayed null and `getVersion()` fell back to the string that was *requested*, which for `latest` is the literal `"latest"` — the value the test asserts against. One pattern now matches both forms, so the historical releases keep working; a regex that only accepted the new form would silently break every pinned `develop*` version.

## The columns

| Format | Columns | Column header row |
|---|---|---|
| v1.4 / v1.5 (develop124, develop204) | 15 | `#uid …` |
| v1.6 (develop291) | 16 — `f_id` renamed `t_id`, `unp_acc` inserted at 9 | `#uid …` |
| v294.1 | 23 | `uid …` — no longer commented out |
| v295 | 25 — adds `ligand_comp_ids`, `ligand_pdbnum` | `uid …` |

Rather than add a fourth positional special case, files that declare a column header — every release since develop101 — are now read **by column name**. That covers all four layouts at once and will not need revisiting the next time a column moves. Files older than develop101 have no header and keep the existing positional path, unchanged.

Everything else that moved with the columns:

- `manual_rep` holds `True`/`False` rather than `MANUAL_REP`/`AUTO_NONREP`;
- `assembly_id` is declared but empty on all 2,945,500 rows, meaning what `NOT_DOMAIN_ASSEMBLY` used to mean;
- the ligand list moved to `ligand_comp_ids`;
- `ligand_pdbnum`, the last column, is empty on 2,399,569 rows, so `split("\t")` was dropping it and making those rows look a column short. Trailing empty fields are now kept;
- `f_name` is empty rather than `F_UNCLASSIFIED` for unclassified domains — deliberately left as it is, see below;
- **1,573,776 rows (53%) describe domains found in AlphaFold models**, with ids like `P44140_F1_nD2` and an empty `pdb` column. `EcodDomain` is keyed by `PdbId` and cannot represent them, so they are counted and reported at INFO rather than each logged as an error.

Two failure modes that hid this are also closed: parsing *nothing at all* is now logged as an error instead of quietly returning an empty list, and the per-line `NumberFormatException` warning is capped like the three warnings beside it, rather than printing a stack trace per row.

## `getVersion()` no longer parses the whole file

It reads the header. Parsing the current release in full to answer "which version?" costs **1,186 MB of heap and 9.8 s** — and the test JVM is given `-Xmx1500M` by the root pom. The download is unchanged; only the parse is skipped. This is a small behaviour change, in that `getVersion()` no longer has the side effect of loading every domain, which the 7.3.0 target allows.

## Measured

| File | domains before | domains after |
|---|---|---|
| `ecod.develop124.domains.txt` | 468,680 | 468,680 |
| `ecod.develop204.domains.txt` | 592,427 | 592,427 |
| `ecod.develop291.domains.txt` | **0** | 1,083,021 |
| `ecod.latest.domains.txt` (v295) | **0** | 1,371,723 + 1,573,776 model rows reported |

The first two are the counts `EcodInstallationTest` already asserts, so the old path is provably untouched.

## Tests

`EcodParserTest` is new and entirely offline. `EcodParser` already accepts a `Reader`, so fixtures taken verbatim from the real files pin all five layouts, both version comment forms, a row ending in an empty column, an AlphaFold-derived row and a malformed row. 15 tests, no network, well under a second:

```
mvn test -pl biojava-structure -Dtest=EcodParserTest
```

That is the part I would most like kept. The distribution is 657 MB, so the only test that ever touched the current release was `testVersion`, asserting on one string; everything else pins `develop204`. That is exactly why a wholesale format change surfaced as a single one-line assertion failure, months after it happened.

`testVersion` now also parses the first few thousand lines of the local file — enough to notice a column change without building three million domains.

```
mvn test -pl biojava-integrationtest -Dtest=EcodInstallationTest
```

9 tests, 1 skipped (`testAllVersions`, already `@Ignore`d for being slow), green in 124 s.

## Note on CI

Branched off `master`, so its `PR Build` will still show the CATH failure until #1133 merges. That is #1138, not something new here.

## Three judgement calls I would rather not make alone

@sbliven — you wrote this parser and know the format history far better than I do. If you have the time, I would value your eyes on it, particularly on:

1. **`unp_acc` is read and discarded.** develop291 carries a UniProt accession that `EcodDomain` has nowhere to put. Adding a field means touching `equals`/`hashCode`/`toString`/`Comparable`, so I left it out — but it is real data we are dropping.
2. **An empty `f_name` is left empty rather than translated to `F_UNCLASSIFIED`.** They used to be equivalent; now `f_id` classifies to a fourth level (`1.1.1.3`) while the name is blank, so I do not think they are the same thing any more. If you disagree it is a one-line change.
3. **An empty `assembly_id` is mapped onto the old `NOT_DOMAIN_ASSEMBLY` semantics** (`assemblyId = uid`). It is empty on every row of v294+, so assembly information is simply gone from the distribution; this keeps `getAssemblyId()` returning what callers expect rather than null.

Also deliberately out of scope, and worth a follow-up issue rather than growing this PR: `valid_structure`, `ligand_binding`, `ligand_pdbnum`, `range_count` and the five `*_manual` flags are new columns with no home in `EcodDomain`, and the AlphaFold half of the distribution needs a model that is not keyed by `PdbId`.

## Unrelated but useful

ECOD now publishes `ecod.latest.md5` in coreutils format alongside the distribution. That is a real `hashURL` for `FileDownloadUtils.createValidationFiles(url, file, hashURL, Hash.MD5)` — the path #1133 repairs and which, before it, could never write a hash file at all. Not wired up here; noting it so it is not lost.

### DIFF
--- a/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/ecod/EcodInstallationTest.java
+++ b/biojava-integrationtest/src/test/java/org/biojava/nbio/structure/test/ecod/EcodInstallationTest.java
@@ -22,8 +22,12 @@ package org.biojava.nbio.structure.test.ecod;
 
 import static org.junit.Assert.*;
 
+import java.io.BufferedReader;
 import java.io.File;
+import java.io.FileReader;
 import java.io.IOException;
+import java.io.Reader;
+import java.io.StringReader;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -47,6 +51,7 @@ import org.biojava.nbio.structure.ecod.EcodDatabase;
 import org.biojava.nbio.structure.ecod.EcodDomain;
 import org.biojava.nbio.structure.ecod.EcodFactory;
 import org.biojava.nbio.structure.ecod.EcodInstallation;
+import org.biojava.nbio.structure.ecod.EcodInstallation.EcodParser;
 import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
@@ -277,12 +282,47 @@ public class EcodInstallationTest {
 		assertEquals(expected,actual);
 	}
 
+	/**
+	 * Checks that the current release can still be read.
+	 * <p>
+	 * The version is read from the file's header without parsing the domains, so this
+	 * additionally parses the first few thousand lines of the same file. That is enough to
+	 * notice a column change — which is what ECOD did at v294.1, unnoticed for months —
+	 * without building the three million domains the whole file now holds.
+	 */
 	@Test
 	public void testVersion() throws IOException {
 		EcodDatabase ecod3 = EcodFactory.getEcodDatabase("latest");
 		String version = ecod3.getVersion();
 		assertNotNull(version);
 		assertNotEquals("latest", version);
+		System.out.println("latest version of ECOD is "+version);
+
+		File domainsFile = new File(((EcodInstallation) ecod3).getCacheLocation(),
+				"ecod.latest.domains.txt");
+		assertTrue("No local copy of the domains file at "+domainsFile, domainsFile.exists());
+
+		EcodParser parser = new EcodParser(firstLines(domainsFile, 5000));
+		assertEquals(version, parser.getVersion());
+		assertFalse("No domains parsed from ECOD "+version
+				+ "; the distribution format has probably changed",
+				parser.getDomains().isEmpty());
+	}
+
+	/**
+	 * @return a reader over the first {@code maxLines} lines of the file
+	 */
+	private static Reader firstLines(File f, int maxLines) throws IOException {
+		StringBuilder head = new StringBuilder();
+		try (BufferedReader in = new BufferedReader(new FileReader(f))) {
+			String line;
+			int n = 0;
+			while (n < maxLines && (line = in.readLine()) != null) {
+				head.append(line).append('\n');
+				n++;
+			}
+		}
+		return new StringReader(head.toString());
 	}
 
 	/**
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/ecod/EcodInstallation.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/ecod/EcodInstallation.java
@@ -272,17 +272,70 @@ public class EcodInstallation implements EcodDatabase {
 	 *
 	 * Note that this may differ from the version requested in the constructor
 	 * for the special case of "latest"
+	 * <p>
+	 * Since 7.3.0 this reads only the file's header rather than parsing the whole
+	 * file, so it no longer has the side effect of loading every domain.
 	 * @return the ECOD version
 	 * @throws IOException If an error occurs while downloading or parsing the file
 	 */
 	@Override
 	public String getVersion() throws IOException {
-		ensureDomainsFileInstalled();
+		domainsFileLock.readLock().lock();
+		logger.trace("LOCK readlock");
+		try {
+			if( parsedVersion != null ) {
+				return parsedVersion;
+			}
+		} finally {
+			logger.trace("UNLOCK readlock");
+			domainsFileLock.readLock().unlock();
+		}
+
+		// The version is declared in the first few lines of the file, so read those rather
+		// than the millions of domain records behind them. The current release is 657 MB and
+		// holds nearly three million records; parsing it in full to answer this question
+		// costs over a gigabyte of heap and several seconds.
+		ensureDomainsFileDownloaded();
+
+		domainsFileLock.writeLock().lock();
+		logger.trace("LOCK writelock");
+		try {
+			if( parsedVersion == null ) {
+				parsedVersion = parseVersionOnly();
+			}
+		} finally {
+			logger.trace("UNLOCK writelock");
+			domainsFileLock.writeLock().unlock();
+		}
 
 		if( parsedVersion == null) {
 			return requestedVersion;
 		}
 		return parsedVersion;
+	}
+
+	/**
+	 * Reads the version from the header of the local domains file without parsing the
+	 * domains themselves.
+	 * @return the version, or null if the header does not declare one
+	 * @throws IOException if the file cannot be read
+	 * @since 7.3.0
+	 */
+	private String parseVersionOnly() throws IOException {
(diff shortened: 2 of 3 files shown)


aalhossary:
CI result on this branch, as expected: **five jobs fail, all on `CathDomainTest` alone.** ECOD no longer appears anywhere (run `31985806634`).

```
[ERROR] Tests run: 164, Failures: 0, Errors: 1, Skipped: 10
[ERROR]   CathDomainTest.test:40 NullPointer
```

`EcodInstallationTest` passes on every job:

```
[INFO] Running org.biojava.nbio.structure.test.ecod.EcodInstallationTest
latest version of ECOD is v295
[WARNING] Tests run: 9, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 164.1 s
```

That is the last remaining failure in the module, it is #1138, and #1133 fixes it — so with #1133 and this merged, `PR Build` should be green with the integration tests still running, for the first time since 2025-12-14.

The `ERROR`-level ECOD line visible in the `biojava-structure` log is deliberate: `EcodParserTest.unparseableLinesAreSkippedNotFatal` feeds the parser a malformed row on purpose, to check that one bad line does not take the good ones with it.

josemduarte:
@aalhossary can you merge master into this branch? I'm hoping that the tests will pass now that separate CATH issue is fixed by #1133

josemduarte (approved):
LGTM thanks! Tests passing (with no integration). I'll run the integrationtests locally before merging.

josemduarte:
All good: integration tests are finally passing with this and with #1145

PR #1142, opened 2026-08-17
Bump the development version to 7.3.0-SNAPSHOT 

aalhossary:
The electron density work and the FileDownloadUtils changes that go with it add public API and change a default, which belongs in a minor release rather than a patch. Their @since tags say 7.3.0, so the project version should agree.

Standalone so it can be taken or dropped independently of the feature work, and so the release manager decides when it lands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,55 @@
 BioJava Changelog
 -----------------
 
+BioJava 7.3.0
+==============================
+### Added
+* Fetching, caching and display of electron density and cryo-EM maps #1134
+* Checksum verification (MD5, SHA-1, SHA-256) in `FileDownloadUtils`, replacing the previous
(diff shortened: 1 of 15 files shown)


aalhossary:
With #1133 and #1141 merged, 26 tags across `FileDownloadUtils`, `HttpStatusException`, `LocalPDBDirectory`, `EcodInstallation` and their tests have their javadoc promises a version not yet available.

I rebased this commit onto current master and CI is green on all five jobs.                                                                                                                                                                                                            

The new version should probably land soon, especially as soon as #1134 is merged.

josemduarte (approved):
LGTM thanks

josemduarte:
@aalhossary one ask: could you start drafting a 7.3.0 section in CHANGELOG.md with all changes that will go into that release? Thank you!

aalhossary:
Sure, as soon as I finish some tasks this weekend inshaAllah.

PR #1144, opened 2026-08-18
Write the entry identifier as a data item, not only as the block name

### LINKED ISSUE #1143: mmCIF writer emits no entry identifier: getPdbId() is lost on a write-then-read round trip
### Summary

`AbstractCifFileSupplier` writes the PDB identifier **only** as the name of the `data_` block. It emits neither `_entry.id` nor `_struct.entry_id`, so the identifier is lost for every consumer that reads it as a data item rather than from the block header.

Two independent consumers are affected:

| consumer | reads | result today |
|---|---|---|
| BioJava's own `CifStructureConsumerImpl.consumeStruct` | `_struct.entry_id` | `Structure.getPdbId()` is **null** after a write-then-read round trip |
| Jmol's mmCIF reader | `_entry.id` | `_M.pdbID` is **empty** for a structure passed via `openStringInline(structure.toMMCIF())` |

### Steps to reproduce

```java
Structure s = CifStructureConverter.fromInputStream(
        new GZIPInputStream(...open 4hhb.cif.gz...));
System.out.println(s.getPdbId());                 // 4HHB

String cif = CifStructureConverter.toText(s);
System.out.println(cif.contains("_entry.id"));        // false
System.out.println(cif.contains("_struct.entry_id")); // false

Structure back = CifStructureConverter.fromInputStream(
        new ByteArrayInputStream(cif.getBytes()));
System.out.println(back.getPdbId());              // null   <-- expected 4HHB
```

Actual output of the block header and identifier items as written today:

```
data_4HHB
(no _entry.id, no _struct.entry_id)
```

### Expected

`Structure.getPdbId()` should survive a `toText()` / `fromInputStream()` round trip, and the written file should carry the identifier where mmCIF consumers look for it.

### Why both items matter

`_entry.id` is the canonical place for the identifier and is what Jmol reads. BioJava's own reader (`CifStructureConsumerImpl.consumeStruct`) only looks at `_struct.entry_id`. Writing just one of the two leaves the other consumer broken, so both should be written.

Measured with Jmol, loading the identical structure inline with one item added at a time:

```
as written today (block header only)  ->  _M.pdbID = []
plus  _entry.id         3ALB          ->  _M.pdbID = [3ALB]
plus  _struct.entry_id  3ALB          ->  _M.pdbID = []
plus  both                            ->  _M.pdbID = [3ALB]
```

A practical consequence downstream: Jmol's `isosurface ... eds` fetches an electron-density box around the selection, but it needs an entry to ask about. With `load =3alb` it renders; with the same structure pushed inline via `toMMCIF()` it renders nothing, purely because the identifier was lost.

### Location

`biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/AbstractCifFileSupplier.java`, in `getInternal(...)` — the identifier appears only in the `enterBlock(...)` call.

### Version

Reproduced on `master` (`5b29945`). The behaviour is long-standing, not a recent regression.

aalhossary:
Fixes #1143

### The problem

`AbstractCifFileSupplier` wrote the PDB identifier only as the name of the `data_` block, and emitted neither `_entry.id` nor `_struct.entry_id`. Consumers read the identifier as a data item, not from the block header, so it did not survive being written:

- BioJava's own `CifStructureConsumerImpl.consumeStruct` reads `_struct.entry_id`, so `Structure.getPdbId()` came back **null** after a `toText()` / `fromInputStream()` round trip.
- Jmol's mmCIF reader takes `_M.pdbID` from `_entry.id`, so a structure handed over with `openStringInline(structure.toMMCIF())` arrived with no identifier — which in turn breaks anything keyed on the entry, such as `isosurface ... eds`, which needs an entry to fetch the electron density map for.

### The change

Write both categories immediately after entering the block, guarded on the identifier being present (mirroring the existing null check):

```java
blockBuilder.enterEntry().enterId().add(pdbId.getId()).leaveColumn().leaveCategory();
blockBuilder.enterStruct().enterEntryId().add(pdbId.getId()).leaveColumn().leaveCategory();
```

Both are needed, not one: `_entry.id` is the canonical item and the one Jmol reads, while BioJava's own reader only looks at `_struct.entry_id`, so omitting that one would leave BioJava still unable to read back what it just wrote.

`_struct` was not previously written anywhere in this class (only `_struct_keywords`, a separate category), so there is no duplicate-category concern.

### Tests

`CifFileSupplierImplTest` gains two cases — there was previously nothing asserting that an identifier survives `toMMCIF()`:

- `shouldWriteEntryIdAndSurviveRoundTrip` — reads `4hhb.cif.gz`, asserts both items are written, reads the text back and asserts `getPdbId()` still equals `4HHB`.
- `shouldNotWriteEntryIdWhenPdbIdIsAbsent` — a structure with no identifier gains no empty entry categories.

Verified that the round-trip test fails on unpatched `master` (`_entry.id must be written`) and passes with the change; the pre-existing test in that class is unaffected.

### Notes

- Based directly on upstream `master` (`5b29945`); independent of any other open PR.
- Behaviour change is additive — two items appear in written mmCIF files that were absent before, and only when the structure carries an identifier.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/AbstractCifFileSupplier.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/AbstractCifFileSupplier.java
@@ -40,8 +40,27 @@ public abstract class AbstractCifFileSupplier<S> implements CifFileSupplier<S> {
         // entity information
         List<EntityInfo> entityInfos = structure.getEntityInfos();
 
+        PdbId pdbId = structure.getPdbId();
+
         MmCifBlockBuilder blockBuilder = CifBuilder.enterFile(StandardSchemata.MMCIF)
-                .enterBlock(structure.getPdbId() == null? "" : structure.getPdbId().getId());
+                .enterBlock(pdbId == null? "" : pdbId.getId());
+
+        if (pdbId != null) {
+            // The block header alone does not carry the identifier for consumers: readers pick it up from
+            // _entry.id (e.g. Jmol) or from _struct.entry_id (BioJava's own CifStructureConsumerImpl).
+            // Both are written so that the identifier survives a write-then-read round trip either way.
+            blockBuilder.enterEntry()
+                    .enterId()
+                    .add(pdbId.getId())
+                    .leaveColumn()
+                    .leaveCategory();
+
+            blockBuilder.enterStruct()
+                    .enterEntryId()
+                    .add(pdbId.getId())
+                    .leaveColumn()
+                    .leaveCategory();
+        }
 
         blockBuilder.enterStructKeywords().enterText()
         .add(String.join(", ", structure.getPDBHeader().getKeywords()))
--- a/biojava-structure/src/test/java/org/biojava/nbio/structure/io/cif/CifFileSupplierImplTest.java
+++ b/biojava-structure/src/test/java/org/biojava/nbio/structure/io/cif/CifFileSupplierImplTest.java
@@ -1,5 +1,6 @@
 package org.biojava.nbio.structure.io.cif;
 
+import org.biojava.nbio.structure.PdbId;
 import org.biojava.nbio.structure.Structure;
 import org.biojava.nbio.structure.io.FileParsingParameters;
 import org.biojava.nbio.structure.io.PDBFileParser;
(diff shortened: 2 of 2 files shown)


josemduarte (approved):
LGTM thanks. But let's wait for #1141 to be merged so that we let CI run a clean testing round on this.

PR #1145, opened 2026-08-26
Fix ArrayIndexOutOfBoundsException and file validation in CathInstallation

josemduarte:
Even after the CATH url was fixed there were still some errors in the integration test. This should fix them.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathInstallation.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathInstallation.java
@@ -535,7 +535,7 @@ public class CathInstallation implements CathDatabase{
 	private CathNode parseCathNamesFileLine(String line) {
 		String[] token = line.trim().split("\\s+",3);
 		if (token.length < 3) {
-			LOGGER.info("Invalid line in cath names file, was expecting 3 tokens but got {} tokens: {}", token.length, line);
+			LOGGER.debug("Invalid line in cath names file, was expecting 3 tokens but got {} tokens: {}", token.length, line);
 			return null;
 		}
 		CathNode cathNode = new CathNode();
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathInstallation.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathInstallation.java
@@ -350,15 +350,15 @@ public class CathInstallation implements CathDatabase{
 		parseCathDomainList(buffer);
 	}
 
-	private void parseCathDomainList(BufferedReader bufferedReader) throws IOException{
+	protected void parseCathDomainList(BufferedReader bufferedReader) throws IOException{
 		String line;
-	 //   int counter = 0;
+		int counter = 0;
 		while ( (line = bufferedReader.readLine()) != null ) {
 			if ( line.startsWith("#") ) continue;
 			if ( line.trim().isEmpty() ) continue;
 			CathDomain cathDomain = parseCathListFileLine(line);
 			if ( cathDomain == null ) continue;
-		   // counter++;
+			counter++;
 
 			String pdbId = cathDomain.getPdbIdAndChain().substring(0,4); // includes chain letter
 
@@ -373,6 +373,9 @@ public class CathInstallation implements CathDatabase{
 			domainList.add(cathDomain);
 
 			domainMap.put( cathDomain.getDomainName(), cathDomain );
+		}
+		if (counter == 0) {
+			throw new IOException("Could not parse any CATH domains from the domain list file.");
 		}
 	}
 
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathInstallation.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathInstallation.java
@@ -535,6 +535,7 @@ public class CathInstallation implements CathDatabase{
 	private CathNode parseCathNamesFileLine(String line) {
 		String[] token = line.trim().split("\\s+",3);
 		if (token.length < 3) {
+			LOGGER.info("Invalid line in cath names file, was expecting 3 tokens but got {} tokens: {}", token.length, line);
 			return null;
 		}
 		CathNode cathNode = new CathNode();
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathInstallation.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathInstallation.java
@@ -355,7 +355,9 @@ public class CathInstallation implements CathDatabase{
 	 //   int counter = 0;
 		while ( (line = bufferedReader.readLine()) != null ) {
 			if ( line.startsWith("#") ) continue;
+			if ( line.trim().isEmpty() ) continue;
 			CathDomain cathDomain = parseCathListFileLine(line);
+			if ( cathDomain == null ) continue;
 		   // counter++;
 
 			String pdbId = cathDomain.getPdbIdAndChain().substring(0,4); // includes chain letter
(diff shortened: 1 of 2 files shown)


biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathInstallation.java
```diff
@@ -504,8 +509,11 @@ private void parseCathDomainDescriptionFile(BufferedReader bufferedReader) throw
 	}*/
 
 	private CathDomain parseCathListFileLine(String line) {
+		String [] token = line.trim().split("\\s+");
+		if (token.length < 12) {
```
aalhossary, line 516: **The skip is silent.** Returning `null` here, with the caller doing `continue`, means a truncated or malformed CATH file yields fewer domains and nothing says so.

That is the same failure shape as #1139: ECOD returned an empty domain list for months because the parser rejected every line quietly. A capped warning would cost two lines:

```java
if (token.length < 12) {
    if (warnShortLines-- > 0) logger.warn("Skipping malformed domain list line: {}", line);
    return null;
}
```

Capped rather than unconditional, so a badly broken file cannot produce a warning per line.

biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathInstallation.java
```diff
@@ -522,8 +530,11 @@ private CathDomain parseCathListFileLine(String line) {
 	}
 
 	private CathNode parseCathNamesFileLine(String line) {
+		String[] token = line.trim().split("\\s+",3);
+		if (token.length < 3) {
```
aalhossary, line 537: **This one may be too strict.** With `split("\\s+", 3)`, a node whose description happens to be empty yields 2 tokens and is dropped as malformed, losing a real node.

Is a missing description actually invalid here, or should it default to `""` and keep the node? The domain-list guard above is unambiguous — all 12 columns are read — but this one changes what counts as a valid record rather than just guarding an index.

biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathInstallation.java
```diff
@@ -671,25 +682,25 @@ protected void downloadFileFromRemote(URL remoteURL, File localFile) throws IOEx
 	private boolean domainDescriptionFileAvailable(){
 		String fileName = getDomainDescriptionFileName();
 		File f = new File(fileName);
-		return f.exists();
+		return f.exists() && FileDownloadUtils.validateFile(f);
```
aalhossary, line 689: I am happy that `FileDownloadUtils.validateFile()` is being immediately used.  :-)

biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathInstallation.java
josemduarte: Fair point: I've added a safeguard (plus a test) with 83fe61a

biojava-structure/src/main/java/org/biojava/nbio/structure/cath/CathInstallation.java
josemduarte: I think that before it would have had an exception if description was missing. With this it is a little more robust. I'd argue that a logging statement would be nice anyway, now added with 27bb418c

aalhossary (approved):
Both look good. Throwing is better than the warning I suggested.
An empty domain map only surfaces later and further away. 
And you're right about the names file: the old code would have thrown on token[2], so returning null is strictly an improvement.

aalhossary (approved):

biojava-structure/src/test/java/org/biojava/nbio/structure/cath/CathInstallationTest.java
```diff
@@ -0,0 +1,84 @@
...
+		String data = "# CATH domain list\n"
+				+ "\n"
+				+ "1oaiA00     1    10   490    10     1     1     1     1     1   124 1.80\n"
+				+ "1oaiA01     1    10   490    10     1     1     1     1     2   150 1.80\n";
+
+		CathInstallation installation = new CathInstallation("");
+		BufferedReader reader = new BufferedReader(new StringReader(data));
+		installation.parseCathDomainList(reader);
+
+		CathDomain domain = installation.getDomainByCathId("1oaiA00");
```
aalhossary: ### This is why both Windows jobs fail

```
CathInstallationTest.testParseCathDomainListSuccess:50  expected: <490> but was: <8>
```

`8` is the *real* topology id for `1oaiA00` in CATH. The fabricated data is parsed correctly and then overwritten by the actual domain list.

The route is `getDomainByCathId` → `ensureDomallInstalled` → `ensureDomainListInstalled`. Calling `parseCathDomainList(reader)` directly populates `domainMap` but never sets `installedDomainList`, so the accessor still believes nothing is installed, downloads the real file and re-parses it over the test data.

**The three Linux passes are luck rather than correctness.** `ensureDomainListInstalled` swallows download failures:

```java
} catch (Exception e){
    LOGGER.error("Could not download CATH domain list file. Error: {}", e.getMessage());
    installedDomainList.set(false);
    return;      // test data survives
}
```

On the Linux runners the download did not complete in time, so the fabricated data survived and the assertions passed. On Windows it completed. That is a race against a large download, not an OS difference — the same test can fail on Linux on any day CATH happens to be fast.

Small irony: it only reaches the server at all because of `ed36d3a` in #1133. Before that, plain http always 301'd and the download always failed, so this test would have passed everywhere for the wrong reason.

**Fix** — both flags already have public setters, so no visibility change is needed:

```java
installation.parseCathDomainList(reader);
installation.setInstalledDomainList(new AtomicBoolean(true));
installation.setInstalledDomall(new AtomicBoolean(true));

CathDomain domain = installation.getDomainByCathId("1oaiA00");
```

`ensureDomainListInstalled` then returns immediately, and `ensureDomallInstalled` returns at its own `installedDomall` check, so nothing is fetched.

The other three tests are unaffected — they assert on the thrown `IOException` and never touch the accessor.

aalhossary, line 48: ### This is why both Windows jobs fail

```
CathInstallationTest.testParseCathDomainListSuccess:50  expected: <490> but was: <8>
```

`8` is the *real* topology id for `1oaiA00` in CATH. The fabricated data is parsed correctly and then overwritten by the actual domain list.

The route is `getDomainByCathId` → `ensureDomallInstalled` → `ensureDomainListInstalled`. Calling `parseCathDomainList(reader)` directly populates `domainMap` but never sets `installedDomainList`, so the accessor still believes nothing is installed, downloads the real file and re-parses it over the test data.

**The three Linux passes are luck rather than correctness.** `ensureDomainListInstalled` swallows download failures:

```java
} catch (Exception e){
    LOGGER.error("Could not download CATH domain list file. Error: {}", e.getMessage());
    installedDomainList.set(false);
    return;      // test data survives
}
```

On the Linux runners the download did not complete in time, so the fabricated data survived and the assertions passed. On Windows it completed. That is a race against a large download, not an OS difference. The same test can fail on Linux on any day CATH happens to be fast.

Small irony: it only reaches the server at all because of `ed36d3a` in #1133. Before that, plain http always 301'd and the download always failed, so this test would have passed everywhere for the wrong reason.

**The Fix**:

```java
installation.parseCathDomainList(reader);
installation.setInstalledDomainList(new AtomicBoolean(true)); //1
installation.setInstalledDomall(new AtomicBoolean(true)); //2

CathDomain domain = installation.getDomainByCathId("1oaiA00");
```

`ensureDomainListInstalled` then returns immediately, and `ensureDomallInstalled` returns at its own `installedDomall` check, so nothing is fetched.

**WRONG Fix**: Adjusting the fake values to match real CATH instead would go green, but it would not be a fix: the download still happens, the assertions would then be checking the real file rather than the fake input (so `parseCathListFileLine` could be broken and the test would still pass), and it would break on the next CATH reclassification.

The other three tests are unaffected — they assert on the thrown `IOException` and never touch the accessor.

aalhossary (changes requested):

PR #1147, opened 2026-08-28
Contact calculation performance tweaks

josemduarte:
A few performance tweaks that increase contact calculation performance ~ 1.5x

- Use sqrt only when needed, use distance squared otherwise
- Primitive arrays instead of lists in GridCell
- Presize HashMap

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/contact/AtomContactSet.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/contact/AtomContactSet.java
@@ -37,12 +37,32 @@ public class AtomContactSet implements Serializable, Iterable<AtomContact> {
 
 	private static final long serialVersionUID = 1L;
 
+	/**
+	 * The default load factor of a {@link HashMap}, needed to size the map from an expected number of
+	 * entries.
+	 */
+	private static final float DEFAULT_LOAD_FACTOR = 0.75f;
+
 	private HashMap<Pair<AtomIdentifier>, AtomContact> contacts;
 	private double cutoff;
 
 	public AtomContactSet(double cutoff) {
 		this.cutoff = cutoff;
 		this.contacts = new HashMap<>();
+	}
+
+	/**
+	 * Creates an AtomContactSet sized to hold the given number of contacts, so that the underlying map
+	 * doesn't have to be repeatedly resized and rehashed as contacts are added. Contact calculations
+	 * produce hundreds of thousands of contacts for a large structure, where the repeated rehashing
+	 * that growing from the default capacity entails is a significant part of the cost.
+	 * @param cutoff the distance cutoff
+	 * @param expectedSize the number of contacts expected to be added. Only affects performance: an
+	 *                     over-estimate merely leaves the map larger than it needs to be.
+	 */
+	public AtomContactSet(double cutoff, int expectedSize) {
+		this.cutoff = cutoff;
+		this.contacts = new HashMap<>((int) (expectedSize / DEFAULT_LOAD_FACTOR) + 1);
 	}
 
 	public void add(AtomContact contact) {
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/contact/AtomContactSet.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/contact/AtomContactSet.java
@@ -28,6 +28,22 @@ import java.util.*;
 
 /**
  * A set of atom-atom contacts to hold the results of intra and inter-chain contact calculations
+ * <p>
+ * Contacts are keyed by the <i>ordered</i> pair of {@link AtomIdentifier}s of the 2 atoms, i.e. the
+ * pair (a,b) and the pair (b,a) are 2 different keys. Thus look-ups ({@link #hasContact(Atom, Atom)},
+ * {@link #getContact(Atom, Atom)}) must give the 2 atoms in the same order in which the contact was
+ * calculated. The order produced by the calculation ({@link Grid}) is:
+ * <ul>
+ * <li>for contacts within a single set of atoms (e.g. the intra-chain contacts of
+ * <code>StructureTools.getAtomsInContact(Chain, double)</code>), the atom that comes first in the
+ * atom array is the first member of the pair</li>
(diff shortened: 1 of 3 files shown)


biojava-structure/src/main/java/org/biojava/nbio/structure/contact/AtomContactSet.java
```diff
@@ -37,6 +37,12 @@ public class AtomContactSet implements Serializable, Iterable<AtomContact> {
 
 	private static final long serialVersionUID = 1L;
 
+	/**
+	 * The default load factor of a {@link HashMap}, needed to size the map from an expected number of
+	 * entries.
+	 */
+	private static final float DEFAULT_LOAD_FACTOR = 0.75f;
+
 	private HashMap<Pair<AtomIdentifier>, AtomContact> contacts;
```
aalhossary, line 62: `AtomIdentifier` contains only the Atom **serial number** and **Chain ID**. What about the Alternate Location? Isn't it important when we want to calculate contact matrix? What if an Atom is within the cutoff in an AltLoc but not in the other?
Taking into consideration the the `Atom` interface itself defines `setAltLoc()` and `getAltLoc()`, this may actually need a substantial change in the structure.
Or what do you think?

biojava-structure/src/main/java/org/biojava/nbio/structure/contact/AtomContactSet.java
aalhossary, line 129: Shouldn't the atom order be enforced here?
e.g.
`if atom1 serial < atom2 serial then return hasContact(atomId2,atomId1)`

biojava-structure/src/main/java/org/biojava/nbio/structure/contact/GridCell.java
```diff
@@ -74,23 +106,31 @@ public List<Contact> getContactsWithinCell(){
...
+				Point3d atomI = iAtoms[i];
+				for (int b=0; b<numIindices; b++) {
+					int j = iIndices[b];
 					if (j>i) {
-						double distance = iAtoms[i].distance(iAtoms[j]);
-						if (distance<cutoff) contacts.add(new Contact(i, j, distance));
+						double distanceSq = atomI.distanceSquared(iAtoms[j]);
+						if (distanceSq<cutoffSq) contacts.add(new Contact(i, j, Math.sqrt(distanceSq)));
 					}
 				}
```
aalhossary, line 123: As long as you are comparing atoms of **iAtoms** to **iAtoms** (not to **jAtoms**), then you can safely make it:
`for (int b = a+1; b<numIindices; b++) { // instead of int b = 0`
and drop the check `if (j>i)` to be
```
				for (int b = a + 1; b<numIindices; b++) {
					int j = iIndices[b];
					double distanceSq = atomI.distanceSquared(iAtoms[j]);
					if (distanceSq<cutoffSq) contacts.add(new Contact(i, j, Math.sqrt(distanceSq)));
				}
```

aalhossary (changes requested):
I recommend 2 small updates and raised another concern that may need a substantial change.
This is Amr (Not the AI 😊).

biojava-structure/src/main/java/org/biojava/nbio/structure/contact/AtomContactSet.java
josemduarte: I think that is not an issue in practice because BioJava only reads one of the alternative comformations (see https://github.com/biojava/biojava/blob/dafb0cdc67b6348f7c8fa988b7529e7367f4c6db/biojava-structure/src/main/java/org/biojava/nbio/structure/Group.java#L117). Additionally even if both alt locs were there, they are assigned different atom serial numbers.

biojava-structure/src/main/java/org/biojava/nbio/structure/contact/GridCell.java
josemduarte: Good point. I did come across this when doing the refactoring with Claude. After checking it, the performance gain was about 1% and I decided against it not to introduce unneeded complexity in the code, and also not to rely on the fact that the indices arrays happen to be sorted.

biojava-structure/src/main/java/org/biojava/nbio/structure/contact/AtomContactSet.java
aalhossary: You are right. It will have a different serial number

biojava-structure/src/main/java/org/biojava/nbio/structure/contact/GridCell.java
aalhossary: OK as you like

biojava-structure/src/main/java/org/biojava/nbio/structure/contact/AtomContactSet.java
aalhossary: @josemduarte I checked Pair class [hashCode()](https://github.com/biojava/biojava/blob/dafb0cdc67b6348f7c8fa988b7529e7367f4c6db/biojava-structure/src/main/java/org/biojava/nbio/structure/contact/Pair.java#L65) and confirmed that `Pair(atom1, atom2).hashCode() != Pair(atom2, atom1).hashCode()`, which is **definitely** correct.
However, for **_contact_** it should not be the case. We should update `hasContact()`, `getContact()` and `getAtomIdPairFromContact()`
Do you agree?

josemduarte: I think you are right. But the issue is that changing this implementation can be a disruptive change for anywhere that assumes the current behaviour. I think what we can do right now is document it properly so no one trips over this. I've now added docs in e73ce56f

aalhossary: OK I'll create a new issue for this point, leave this thread unresolved, while approving the PR.
Hopefully I can find time (I doubt) to study the effect of the proposed update on other BioJava code and outer code.
Do you expect such behaviour change to qualify for BioJava **7.3.0** or **8.0.0**?

aalhossary (approved):
The PR provides useful updates.
However, reviewing it raised an issue (a potential bug), to be addressed later.

PR #1148, opened 2026-08-28
ASA calculation performance tweaks

josemduarte:
The runtime gain should be around 1.2-1.3x. And together with #1147 it is about 1.3-1.5x 

Implemented by Claude and fully tested. Output is identical to the previous implementation.

The changes are to replace objects with flat primitive arrays so the hot loops are sequential scans over contiguous memory instead of pointer chases.


sbittrich (approved):
LGTM, thanks

Issue #1149, opened 2026-08-30
FileDownloadUtils detects redirects but does not follow the ones HttpURLConnection won't (307/308, cross-protocol)


aalhossary:
`FileDownloadUtils` learned in #1133 to *detect* redirects and refuse to cache their bodies. It never learned to *follow* the ones `HttpURLConnection` will not follow by itself — and that gap has now broken the build twice in nine months, for two different services.

## The JDK's two blind spots

`HttpURLConnection` follows 301, 302 and 303 **within the same protocol**. It does not follow 307 or 308 at all, and it does not follow any redirect that changes http to https. Both apply to us right now, with `setInstanceFollowRedirects(true)`:

```java
for (String u : new String[]{
        "http://prodata.swmed.edu/ecod/distributions/ecod.latest.domains.txt",
        "http://download.cathdb.info/cath/releases/"}) {
    HttpURLConnection c = (HttpURLConnection) new URL(u).openConnection();
    c.setRequestMethod("HEAD");
    c.setInstanceFollowRedirects(true);          // the default
    System.out.printf("%s -> %d  Location=%s%n", u, c.getResponseCode(), c.getHeaderField("Location"));
}
```

```
.../ecod/distributions/ecod.latest.domains.txt  -> 308  Location=/ecod-legacy/distributions/ecod.latest.domains.txt
http://download.cathdb.info/cath/releases/      -> 301  Location=https://download.cathdb.info/cath/releases/
```

The ECOD one is the sharper case: **same host, same protocol, relative `Location`** — a redirect that would be followed transparently if it were a 301, and is simply dropped because it is a 308.

## What it has cost us so far

| When | Service | Redirect | Symptom |
|---|---|---|---|
| 2025-12 | CATH | 301 http→https | eight months of red CI (#1138); the redirect body was cached as classification data, so the failure surfaced later and elsewhere as an NPE |
| 2026-08-29 | ECOD | 308 path move | nightly red on all `EcodInstallationTest` (runs `33303708377`, `33247173632`) |

ECOD relaunched their site, so `/ecod/` is now an application and the static files were rewritten to `/ecod-legacy/distributions/`. Their own pages still link to `/ecod/distributions/`, and their API returns absolute URLs on that path, so the old address remains the advertised one — this is an internal rewrite, not a published move. **Chasing the new path in our code would be following an implementation detail they do not advertise**, which is why this issue proposes following the redirect rather than editing `DOMAINS_PATH`.

[...]

PR #1150, opened 2026-08-30
Keep the ECOD read lock balanced when loading fails

aalhossary:
`getDomainsForPdb` and `getAllDomains` release the read lock **inside** their `try` block — so the loader they call can take the write lock — and re-acquire it immediately afterwards:

```java
domainsFileLock.readLock().lock();
try {
    while (domainMap == null) {
        domainsFileLock.readLock().unlock();
        indexDomains();                       // if this throws...
        domainsFileLock.readLock().lock();    // ...this never runs
    }
    ...
} finally {
    domainsFileLock.readLock().unlock();      // ...and this unlocks a lock we do not hold
}
```

When the loader throws, the re-acquisition is skipped, and the outer `finally` unlocks a lock the thread no longer holds. `IllegalMonitorStateException` is then thrown *from a finally block*, so it **supersedes the `IOException` that actually caused the failure**.

## Why it matters, concretely

Last night's nightly (`33303708377`) is a clean demonstration. One upstream change — ECOD now answers `http://prodata.swmed.edu/ecod/distributions/…` with a 308 that `HttpURLConnection` will not follow, see #1149 — produced eight failures wearing three different faces:

```
EcodInstallationTest.testDownloads:93   » HttpStatus HTTP 308 Permanent Redirect for http://…
EcodInstallationTest.testByPDB:148      » IllegalMonitorState attempt to unlock read lock, not locked by current thread
EcodInstallationTest.testAllDomains:104 NullPointer Cannot invoke "…getAllDomains()" because "ecod" is null
```

Only the first names the cause. The second is this bug. Diagnosing it meant reading past the noise to find the three tests that had told the truth.

## The fix

Re-acquire in a `finally` of its own, so the lock is balanced on both paths and the original exception propagates intact:

```java
domainsFileLock.readLock().unlock();
try {
    indexDomains();
} finally {
    domainsFileLock.readLock().lock();
}
```

Six lines in each of the two methods, no behaviour change on the success path. The other four read-lock sites in the class are already balanced — they lock, return, and unlock in a `finally` without releasing in between — so they are untouched.

## Scope

This does not fix the ECOD download itself; that is #1149, and it needs `FileDownloadUtils` to follow the redirects the JDK declines. This only stops a failing load from disguising its own cause.

The third face above — `EcodFactory` handing back `null` after swallowing the error — is the same family of masking and is also not addressed here.

## Testing

`mvn test -pl biojava-structure -Dtest=EcodParserTest` — green, and the class compiles unchanged otherwise. The failure path is not covered by an automated test: reproducing it needs a download that throws, which today means either the network being down or ECOD still being broken. If you would like it pinned, the natural way is a small unit test that injects a failing loader, which would mean making the loader overridable — happy to add that if you think it is worth the seam.

### DIFF
--- a/biojava-structure/src/main/java/org/biojava/nbio/structure/ecod/EcodInstallation.java
+++ b/biojava-structure/src/main/java/org/biojava/nbio/structure/ecod/EcodInstallation.java
@@ -137,9 +137,15 @@ public class EcodInstallation implements EcodDatabase {
 				// unlock to allow ensureDomainsFileInstalled to get the write lock
 				logger.trace("UNLOCK readlock");
 				domainsFileLock.readLock().unlock();
-				indexDomains();
-				domainsFileLock.readLock().lock();
-				logger.trace("LOCK readlock");
+				try {
+					indexDomains();
+				} finally {
+					// re-acquire even if indexing failed, so the outer finally has a
+					// lock to release; otherwise IllegalMonitorStateException replaces
+					// the real cause and the failure becomes unreadable
+					domainsFileLock.readLock().lock();
+					logger.trace("LOCK readlock");
+				}
 			}
 
 			PdbId pdbId = null;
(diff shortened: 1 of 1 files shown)


josemduarte (approved):
LGTM thanks

PR #1151, opened 2026-08-30
Follow the redirects HttpURLConnection declines (307/308, cross-protocol)

### LINKED ISSUE #1149: FileDownloadUtils detects redirects but does not follow the ones HttpURLConnection won't (307/308, cross-protocol)

aalhossary:
Fixes #1149.

`HttpURLConnection` follows 301, 302 and 303 within a protocol. It never follows 307 or 308, and never follows a redirect that changes http to https. Both gaps have broken this build:

| When | Service | Redirect |
|---|---|---|
| 2025-12 | CATH | 301 http→https — cached the redirect body as classification data (#1138) |
| 2026-08-29 | ECOD | 308 to a rewritten path — nightly red on all `EcodInstallationTest` |

A browser follows either without comment, so neither service had reason to expect the change would break us.

## What this adds

`openConnectionFollowingRedirects(URL, timeout)` handles what the JDK leaves:

- **307 and 308**, and **cross-protocol 301/302/303** — the only ones that reach us, since the JDK already consumed the rest.
- **Relative `Location`** resolved against the request. ECOD's is relative.
- **Five-hop cap**, and a loop reported by name rather than chased.
- **https→http is refused.** A redirect must not downgrade the transport silently. Anything refused is returned as-is, so `checkHttpStatus` still decides what a non-2xx means.

Used by `downloadFile`, `downloadFileWithValidation` and the `URL` overload of `createValidationFiles`.

The decision is split into `redirectTargetFor(code, location, url)` so the rules are testable without a server.

## Note on ECOD specifically

This deliberately does **not** change `DOMAINS_PATH`. ECOD's own site and API still publish `/ecod/distributions/…`; the 308 to `/ecod-legacy/` is an internal rewrite they introduced when their new app took over `/ecod/`. BioJava's URL is already the advertised one — the only defect was not following the redirect. Hardcoding `/ecod-legacy/` would pin us to something unadvertised.

## Testing

`mvn test -pl biojava-core -Dtest=FileDownloadRedirectTest` — 13 tests, no network:

- 7 rule tests, pure, covering each status, relative and absolute locations, the https→http refusal, and unusable locations.
- 6 end-to-end through a local `HttpServer`: a 308 and a 307 followed, the validated download storing the real body with a matching `.size`, a loop, an over-long chain, and a redirect with no `Location`.

Full `biojava-core` suite: 535 tests, no failures.

## Scope

Other call sites have the same exposure — `AllChemCompProvider`, `SiftsChainToUniprotMapping`, `Astral`, `PDBStatus`, `URLConnectionTools`, and `UniprotProxySequenceReader`, whose hand-rolled version carries the comment *"This method should be moved to a utility class in BioJava 5.0"*. Migrating them is a follow-up rather than part of this, since each needs its own judgement; inventory filed separately.

### DIFF
--- a/biojava-core/src/main/java/org/biojava/nbio/core/util/FileDownloadUtils.java
+++ b/biojava-core/src/main/java/org/biojava/nbio/core/util/FileDownloadUtils.java
@@ -29,6 +29,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintStream;
 import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
 import java.net.SocketTimeoutException;
 import java.net.URL;
 import java.net.URLConnection;
@@ -38,7 +39,9 @@ import java.nio.file.attribute.BasicFileAttributes;
 import java.security.DigestInputStream;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.util.LinkedHashSet;
 import java.util.Scanner;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -53,6 +56,9 @@ public class FileDownloadUtils {
 
 	/** Buffer used when streaming a file through a {@link MessageDigest}. */
 	private static final int DIGEST_BUFFER_SIZE = 64 * 1024;
+
+	/** Redirects to follow before giving up, in case a server sends us in a circle. */
+	private static final int MAX_REDIRECTS = 5;
 
 	/** A bare hex digest, optionally followed by whitespace and a file name (the
 	 * layout written by <code>md5sum</code>, <code>sha1sum</code> and friends). */
@@ -139,8 +145,7 @@ public class FileDownloadUtils {
 		try {
 			while (true) {
 				try {
-					URLConnection connection = prepareURLConnection(url.toString(), timeout);
-					connection.connect();
+					URLConnection connection = openConnectionFollowingRedirects(url, timeout);
 					checkHttpStatus(connection);
 					try (InputStream inputStream = connection.getInputStream()) {
 						// Files.copy loops until end of stream. FileChannel.transferFrom(), used
@@ -199,8 +204,7 @@ public class FileDownloadUtils {
 
 		File tempFile = createTempFileFor(destination);
 		try {
-			URLConnection connection = prepareURLConnection(url.toString(), timeout);
-			connection.connect();
+			URLConnection connection = openConnectionFollowingRedirects(url, timeout);
 			checkHttpStatus(connection);
 
 			long declaredSize = connection.getContentLengthLong();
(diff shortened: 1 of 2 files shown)


josemduarte (approved):
LGTM thanks

PR #1154, opened 2026-09-06
Do not pin the status an absent file returns

aalhossary:
`files.wwpdb.org` now sits behind Amazon S3, which answers a missing key with **403** rather than 404 when the caller cannot list the bucket:

```
$ curl -sI https://files.wwpdb.org/pub/pdb/data/structures/divided/mmCIF/zz/zzzz.cif.gz
HTTP/1.1 403 Forbidden
Server: AmazonS3
```

`notFoundThrowsAndLeavesNothingBehind` asserted 404, so it has failed every nightly since 2 September. Because it is in `biojava-core`, the other thirteen modules are skipped — including `biojava-integrationtest`, so the ECOD failure behind it (#1149) is invisible.

Nothing is broken: real files still return 200 with `Content-Length`, `Last-Modified`, `ETag` and `Accept-Ranges` intact, the same MD5 as before.

The test now asserts the contract rather than the status code — an error status throws, nothing is written, no `.size` sidecar — and `isNotFound()` gets its own offline test, since 404/410 versus 403/500 decides whether the download fallback chain continues or aborts.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/biojava-core/src/test/java/org/biojava/nbio/core/util/FileDownloadUtilsTest.java
+++ b/biojava-core/src/test/java/org/biojava/nbio/core/util/FileDownloadUtilsTest.java
@@ -217,9 +217,19 @@ class FileDownloadUtilsTest {
     @Nested
     class HttpStatus {
 
+        /**
+         * Which status an absent file comes back with is the server's business, and it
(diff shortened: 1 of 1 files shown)


josemduarte (approved):
LGTM thanks

PR #1156, opened 2026-09-14
Fix: CIF parser wasn't respecting parseCAOnly

josemduarte:
Including a new unit test for it.


biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/CifStructureConsumerImpl.java
```diff
@@ -319,7 +319,8 @@ public void consumeAtomSite(AtomSite atomSite) {
             }
 
             if (params.isParseCAOnly()) {
-                if (!labelAtomId.get(atomIndex).equals(StructureTools.CA_ATOM_NAME) && "C".equals(typeSymbol.get(atomIndex))) {
+                // keep only C-alpha atoms; requiring element C excludes calcium ions, whose atom name is also CA
+                if (!(labelAtomId.get(atomIndex).equals(StructureTools.CA_ATOM_NAME) && "C".equals(typeSymbol.get(atomIndex)))) {
```
sbittrich, line 323:
```suggestion
        for (int atomIndex = 0; atomIndex < atomSite.getRowCount(); atomIndex++) {
            if (params.isParseCAOnly()) {
                // keep only C-alpha atoms; requiring element C excludes calcium ions, whose atom name is also CA
                if (!(labelAtomId.get(atomIndex).equals(StructureTools.CA_ATOM_NAME) && "C".equals(typeSymbol.get(atomIndex)))) {
                    continue;
                }
            }

            boolean startOfNewChain = false;
```

There may be another bug here. Because the check happens this late, lots of empty groups and chains are created.
How about moving it all the way up and making it the first action in the loop?

sbittrich (approved):
LGTM, thanks.
Would consider moving the fixed logic though.

biojava-structure/src/main/java/org/biojava/nbio/structure/io/cif/CifStructureConsumerImpl.java
josemduarte: Indeed, good point. The empty chain/groups are actually another bug by itself. Should be fixed in [0b8427a](https://github.com/biojava/biojava/pull/1156/commits/0b8427a68aa027e814babab7ccd435080481a8f1)

sbittrich (approved):
Nice, thank you

